PR #3, opened 2023-08-17
Memory duplication speed up

johnryan465:
By making the shared memory D times larger, we can reduce the amount of collisions when performing atomic adds reducing the overall time. 
Current benchmark time: 151 seconds on 3090


MrNeRF (approved):
looks good to me. Thank you!

PR #15, opened 2023-08-20
Backward pass speed up

johnryan465:
Runs on a 3090 in between 131 and 134 seconds, and approximate 7 second speed up.

Approach is to use local memory to save some intermediate values to allow for offsetting atomic adds to reduce wait times.


MrNeRF:
Nice, sound phenomenal! I am keen to test it out.
But currently, it is hard for me to review it and it has a merge conflict. Can you please pull current main, resolve it, and squash your commits so that I can review one crisp commit?

johnryan465:
Squashed and rebased with your latest changes and verified training correctness. @MrNeRF

MrNeRF:
@johnryan465 I've run multiple tests to ensure the accuracy of my observations. While I did notice a considerable speedup (1:13 wow!), the number of splats and growth observed were inconsistent. Moreover, the visual output didn't meet expectations. Perhaps an issue arose during the rebase?

In the rasterizer.cuh file, there's a WRITE_TEST_DATA macro. If you define it, it enables writing test data for a single iteration during the backward pass.

When building the tests with the -DBUILD_TESTS=ON flag, it's possible to produce an output file from the most recent main commit. This output can be utilized to verify computations in the RasterizeGaussiansTest during the backward pass. If you're interested in multiple iterations, consider defining a static counter variable. This allows for renaming the tests based on its current iteration count.

Screenshot from 2023-08-20 21-51-54

johnryan465:
Do you have any comparision images when you say the visual output doesn't meet expectations? In my limited testing I don't notice substantial differences with the caveat that the viewer segfaults for me after a few seconds open so hard to do a proper comparision to identify what you mean to try and address that.

MrNeRF:
This is the quality I would expect after 7000 iterations. If it takes too long, you can also try to stop earlier at 5500 or so...
I am just guessing but I think the main problem with the viewer is, that it loads the data within the rendering loop and completely blocks it. It works for me if I don't touch anything while it starts. As soon as it starts rendering, it does not crash anymore. 

Screenshot from 2023-08-21 08-54-51

johnryan465:
The consistency of the number of splats is an issue which I didn't catch, have converted to draft until addressed.

johnryan465:
@MrNeRF The base visual performance I am getting from master is much poorer than what you have shown. There is a  chance that this is caused by my viewer setup (I'm running it in a docker container as it won't compile locally due to newer versions of libraries). The image with the worse wheel artifacts is the current master, is your provided image from master?. It would be helpful if you could test it yourself. One additional thing that could be the cause is that I added a block.sync() call to my PR which I believe is needed for correctness.  
Point view_2023_08_21_12_39_03   
Point view_2023_08_21_12_45_34
Edit: That wasn't a bug, the randomness of the file writing made it seem like changing that fixed the issue.

MrNeRF:
Wow. That looks evil. I removed my build folder and the rendering looks now about the same for me. Something seems to stick in the build folder and survives the rebuilds. I am on it trying to find the error.

johnryan465:
I wasn't changing the final iteration in main.cu flag to true when exiting at 7000 iterations and I think only part of the file was being writen which was causing some of these random visual artifacts. With that fixed I'm getting very nice looking images for the PR now.       
image

MrNeRF:
Yeah, I also stumbled over it. I made a fresh checkout and the image looks nice and crispy. Ok, now I gonna look in depth into your pull request! Puuh, I thought something really severe was messed up.

MrNeRF:
I've noticed that it appears to be cloudier. Did you create a test file to determine if there has been a significant change in the calculations? Could you please compare it again with the master? Perhaps, you could train a model using the master and another with your modifications, and then run two instances of the viewer? This way, you should be able to see any differences immediately. I've also observed a reduced number of splats. I'm uncertain about compromising quality for speed, although my current timing stands at 1:16 (including writing the output) which is pretty nice.

Also, please use clang-format on the files you're committing and always squash so I can better see what is going on. It slowly starts becoming obfuscated :-). For me I've added a script in the git hooks which should automatically execute before a commit is made. If you need it, I can provide the script.

johnryan465:
It would be great to have all of the steps, perhaps added to the readme for contributors. Having run both viewers at the same time, I see what you mean about less clarity. There is no intentional reduction in the number of splats for performance in what I'm doing. From the paper this seems to imply that the magnitude of the gradients are smaller in this PR.

MrNeRF:
Good point, I gonna write tonight some contribution guidelines. Basically squash and apply "clang-format -i" to changed files. Currently there is a lot to do to get everything straight and to make it a nice developer experience.

I just wondering if there is some subtle bug or just small numerical differences that add up. Either way, I can take later a deeper look. I am thankful for your great work. I just want to make sure that we  keep the visual quality high.

PR #26, opened 2023-08-23
Status line

paulmelis:
Okay, seems I mistimed trying to add a status line with your addition of the convergence threshold :)

This adds a status line that is continuously refreshed with basic stats, including avg iterations/s and overall time. I commented out the periodic pruning and opacity-clearing messages as they were causing the status line to get newlined and start over.

If you don't want to use this then don't.

PS I need to start using feature branches on my end...

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ I embarked on this project to deepen my understanding of the groundbreaking pape
   - Command-line parameters have been added to specify the training data path and the output path. Note that the output path will not be automatically overwritten anymore.
   - CUDA version restrictions have been relaxed a bit to 11.7 or higher. The CUDA architecture of the system on which the software is being compiled is now auto-detected automatically.
   - Experimental feature to monitor the average convergence rate throughout training has been added (see the command-line options section for more details).
+  - Added updating status line, instead of printing many lines of output during a run
   - There are a lot good first issues to grab if you would like to contribute.
   
 If you encounter any problems or issues, please [open an issue](https://github.com/MrNeRF/gaussian-splatting-cuda/issues) on GitHub.
--- a/src/gaussian.cu
+++ b/src/gaussian.cu
@@ -158,11 +158,9 @@ void prune_optimizer(torch::optim::Adam* optimizer, const torch::Tensor& mask, t
 }
 
 void GaussianModel::prune_points(torch::Tensor mask) {    
-    //std::cout << "Before pruning " << mask.sizes()[0] << " points" << std::endl;
     // reverse to keep points
     auto valid_point_mask = ~mask;
     int true_count = valid_point_mask.sum().item<int>();
-    //std::cout << "Pruning left " << true_count << " points" << std::endl;
     auto indices = torch::nonzero(valid_point_mask == true).index({torch::indexing::Slice(torch::indexing::None, torch::indexing::None), torch::indexing::Slice(torch::indexing::None, 1)}).squeeze(-1);
     prune_optimizer(_optimizer.get(), indices, _xyz, 0);
     prune_optimizer(_optimizer.get(), indices, _features_dc, 1);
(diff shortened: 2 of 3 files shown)


MrNeRF:
Yeah, that's useful information. I like it. It should be a bit more compact, and the columns more aligned (there are fast more than 1M. splats). I would abbreviate most information:
iter, loss, acr, splats, and maybe points before and after the the pruning operation => before/after.

So this information is refreshing in a single line on your console? How that?
If you provide a short note, when you working on an issue, might be helpful for others to know. Anyway, cool stuff!

paulmelis:
> Yeah, that's useful information. I like it. It should be a bit more compact, and the columns more aligned (there are fast more than 1M. splats).  

More compact makes sense. But each field already has a fixed width (and precision, if float), including fitting millions of splats?

> iter, loss, acr, splats, and maybe points before and after the the pruning operation => before/after.

The pruning stats are a bit weird to handle, as they have their own update interval, which is independent of the stats update interval. Edit: I can print something like "last pruning X -> Y points".

> So this information is refreshing in a single line on your console? How that? If you provide a short note, when you working on an issue, might be helpful for others to know. Anyway, cool stuff!

It output the stats line without ending with a newline. Then on the next update it prints a single `\r` to get the cursor back to the beginning of the line and then overwrites with new data.

MrNeRF:
> More compact makes sense. But each field already has a fixed width (and precision, if float), including fitting millions of splats?
> 
Screenshot from 2023-08-23 20-52-55
I think if you abbreviate the names of the values, you will get automatically a more compact representation.
> > iter, loss, acr, splats, and maybe points before and after the the pruning operation => before/after.
> 
> The pruning stats are a bit weird to handle, as they have their own update interval, which is independent of the stats update interval.
> 
So than leave it out. I think right now it is also giving no real additional value. It should be the difference in splats before and after pruning.  I used this to monitor when the part was still buggy. You can infer this from the total number of splats.
> > So this information is refreshing in a single line on your console? How that? If you provide a short note, when you working on an issue, might be helpful for others to know. Anyway, cool stuff!
> 
> It output the stats line without ending with a newline. Then on the next update it prints a single `\r` to get the cursor back to the beginning of the line and then overwrites with new data.

Seems not to work for me as you see in the screenshot. But that should not stop you. Might be due to my terminal application or zsh? Not sure, will figure this out

paulmelis:
> Seems not to work for me as you see in the screenshot. But that should not stop you. Might be due to my terminal application or zsh? Not sure, will figure this out

No, it's the Pruning part (which gets printed elsewhere, with a newline) that forces the many status lines, instead of a single one

paulmelis:
Okay, I shortened the item names. Can you try again if you now get a single line that gets updated periodically?

paulmelis:
By the way, I now see that in your case the thousand separators *are* printed as they should. For some reason they don't in my case. And indeed with thousands seps the column widths are slightly too small, will fix that.

MrNeRF:
> No, it's the Pruning part (which gets printed elsewhere, with a newline) that forces the many status lines, instead of a single one

You are right. I only applied the changes from main and did not notice that everything is printed in one line until the pruning starts. I'll test the updated version now.

MrNeRF:
Great, I think this is pretty good. You are making the users happy :)
Instead of commenting out the std::cout stuff in the gaussian.cu, please just delete it. 
And please, also don't forget to make an update whats new today in the Readme.md. We should keep this rolling like a sliding window maybe over the past 7 days. I see this as marketing tool. So people can immediately see that this project is alive and I hope we gain more contributors over time to raise the pace we are pushing forward.

MrNeRF:
I think we are ready for prime time, right? Thank you! 

A future TODO might be to factor out the logging to unclutter main. But that's up to you if you like to do this next or something else. At least my backlog feels infinite at the moment :)

MrNeRF (approved):

PR #29, opened 2023-08-25
[ReadMe] add tbb to software dependencies

cpheinrich:
Appreciate all the work on this faster implementation! The ReadMe was very clear -- only issue I hit was an undocumented dependency on TBB, which I've added here.


MrNeRF:
Thank you Chris!

PR #31, opened 2023-08-29
When images fail to load, show a reason why (as reported by STB)

paulmelis:


MrNeRF:
Nice, wasn't aware of stbi_failure_reason(). Thx!

PR #33, opened 2023-08-30
Show a line with info the set of training images

paulmelis:


src/read_utils.cu
```diff
@@ -440,7 +440,14 @@ std::unique_ptr<SceneInfo> read_colmap_scene_info(std::filesystem::path file_pat
         sceneInfos->_point_cloud = read_ply_file(file_path / "sparse/0/points3D.ply");
     }
     sceneInfos->_ply_path = file_path / "sparse/0/points3D.ply";
-    sceneInfos->_cameras = read_colmap_cameras(file_path / "images", cameras, images);
+
+    auto& cams = sceneInfos->_cameras = read_colmap_cameras(file_path / "images", cameras, images);
```
MrNeRF, line 444: Could you please separate this in two lines

src/read_utils.cu
```diff
@@ -440,7 +440,14 @@ std::unique_ptr<SceneInfo> read_colmap_scene_info(std::filesystem::path file_pat
...
+    const float image_mpixels = cams[0]._img_w * cams[0]._img_h / 1000000.0f;
```
MrNeRF: And maybe name this constant 1'000'000.f  If you use these upper ticks, than its easier to read (for instance, see parameters.cuh)

src/read_utils.cu
MrNeRF: Could you please separate this in two lines. Its a bit hacky.

MrNeRF (changes requested):
There are two small comments. Otherwise great addition!

MrNeRF (approved):

PR #52, opened 2023-12-26
added NVIDIA RTX 3050

Venkatakrishnan-Ramesh:
Added an Version for Ubuntu 20.04 , NVIDIA RTX 3050


MrNeRF:
Cool. Thx. Somewhat slow. Did you test on the original implementation? Slowdown might be due how shared memory is used which might not suite less powerful gpus well.

Venkatakrishnan-Ramesh:
Original implementation ,wasn't supporting in my system when i tried with a kaggle notebook using t4 it took around 7 mins approx. Can you guide in how to reduce the time lag even more

PR #71, opened 2025-04-30
use Torch_DIR to find libtorch instead of CMAKE_PREFIX_PATH

miriameng:
As described in https://cmake.org/cmake/help/latest/command/find_package.html, `<PackageName>_DIR` can be used for `find_package()`, the motivation of switching from `CMAKE_PREFIX_PATH` to `<PackageName>_DIR` for libtorch is because ideally `CMAKE_PREFIX_PATH` should be reserved for a central package manager, e.g. vcpkg, for finding other dependencies specified in `find_package()`, e.g. TBB (not by default available on Windows, can be built by vcpkg).


MrNeRF:
Lgtm! Ty!

PR #73, opened 2025-05-09
read quaternions correctly and use toRotationMatrix()

miriameng:
colmap actually stores the quaternions in `w,x,y,z` instead of `x,y,z,w`, if we read it correctly we can just call `toRotationMatrix()` instead of needing to manually do the conversion ourselves to accidentally getting it right


MrNeRF (approved):
Lgtm! Thank you!

PR #74, opened 2025-05-14
remove transpose() on camera_info->_R that causes confusion for reading

miriameng:
Couldn't figure out why the `transpose()` was there at the first place, in my opinion this causes confusion and should be removed. After fixing up `getWorld2View2` and `getWorld2View2Eigen` I can confirm that the truck dataset works like usual.


MrNeRF:
Lgtm!

PR #75, opened 2025-05-15
remove unnecessary std::reverse

miriameng:
`std::reverse` in `get_random_indices` looks like unnecessary computation to me, so proposing to remove it, also truck dataset still works like usual after this change


MrNeRF:
I agree! That's not necessary!

PR #77, opened 2025-05-23
use torch::abs instead of torch::linalg::vector_norm for better readability since grads is [N, 1]

miriameng:
Probably will save people a few mins who read the code in the future. At least initially I got confused for a little bit on why L2 norm is called.


MrNeRF:
Yes true, but the it even does not need the abs right? Because the norm is already taken in Add_densification_stats and that's always non-negative, right?

MrNeRF:
lgtm

PR #78, opened 2025-05-23
keep selected_pts_mask as kBool

miriameng:
converting `selected_pts_mask` to `kLong` looks like a mistake to me, although it doesn't affect the outcome


MrNeRF (commented):
lgtm

MrNeRF (approved):

PR #80, opened 2025-05-23
remove unused max_screen_size

miriameng:
What is the original intention of this arg? Doesn't seem to be used in any commented out code either. Let's remove for better readability?


MrNeRF:
That's fine. The intention was to prune Gaussian that take a lot of screen space. However, that's gone. It is fine to remove it!

PR #99, opened 2025-06-04
run gaussian-splatting-cuda successfully on Windows

miriameng:
I am able to run the truck dataset successfully on Windows with the changes in this PR.

Didn't use `TARGET_BUNDLE_DLLS` because some dlls (e.g. cudnn) are dynamic loaded so `TARGET_BUNDLE_DLLS` can't find them.

`.parent_path().parent_path().parent_path()` is needed because Visual Studio is assumed, and it is a multi config generator so there will be an extra layer of folder like `Debug` or `Release`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,6 +159,18 @@ target_link_libraries(${PROJECT_NAME}
         Python3::Python
 )
 
+if(WIN32)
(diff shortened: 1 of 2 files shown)


MrNeRF (approved):
Lgtm!

Note, I am not sure how good it matches gsplat quality performance yet but it should be faster.

You can chane the max_cap to get a higher fidelity (cmd arg tbd).

PR #102, opened 2025-06-10
save splat_iterations.ply instead of splat_iterations+1.ply

miriameng:
Now if `-i 1000`, it will save splat_1001.ply, this change makes it save splat_1000.ply, which I think should be better align with what the user expect it to behave


MrNeRF (approved):
Ty. Makes sense!

PR #118, opened 2025-06-17
Selective adam implementation

MrNeRF:
Shaves off at least one-third of the training time. Though the metrics are thrown under the bus :bus:  😅


MrNeRF:
Should be fine for now. Per default deactivated anyway.

PR #119, opened 2025-06-18
grad_outputs should already be in CUDA and contiguous

miriameng:
My understanding is grad_outputs should already be in CUDA and contiguous so these should be redundant and can be removed for better readability. Ran the truck dataset and it seems to run fine.


MrNeRF:
LGTM! Thank you!

PR #142, opened 2025-06-26
only link to dl in UNIX and there's no need to copy opengl dlls in Windows since they are in C:\Windows\System32

miriameng:


MrNeRF:
Ah, let's discuss this in an issue. I haven't used vcpkg yet so I don't know :)

Issue #146, opened 2025-07-01
Remove abs from scale and opacity regularizatiin


MrNeRF:
The abs is not necessary for opacity and scale regularization. Please remove it.
https://github.com/MrNeRF/gaussian-splatting-cuda/blob/9bfd4f22bf191744e9f03df5d1b59bd500a3034e/src/trainer.cpp#L55

https://github.com/MrNeRF/gaussian-splatting-cuda/blob/9bfd4f22bf191744e9f03df5d1b59bd500a3034e/src/trainer.cpp#L60

Tomoya-Matsubara:
Hi @MrNeRF!

Can I work on this?

MrNeRF:
Sure!

PR #149, opened 2025-07-01
Removed scale, opacity absloute value operation

OsamaS99:
Removed scale, opacity absloute value operation #146


MrNeRF:
I am sorry, this has been already submitted and merged. However, there is another beginner friendly issue open in case you are interested.

Thank you for your contribution!

PR #159, opened 2025-07-05
Fix: corrects 1D gaussian kernel in metrics

darshanmakwana412:
Corrects the bug in `metrics.cpp` which uses asymm 1D gaussian kernel for computing ssim
The affect on metric is negligible
SSIM with asymmetric 1D gaussian kernel SSIM: 0.85838
SSIM with symmetric 1D gaussian kernel SSIM: 0.85798

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -11,10 +11,18 @@ namespace gs {
 
         // 1D Gaussian kernel
         torch::Tensor gaussian(const int window_size, const float sigma) {
-            torch::Tensor gauss = torch::empty(window_size);
-            for (int x = 0; x < window_size; ++x) {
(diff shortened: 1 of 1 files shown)


MrNeRF (changes requested):
I think this should be correct:

torch::Tensor gaussian(int window_size, float sigma)
{
    TORCH_CHECK(window_size % 2 == 1, "Window size must be odd.");

    auto gauss = torch::empty({window_size}, torch::kFloat32);

    const int   centre = window_size / 2;
    const float denom  = 2.0f * sigma * sigma;

    for (int x = 0; x < window_size; ++x)
    {
        const int   dist_i  = x - centre;
        const float dist_sq = static_cast<float>(dist_i * dist_i);
        gauss[x] = std::exp(-dist_sq / denom);
    }

    return gauss / gauss.sum();
}

MrNeRF (approved):
Lgtm, ty!

PR #160, opened 2025-07-05
Add loading support for other Colmap camera models

### LINKED ISSUE #45: camera models
Hi, are your going to support more camera models in  future? such as opencv model.

calvin-laurenson:
Closes #45. Doesn't do any undistortion and just throws an error if you try training on distorted/fisheye cameras. Stores the required camera instrinsics for 3DGUT. I also changed the conversion back and forth to fov to just store focal length as well as using the actual optical center values from Colmap.


calvin-laurenson:
Switching from fov to focal length has a lot more side effects than I expected. Evals currently fail because they don't use the built in image scaling and instead load from the already scaled images folder so the code doesn't know the images are being scaled. I have a fix in mind. I think it is a lot more pleasant to not have to deal with conversion between fov and focal length but I am fine reverting it.

MrNeRF:
@calvin-laurenson Thank you for this pr. 

I like how you deal with fov vs focal_length. The changes are looking very reasonable to me. You could add comments on the vars what _image_width/_height vs _camera_width/_height denotes. It wasn’t necessarily clearer before but this could be a good opportunity to improve it.

I am currently travelling so can’t run it. I have to trust that it won’t break the basic training. Could you please run the eval at least for the first two scenes and post it. It should roughly match the previous statistics.

altineller:
Thank you @calvin-laurenson for doing this.

@MrNeRF  I will test it as soon as it becomes available. :) The only reason I am not doing right now is calvin's branch is 3 commits behind, and the last commit the vcpkg added, and many things changed.

Also, I had some nvcc (?pxtas? related) errors while building, and since my system has cuda 12, I have replaced the libtorch library to libtorch-cxx11-abi-shared-with-deps-2.3.0+cu121.zip - and it compiles and runs. So it runs with cuda12 and torch 2.7.1.

I think this project will have many contributors.

calvin-laurenson:
The results seem generally lower than the README values but are consistent between the PR and master on my computer so it would be nice if someone could confirm there are no regressions. I have also pulled from main @altineller.

This PR
| Scene    | Iteration | PSNR          | SSIM         | LPIPS        | Num Gaussians |
| -------- | --------- | ------------- | ------------ | ------------ | ------------- |
| garden   | 30000     | 27.195826     | 0.858489     | 0.156860     | 1000000       |
| bicycle  | 30000     | 25.345762     | 0.777899     | 0.254601     | 1000000       |

On d9f9205d83949c887f4477c00d4a5fb9dd6a44de

| Scene    | Iteration | PSNR          | SSIM         | LPIPS        | Num Gaussians |
| -------- | --------- | ------------- | ------------ | ------------ | ------------- |
| garden   | 30000     | 27.187294     | 0.858248     | 0.156012     | 1000000       |
| bicycle  | 30000     | 25.376255     | 0.777392     | 0.256167     | 1000000       |

MrNeRF:
The results are consistent with the ones posted in the bounty issue. The reason the metrics differ from the readme is how mipnerf360 is evaluated. They use downscaled pngs (compare colmap.py in gsplat repo) and not the images from folder images_4 or images_2. I applied the official protocol to make sure that this implementation matches the gsplat results (it does!). However, I think that’s cheating on the metrics that‘s why I don’t support the official protocol for eval directly.

IMG_0235

MrNeRF (approved):
Lgtm. Ty!

altineller:
I am getting "Training thread error: Training on cameras with distortion is not supported yet."

I have a SIMPLE_RADIAL mode, does this mean we wont be able to use it?

MrNeRF:
Yep. The undistortion is not implemented yet. I think @calvin-laurenson wants probably to activate 3DGUT https://github.com/MrNeRF/gaussian-splatting-cuda/issues/112 ?

PR #173, opened 2025-07-18
python integration

MrNeRF:
That's experimental and will go behind a cmake feature flag before merging. The goal is to make this project scriptable by making python fully integrated. Just curious if it works on Windows.

@miriameng Do you have time to test it?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,7 +288,7 @@ endif()
 add_executable(downloader src/downloader.cpp)
 
 set_target_properties(downloader PROPERTIES
-    CXX_STANDARD 17
+    CXX_STANDARD 23
     CXX_STANDARD_REQUIRED ON
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake")
+set(CMAKE_TOOLCHAIN_FILE "$ENV{VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake")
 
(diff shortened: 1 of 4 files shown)


miriameng:
To be honest I am not sure if I like this PR. It bloats up the cmake config time, which is due to vcpkg setup time since we are building python3 from source, but I think 99% of the people don't get their python3 from vcpkg. By looking at the code quickly I couldn't understand why vcpkg python is necessary. Do you mind explaining why?

MrNeRF:
That's experimental and should be only build when an internal Python build is specified. The python build is likely only relevant for cutting release binaries to bundle it.

I am currently working on a tide integration so that Python is a first class citizen similar to blender, i.e. no pybind etc. The idea is to make all relevant variables accessible in order to replace cor instance the strategy by a python implementation or to access cameras etc during training.

Another example: I want to be able to take the point cloud and filter it anytime in a way that is not predefined or limited by the application's ui.

I hope that the idea is clear?

miriameng:
Thanks for the detailed explanation. Please do hide python3 behind a feature flag in vcpkg.json and conditionally enable it with VCPKG_MANIFEST_FEATURES. As for what you asked me originally, for some reason I am seeing build errors for vcpkg python3 on my Windows machine, so I can't really test it at the moment. I will need to spend some time to sift through the vcpkg logs to figure out why the build fails

miriameng:
To keep you posted, the vcpkg python3 build error I am seeing locally is due to my antivirus flagging an intermediate executable https://github.com/microsoft/vcpkg/issues/46525, asking about it

MrNeRF:
Thank you! I am working on a version so that the splat data can be inspected and changed after the application started. Should be finished soon. This will be pushed here on this branch.

MrNeRF:
Closed for now!

PR #180, opened 2025-07-22
handle with windows debug/release linking problems

dozeri83:
In windows, debug/release libraries  are different. For libtorch we must download those libraries to sperate folders.
I also added instructions for the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ cmake -B build -DCMAKE_BUILD_TYPE=Release
 cmake --build build --config Release -j
 
 ```
-Note: Building in Debug mode requires building debug python libraries (python3*_d.dll, python3*_d.lib) separately . 
+Note: Building in Debug mode requires building debug python libraries (python3*_d.dll, python3*_d.lib) separately.
 
 ### Troubleshooting Build Issues
 


CMakeLists.txt
```diff
@@ -8,7 +8,21 @@ project(gaussian_splatting_cuda LANGUAGES CUDA CXX C)
 
 # Setup
 get_filename_component(PROJ_ROOT_DIR "${CMAKE_CURRENT_SOURCE_DIR}" ABSOLUTE)
-set(Torch_DIR "${PROJ_ROOT_DIR}/external/libtorch/share/cmake/Torch")
+
+# Set Torch_DIR based on platform and build type
+if (WIN32)
+    if (CMAKE_BUILD_TYPE STREQUAL "Release")
+        set(Torch_DIR "${PROJ_ROOT_DIR}/external/release/libtorch/share/cmake/Torch")
```
MrNeRF, line 15: Someone mentioned to use cmake generator for this. Is that helpful for win?

dozeri83: I actualy dont know ..:) not a cmake wizard.

src/parameters.cpp
```diff
@@ -33,7 +33,14 @@ namespace gs {
                 char executablePathWindows[MAX_PATH];
                 GetModuleFileNameA(nullptr, executablePathWindows, MAX_PATH);
                 std::filesystem::path executablePath = std::filesystem::path(executablePathWindows);
-                std::filesystem::path parentDir = executablePath.parent_path().parent_path().parent_path();
+                std::filesystem::path parentDir = executablePath;
+                do {
+                    parentDir = parentDir.parent_path();
+                } while (!parentDir.empty() && !std::filesystem::exists( parentDir / "parameter" / filename));
+
+                if (parentDir.empty()) {
```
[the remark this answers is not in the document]
dozeri83: no need for that in my opinion

src/parameters.cpp
```diff
@@ -33,7 +33,14 @@ namespace gs {
...
+                    throw std::runtime_error("could not find "+(std::filesystem::path("parameter")/filename).string());
```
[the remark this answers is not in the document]
dozeri83: not that helpfull - since we look at the excecutable path and not the current path

CMakeLists.txt
MrNeRF: Fair. If someone on win needs that, we will get it :D

MrNeRF:
Great, thank you! @dozeri83 can I merge?

dozeri83:
@MrNeRF yes - tested it in my workstation (btw the master branch is broken right now - I got va_start compilation error from the previous PR)

miriameng:
This PR breaks the Windows build. Can we revert? 

`CMAKE_BUILD_TYPE` is not a meaningful variable in multi config generator like Visual Studio, whether to compile in release or debug is a build time decision instead of a cmake config time decision for Visual Studio. You will see the failure if you actually tried it. 

I filed an issue to complain this to pytorch before https://github.com/pytorch/pytorch/issues/155667.

MrNeRF:
I think the problem is with clion vs visual studio on windows. As long as this repo depends on libtorch (will be hopefully removed as dependency in a couple of weeks), we need to find a pragmatic working solution. 

Can we find a solution that serves both clion and visual studio users so that it builds more generally under windows? @dozeri83 @miriameng

dozeri83:
@MrNeRF  @miriameng  - the problem is with the definition of TORCH_DIR in the cmake, right? 
I am suggesting to introduce a new cmake argument lets say -DSPERATE_TORCH_DIRS , 
if defined it will do the current cmake implementation, if not - it will do the previous  implementation.

b.t.w - i hope that most users will try to build with the developer command prompt and not clion/vs

MrNeRF:
>  the problem is with the definition of TORCH_DIR in the cmake, right?
I am suggesting to introduce a new cmake argument lets say -DSPERATE_TORCH_DIRS 

If that solves the problem, I am good with this. Let's be pragmatic until the root cause is completely gone. Wdyt @miriameng ?

miriameng:
How about detecting whether the current generator is single config or multi config? If it's multi config, stick with the old method until pytorch fixes their cmake, and if it's single config like clion, use the new method?

MrNeRF:
> How about detecting whether the current generator is single config or multi config? If it's multi config, stick with the old method until pytorch fixes their cmake, and if it's single config like clion, use the new method?

Could you provide a pull request so we can test if it works for "everybody"?

dozeri83:
maybe if CMAKE_CONFIGURATION_TYPES ? 
and then 

if (WIN32 AND NOT CMAKE_CONFIGURATION_TYPES)
    if (CMAKE_BUILD_TYPE STREQUAL "Release")
        set(Torch_DIR "${PROJ_ROOT_DIR}/external/release/libtorch/share/cmake/Torch")
    elseif (CMAKE_BUILD_TYPE STREQUAL "Debug")
        set(Torch_DIR "${PROJ_ROOT_DIR}/external/debug/libtorch/share/cmake/Torch")
    else()
        message(FATAL_ERROR "libtorch binaries only available for Debug and Release on Windows. Current build type: '${CMAKE_BUILD_TYPE}'")
    endif()
else()
    set(Torch_DIR "${PROJ_ROOT_DIR}/external/libtorch/share/cmake/Torch")
endif()

it works for me

PR #184, opened 2025-07-23
update cmake for windows multi config generator

miriameng:
@MrNeRF @dozeri83 can you try this out, this works for me


dozeri83:
@miriameng yes it works

MrNeRF:
Awesome! Thank you!

PR #185, opened 2025-07-24
added more verbosity to data set reader

dozeri83:
In case of failure to load I want to add more information to the user (including Colmap directory structures)


MrNeRF (approved):
Lgtm

PR #186, opened 2025-07-24
Gui first - initial implementation

MrNeRF:
This pr brings this repo closer to gui first!

1. Can start empty gui
2. Can load either ply or training set
3. Can switch between ply and training set
4. Headless mode works for pure training


MrNeRF:
So this is still a draft and has some bugs but I am interested if the file directory browser works on windows for now?

dozeri83:
> So this is still a draft and has some bugs but I am interested if the file directory browser works on windows for now?

GetActiveWindow - does the job. but it is a little bit ugly:
image
 
 
 did not try the other method that claude suggested because it did not compiled in the first try and I dont have time at the moment.

MrNeRF:
Thanks. I think it is okay currently as long as it works. Beautifying can be done later.

dozeri83:
@MrNeRF - started training and it crashed. ( in metrics.cpp line 308 - _params looks like points to garbage memory.

MrNeRF:
Yes. It has currently some issues. I will remove the draft when it is ready.

MrNeRF:
I am closing this in favor of https://github.com/MrNeRF/gaussian-splatting-cuda/pull/191. I reimplemented this pr in smaller steps.

PR #187, opened 2025-07-24
Implement 3DGUT

calvin-laurenson:
Forward pass works but training doesn't. Opacity of gaussians just decreases (due to regularization), the gradients from the GUT backward pass are essentially zero. Help would be appreciated. If the backward pass is fixed I will try to implement fisheye and distortion.


MrNeRF:
Thanks. I will take another look over the coming days (likely tomorrow). In  the meantime help is appreciated :)

PR #195, opened 2025-07-25
introducing a window manager

MrNeRF:


MrNeRF:
removing more stuff from viewer.cpp

PR #197, opened 2025-07-25
README: update gfortran to 14 as well

miriameng:
Probably not directly related, but I moved up to gcc 14 because of this project, and compilation for other projects were breaking because gfortran was still at the default 13, so I think it's a good idea to mention updating gfortran to 14 as well in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,15 +176,17 @@ gcc --version  # Should be 14+ for full C++23 support
 
 # Install GCC 14 (Ubuntu 24.04+)
 sudo apt update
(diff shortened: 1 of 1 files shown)


MrNeRF:
As it is unrelated, maybe it good to warn that's not necessary but might be helpful if other compilations break? Otherwise it seems like a requirement.

MrNeRF:
Pardon, pressed the wrong button.

MrNeRF:
Ok, thx!

PR #202, opened 2025-07-26
Added preload-to-ram feature and --preload-to-ram flag

MotivaCG:
Keeping images in ram can speed up the process (disabled by default)


MrNeRF:
When the dataset is small, then you can already load it into vram. Wdyt?

MotivaCG:
> When the dataset is small, then you can already load it into vram. Wdyt?

I didnt see that in the code. How can I keep them in GPU?

Anyway in my test with 105 images I can avoid the load image call per camera and step, having them on RAM introduces a ↑15% performance gain. It makes sense to me to introduce it since having 12/16 Gb of video RAM and 64 of system RAM is a common setup.

MrNeRF:
The bounty is due in a week. I certainly know there will be changes in the data loading so this might become obsolete. The changes are minimal in this pr anyway. You would need to push them onto the cuda device. That is like a single .to(device) change.

Did you apply clang-format?

MotivaCG:
Conflicts resolved

MrNeRF:
ok, lgtm! Thank you!

PR #203, opened 2025-07-26
Fix radii2D bug in MCMC

MotivaCG:
It was not properly updated


MrNeRF:
Thx! Note, that this is not a bug. It is simply not required in mcmc.

MotivaCG:
It is required for the pruning process 😅

MrNeRF:
There is no pruning in mcmc.

MotivaCG:
> There is no pruning in mcmc.

I introduced pruning in the other branch so we can close this pool request, but the change will be required for the masking branch.

MrNeRF:
If you make bigger changes to mcmc, please create another strategy. You can see here https://github.com/MrNeRF/gaussian-splatting-cuda/pull/201 how to do it.

PR #208, opened 2025-07-27
implement crop box mechanism

dozeri83:
1. implemented crop box mechanism
2. render crop box when "show check box" is enable
2. render gaussian out of crop box when "use crop box" is enable

Screenshot 2025-07-27 181019


src/visualizer/render_bounding_box.cpp
```diff
@@ -0,0 +1,173 @@
...
+            glm::mat4 mvp = projection * view * world2BBox_;
+
+            shader_->set_uniform("u_mvp", mvp);
+            shader_->set_uniform("u_color", color_);
+
+            // Bind VAO and draw
+            glBindVertexArray(VAO_);
+
+            glDisable(GL_DEPTH_TEST);
+
```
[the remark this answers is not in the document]
dozeri83: removed glDisable(GL_DEPTH_TEST)

include/visualizer/render_bounding_box.hpp
```diff
@@ -0,0 +1,63 @@
...
+        ~RenderBoundingBox();
+
+        // Set the bounding box from min/max points
+        void setBounds(const glm::vec3& min, const glm::vec3& max) override;
+
+        // Initialize OpenGL resources
+        void init();
+
+        // Enable/disable bounding box rendering
+        // void setVisible(bool visible) { visible_ = visible; }
```
MrNeRF, line 26: Please either implement this or remove it.

src/rasterizer.cpp
```diff
@@ -48,7 +48,8 @@ namespace gs {
         float scaling_modifier,
         bool packed,
         bool antialiased,
-        RenderMode render_mode) {
+        RenderMode render_mode,
+        const gs::BoundingBox* bounding_box) {
```
MrNeRF, line 52: This is okay for now. 
I would like to separate the training from viewer rendering eventually. The training performance should be as high as possible and the training implementation should be uncluttered and minimal.

However, we don't have that yet implement so let's keep it in mind.

MrNeRF (approved):
Ok, that looks great. It needs tuning and finally my work on orienting the splat (which is currently on a branch).

include/visualizer/render_bounding_box.hpp
MrNeRF: We can do this in a subsequent pr. I will just merge it.

PR #212, opened 2025-07-27
Use string keys for optimizer state to fix type mismatch

MotivaCG:
The optimizer state map uses std::string as its key type, but several parts of the code were attempting to use raw TensorImpl* pointers for lookups, insertions, and deletions.

This caused a compilation failure on stricter compilers (e.g., MSVC) due to a type mismatch, as there is no implicit conversion from a pointer to std::string.

This change preserves the original logic of using the tensor's memory address as a unique identifier but corrects the implementation. A new helper function, gs::get_tensor_key(), is introduced to encapsulate the pointer-to-string conversion. It has been placed in a new core/gs_utils.hpp header, which can be used for similar small, shared utility tools in the future.

All optimizer state map accesses in selective_adam.cpp and mcmc.cpp have been updated to use this helper, resolving the compilation error on stricter than gcc compilers.

### DIFF
--- a/include/core/gs_utils.hpp
+++ b/include/core/gs_utils.hpp
@@ -1,0 +1,13 @@
+#pragma once
+
+#include <string>
+#include <sstream>
+
+namespace gs {
+inline std::string get_tensor_key(const void* ptr) {
+    std::stringstream ss;
+    ss << ptr;
+    return ss.str();
+}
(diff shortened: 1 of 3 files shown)


MotivaCG:
It was related to a wrong version of Torch!

PR #218, opened 2025-07-29
Refactor dataloader

MrNeRF:
Single interface for data access. It encapsulates all dataloading. Dataloading is no longer sprinkled across the entire application!

### DIFF
--- a/src/loader/formats/transforms.cpp
+++ b/src/loader/formats/transforms.cpp
@@ -13,6 +13,10 @@ namespace gs::loader {
 
     namespace F = torch::nn::functional;
 
+    // Constants for random point cloud generation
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -92,17 +92,12 @@ find_package(spdlog REQUIRED)
 # Python packages
 find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
 
-if(WIN32)
-    set(CMAKE_CUDA_RUNTIME_LIBRARY "Shared") # when using INTEROP in windows static and dynamic symbols are mixed
-endif()
```
miriameng, line 97: This is a merge error, right? This fix from @dozeri83 is necessary for cuda gl interop to work on windows

PR #220, opened 2025-07-29
Vis module

MrNeRF:
- Unified Public API (include/visualizer/visualizer.hpp)
- Improved Module Structure:
src/visualizer/
├── visualizer.cpp 
├── gui/   
├── scene/
├── rendering/
├── input/
├── training/
└── window/


MrNeRF:
So we have now the loader and visualizer module. 
What remain in the following up work:
1. Make event system simpler. Too complex and hard to grasp at the moment
2. GUI elements should not be mixed with visualizer logic
3. Create tool concept to fit in the tools like crop box

@dozeri83 Your crop box fixes were automatically merged. I am only interested if it works on windows. The rest should be good.

PR #229, opened 2025-07-30
Shader wchar compilation bug fix

dozeri83:
fix wchar convertion compilation bug in windows

"cannot convert const wchar_t* to const char_t*"

apperenatly MSVC  is strict with this conversion.
@MrNeRF  - can you test compilation ? and functionality that uses this code?


miriameng:
Does #227 solves the compilation problem for you?

MrNeRF:
Yeah, this seems minimal: https://github.com/MrNeRF/gaussian-splatting-cuda/pull/227. Maybe it is solved?

dozeri83:
> Does #227 solves the compilation problem for you?

yes. it looks way simpler. closing this PR

PR #234, opened 2025-07-31
Lambda compilation bug fix

dozeri83:
apparently MSVC is stricter than gcc regarding lambda expressions in functions   signature


MrNeRF:
This is very similar: https://github.com/MrNeRF/gaussian-splatting-cuda/pull/235. Which one should we take?

PR #235, opened 2025-07-31
Fix MSVS compilation in windows (trick with std::function)

MotivaCG:


MrNeRF:
This is very similar: https://github.com/MrNeRF/gaussian-splatting-cuda/pull/234 
This one here is maybe a bit more elegant. Wdyt? @dozeri83 ?

dozeri83:
> This is very similar: #234 This one here is maybe a bit more elegant. Wdyt? @dozeri83 ?

I think that one would benefit from a comment about why we this this peculiar trick (like I wrote in my PR) but otherwise it does not matter

PR #241, opened 2025-08-02
Wasd speed

dozeri83:
1. increase wasd speed with ctrl +/- .

2. added a new speed change event  and display user about this change for short period of time

3. increase default speed by a factor of 2

image


MrNeRF:
This looks awesome 🔥

src/visualizer/gui/gui_manager.cpp
```diff
@@ -145,6 +153,104 @@ namespace gs::gui {
...
+    void GuiManager::renderSpeedOverlay() {
+        // Check if overlay should be hidden
+        if (speed_overlay_visible_) {
+            auto now = std::chrono::steady_clock::now();
+            if (now - speed_overlay_start_time_ >= speed_overlay_duration_) {
+                speed_overlay_visible_ = false;
+            }
+        }
+
+        if (!speed_overlay_visible_) {
```
MrNeRF: Either if else or you simply check in the beginning early return

dozeri83:
not backed enough

MrNeRF:
Huh?

dozeri83:
@MrNeRF  - I am having trouble deciding were   renderSpeedOverlay()
should be. currently its in gui_manger but looks like gui_manger  is supposed to be clean.

When I tried to move it, I got into a big mess.  any thoughts?

dozeri83:
> Small suggestion for early return. Further, the overlay could go into an overlay component to not clutter the gui_manager. Either you do it here, or I will refactor later

yeah I was thinking about it. but got stuck.

src/visualizer/gui/gui_manager.cpp
dozeri83: fixed it

Issue #135, opened 2025-06-23
💰 Bounty #001 – $300 for Cutting Training Time by 50%


MrNeRF:
Help us **cut training time in half** and earn **$300**!  
This bounty challenges you to speed up training by at least **50%**, without compromising results.

> 🔁 The winning implementation will be merged, and the repository will switch to a **GPLv3 license**.

---

## 🧾 Rules

1. **Fork the repository**, including all branches, starting from:  
   👉 [`bounty_001` branch](https://github.com/MrNeRF/gaussian-splatting-cuda/tree/bounty_001)

2. **Use this script to benchmark performance**:  
   🕒 [`timing_mipnerf360.sh`](https://github.com/MrNeRF/gaussian-splatting-cuda/blob/bounty_001/scripts/timing_mipnerf360.sh)

3. **Apply your speed-up** and submit a **pull request**.

4. **Do not degrade quality** – final metrics (e.g., PSNR/SSIM) must remain consistent with the baseline.

5. Target a **minimum of 50% speed-up** compared to the current implementation.

6. Use the official **MipNeRF360 dataset**:  
   📦 [360_v2.zip](http://storage.googleapis.com/gresearch/refraw360/360_v2.zip)

7. **Deadline:** July 31, 2025  
   ⏳ All results will be reviewed and the winner announced shortly after.

8. If multiple entries achieve similar speed-ups:
   - The **cleanest** and **earliest** implementation will win.
   - Final decision rests with the repo maintainer.

9. Your pull request must include a **clear summary** of:
   - What you optimized
   - How you achieved the speed-up

10. You **must not use any code that is licensed under a non-permissive license**  
    (e.g., GPLv2-only, SSPL, or similar). Only permissively licensed code (MIT, BSD, Apache 2.0, etc.) is allowed.

11. By participating, you agree to release your submission under the **GPLv3 license**.

---

## 🏆 Prize

- 💸 **$300** to the author of the winning pull request.
- 🧠 Bonus: Your name (or alias) will be featured in the repository’s README.

---

## 📣 Discussion

Join our Discord server to discuss ideas, ask questions, or get help:  
👉 [https://discord.gg/6FaYg29MN7](https://discord.gg/6FaYg29MN7)

---

Good luck, and happy hacking! 🚀

P.S.: I will update this issue with the current metrics and runtime on my side measure with a RTX 4090!

MrNeRF:
For inspiration here are some pointers:
[...]

PR #245, opened 2025-08-03
Bounty Solution (2.4x speedup)

fhahlbohm:
This is a solution for [Issue #135](https://github.com/MrNeRF/gaussian-splatting-cuda/issues/135).
It implements improvements that reduce the training from the original 49m 50s to **20m 30s**, a speedup by over **2.4x**.
To get the time below 20 minutes, I also added an *optional* trick that sacrifices some quality to get the time down to *19m 27s*.

## Changes

The key changes can be summarized as follows:

### New Rasterizer
- Instead of the modular gsplat implementation, the new implementation tries to use as few CUDA kernels as possible.
- I redid all the math for the forward and backward pass to skip all unnecessary computations.
- All activation functions of the Gaussians' parameters are fused into the respective kernels of the forward and backward pass.
- I implemented an improved version of the kernel for the blending backward pass based on Taming 3DGS.
- I separated the sorting into depth and tile sorting passes similar to how it is done in Splatshop.
- I integrated tile-based culling and load balancing based on StopThePop.
- Also see `fastgs/rasterization/README.md`

### MCMC Densification
- After every training iteration, MCMC adds noise to the position of each Gaussian. I fused the required operations into a single kernel.
- The functions for adding and relocating Gaussians could be fused too, but because they do not happen that often during training, the benefit would be small. Therefore, I did not do so and only fixed a few inefficiencies in the libtorch implementation.
- Gaussians are now also "dead" when the quaternion used to represent their rotation has a squared norm below `1e-8` as optimization and rendering would then be numerically unstable anyway. The new rasterizer also culls such Gaussians during training.
- Like most 3DGS implementations, the old implementation used to call the libtorch equivalent of `torch.cuda.empty_cache()` after every densification step to solve memory fragmentation issues. This slightly slows down training as temporary memory is first freed and then reallocated. Enabling the torch memory allocator's setting `expandable_segments:True` allows to get rid of this.
- I removed the `abs()` calls in the opacity and scale regularization losses as they are unnecessary because the Gaussians' opacity and scale are always positive after applying the respective activations.

### Optimizer
- The libtorch Adam implementation is slow because it launches all operations for each update step as separate CUDA kernels. I implemented a fused version of the Adam optimizer step that is significantly faster.
- During the first 1000 iterations, the higher-degree SH coefficients will not be used and therefore always have zero gradients. Therefore, the optimizer step for those can be skipped for a small speedup.
- (optional) Between iteration 1000 and 25000, the expensive update for higher-degree SH coefficients can be batched over two iterations to achieve a small speed up. This is disabled by default as it seems to slightly reduce quality in my tests.

### Data Loading
- The torch dataloader was re-created after every epoch. Now only the random sampler is reset, which avoids unnecessary overhead.
- Image normalization to the [0, 1] range was done on the CPU. Doing it after uploading to the GPU is much faster.
- Images were always loaded from disk, which is actually not that bad with a torch dataloader that uses multiple worker threads, but it has some overhead. Now, heuristics are used to determine whether the dataset fits into GPU memory. If yes, images are cached in VRAM.
- Since the VRAM needed for storing the view matrices and camera positions for all views is negligible, they are now precomputed and stored in VRAM. Example: 10k views -> 0.76 MB (could be lower as I store the full 4x4 view matrix instead of just the relevant 3x4 part)

## Evaluation

Timings are quite consistent across runs as long as one does not touch the PC during benchmarking.
```
=========================================
BENCHMARK SUMMARY
=========================================
garden         : 00h 02m 28s
bicycle        : 00h 02m 14s
stump          : 00h 02m 20s
bonsai         : 00h 03m 17s
counter        : 00h 03m 35s
kitchen        : 00h 03m 42s
room           : 00h 02m 54s
-----------------------------------------
Total time: 0h 20m 30s
=========================================
```
With the optional SH trick enabled, training is slightly faster and the total is below 20 minutes:
```
=========================================
BENCHMARK SUMMARY
=========================================
garden         : 00h 02m 18s
bicycle        : 00h 02m 06s
stump          : 00h 02m 11s
bonsai         : 00h 03m 07s
counter        : 00h 03m 25s
kitchen        : 00h 03m 35s
room           : 00h 02m 45s
-----------------------------------------
Total time: 0h 19m 27s
=========================================
```

Here are the average quality metrics that I got with the original implementation and the new one without and with the optional SH trick enabled.
Note that metric computation still uses gsplat to render the images. This highlights that the new rasterizer used during training is not required during inference to obtain high quality.

| Version              | PSNR      | SSIM     | LPIPS    |
|----------------------|-----------|----------| -------- |
| original             | 29.235601 | 0.879749 | 0.227879 |
| solution             | 29.277397 | 0.879477 | 0.228021 |
| solution w/ SH trick | 29.226835 | 0.879472 | 0.228022 |

I spent a lot of time on making sure the new implementation does not reduce quality.
Note that the results in this table do not confirm that the new implementation is strictly better than the original one.
There are small differences in terms of how things are computed in a numerically stable manner in the rasterizer.
However, this did not make a measurable difference in practice.
The main problem when repeatedly testing both the original and the new implementation is that metrics fluctuate between runs making it impossible to tell which implementation achieves better quality.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ configure_file(
 )
 
 add_subdirectory(gsplat)
+add_subdirectory(fastgs)
 
 # =============================================================================
 # HOST LIBRARY - Compiled with g++ (fast!)
@@ -83,12 +84,15 @@ set(HOST_SOURCES
         src/trainer.cpp
         src/argument_parser.cpp
         src/rasterizer.cpp
+        src/fast_rasterizer.cpp
         src/metrics.cpp
         src/rasterizer_autograd.cpp
+        src/fast_rasterizer_autograd.cpp
         src/viewer.cpp
         src/external/tinyply.cpp
         src/bilateral_grid.cpp
         src/selective_adam.cpp
+        src/fused_adam.cpp
 )
 
 add_library(gaussian_host STATIC ${HOST_SOURCES})
@@ -99,6 +103,7 @@ target_include_directories(gaussian_host
         ${CMAKE_CURRENT_SOURCE_DIR}/include
         ${CMAKE_CURRENT_BINARY_DIR}/include  # For generated config.h
         ${CMAKE_CURRENT_SOURCE_DIR}/gsplat  # Add gsplat headers
+        ${CMAKE_CURRENT_SOURCE_DIR}/fastgs  # Add fastgs headers
         ${CUDAToolkit_INCLUDE_DIRS}         # Add CUDA headers for interop
         PRIVATE
         ${Python3_INCLUDE_DIRS}
@@ -117,6 +122,7 @@ target_link_libraries(gaussian_host
         taywee::args
         Python3::Python
         gsplat_backend  # Link to gsplat
+        fastgs_backend  # Link to fastgs
         ${OPENGL_LIBRARIES}
         CUDA::cudart    # Add CUDA runtime for interop
 )
@@ -251,6 +257,7 @@ target_include_directories(${PROJECT_NAME}
         ${CMAKE_CURRENT_SOURCE_DIR}/include
         ${CMAKE_CURRENT_BINARY_DIR}/include  # For generated config.h
         ${CMAKE_CURRENT_SOURCE_DIR}/gsplat
+        ${CMAKE_CURRENT_SOURCE_DIR}/fastgs
         ${Python3_INCLUDE_DIRS}
         ${CUDAToolkit_INCLUDE_DIRS}
         ${OPENGL_INCLUDE_DIRS}
@@ -261,6 +268,7 @@ set(MAIN_LINK_LIBRARIES
         gaussian_host
         gaussian_kernels
         gsplat_backend
+        fastgs_backend
         Python3::Python
         ${OPENGL_LIBRARIES}
         CUDA::cudart
@@ -330,6 +338,7 @@ endfunction()
 configure_build_type(gaussian_host)
 configure_build_type(gaussian_kernels)
 configure_build_type(gsplat_backend)
+configure_build_type(fastgs_backend)
 if(CUDA_GL_INTEROP_FOUND AND TARGET gaussian_visualizer)
     configure_build_type(gaussian_visualizer)
 endif()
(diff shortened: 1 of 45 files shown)


MrNeRF:

MrNeRF:
image

MrNeRF:
Final eval:
| scene   |   iteration |    psnr |     ssim |    lpips | num_gaussians   |
|:--------|------------:|--------:|---------:|---------:|:----------------|
| garden  |       30000 | 27.2306 | 0.858708 | 0.156047 | 1,000,000       |
| bicycle |       30000 | 25.4351 | 0.777493 | 0.255592 | 1,000,000       |
| stump   |       30000 | 26.7973 | 0.790044 | 0.263982 | 1,000,000       |
| bonsai  |       30000 | 32.6185 | 0.948317 | 0.249532 | 1,000,000       |
| counter |       30000 | 29.2851 | 0.917415 | 0.241510 | 1,000,000       |
| kitchen |       30000 | 31.3604 | 0.933028 | 0.155878 | 1,000,000       |
| room    |       30000 | 32.2236 | 0.931001 | 0.273806 | 1,000,000       |
| mean    |       30000 | 29.2786 | 0.879429 | 0.228050 | 1,000,000       |

hugoycj:
@fhahlbohm Hi! Amazing work on the 2.4x speedup - that's really impressive!
I'm curious about the individual contribution of the New Rasterizer to the overall performance gain. Since you implemented multiple optimizations (New Rasterizer, MCMC, fused Adam, data loading), do you happen to know roughly how much of the speedup came specifically from the rasterizer changes versus the other improvements?

Thanks for sharing this excellent work!

fhahlbohm:
Thanks for the kind words!

> I'm curious about the individual contribution of the New Rasterizer to the overall performance gain. Since you implemented multiple optimizations (New Rasterizer, MCMC, fused Adam, data loading), do you happen to know roughly how much of the speedup came specifically from the rasterizer changes versus the other improvements?

Yes, a small write-up with a bunch of experiments beyond the scope of this bounty is coming soon. Together with a PyTorch API for the rasterizer for the PyTorch-based research community.

andorxornot:
Wow, that's incredibly cool. Quite a lot of changes and most of them are pure optimizations. The papers that will give even more speed have not yet been implemented. Crazy!

PR #246, opened 2025-08-03
Precise tile intersection from FlashGS

robinerd:
As a smaller contribution on performance (part of the bounty 001), I updated the tile intersection logic to use a more precise ellipsoid intersection, as done in the FlashGS paper. This results in fewer tiles processed in the rasterization.

I tested it only on my RTX 4060, and get a 10-15% training speedup on the bicycle scene. Haven't run extensive benchmarking but hopefully this fairly standalone change can be useful together with other optimizations as well.

**Usage:**
Run the timing script with --use-precise-intersection flag to enable the new approach, or directly to gaussian_splatting_cuda. By default if is OFF.
ould be 

**NOTE:**
I've left a debug statistic print for tile intersections for now which could be interesting while evaluating the change. If it ends up being used I would say this should just be always on without a flag, and ofc remove those debug logs.

**NOTE 2:**
The code is based on FlashGS codebase but not an exact copy. Their code is released under MIT license. Should make sure to reference them properly if adding.
See: https://github.com/InternLandMark/FlashGS


melonedo:
Oh 10-15% is no small improvements! The baseline implementation spends a lot of time outside gaussian splatting kernels as is seen in my PR. This patch will likely bring at least 20% improvement to the rasterization process.

robinerd:
> Oh 10-15% is no small improvements! The baseline implementation spends a lot of time outside gaussian splatting kernels as is seen in my PR. This patch will likely bring at least 20% improvement to the rasterization process.

Ah cool, will be interesting to see. I'm not sure how big my speed-up will be on the 4090 as the bottlebeck could be different. 4090 has about 5.5x more TFLOPS (vs my 4060) but just 3.7x more memory bandwidth. From my measuring with nsight the biggest bottleneck on cuda side is global memory bandwidth, and there's a lot of uncoalesced global memory accesses (mainly reading the gaussian params). Memory access is one thing the FlashGS paper also improved on and their paper cites a big speedup vs the gsplat, but their code has no backward kernels so I didn't manage to get it implemented.

melonedo:
That's exactly why I say 10-15% is not small: My implementation resolves the memory coalescing issue and has not touched the tiling logic.



---Original---
From: "Robin ***@***.***&gt;
Date: Sun, Aug 3, 2025 16:10 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [MrNeRF/gaussian-splatting-cuda] Precise tile intersection fromFlashGS (PR #246)


robinerd left a comment (MrNeRF/gaussian-splatting-cuda#246)
  
Oh 10-15% is no small improvements! The baseline implementation spends a lot of time outside gaussian splatting kernels as is seen in my PR. This patch will likely bring at least 20% improvement to the rasterization process.
  
Ah cool, will be interesting to see. I'm not sure how big my speed-up will be on the 4090 as the bottlebeck could be different. 4090 has about 5.5x more TFLOPS but just 3.7x more memory bandwidth. From my measuring with nsight the biggest bottleneck on cuda side is global memory bandwidth, and there's a lot of uncoalesced global memory accesses (mainly reading the gaussian params). Memory access is one thing the FlashGS paper also improved on and their paper cites a big speedup vs the gsplat, but their code has no backward kernels so I didn't manage to get it implemented.

MrNeRF:

robinlindh:
Nice, then it sounds like this could stack nicely. Looking forward to see metrics later

MrNeRF:
Ok, here is the runtime on my system.

Screenshot from 2025-08-03 17-16-26

MrNeRF:
@robinerd Thank you Robin for contributing. I will close this pr. I think we have a solution with load balancing.

robinerd:
> @robinerd Thank you Robin for contributing. I will close this pr. I think we have a solution with load balancing.

Yeah, the implementation in the winning PR looks pretty solid, not applicable for my PR. Thanks for a nice initiative too, very fun to get into. And now I'm familiar with this code base for future ;) Cheers

MrNeRF:
Nice features and improvements are very welcome :)

gsplat/IntersectTile.cu
```diff
@@ -114,6 +164,138 @@ __global__ void intersect_tile_kernel(
...
+    tile_max.y = min(max(0, (uint32_t)ceil(tile_y + tile_radius_y)), tile_height);
+
+    // Get conic and opacity for precise intersection testing
+    const float3 conic = {
+        static_cast<float>(conics[idx * 3]),
+        static_cast<float>(conics[idx * 3 + 1]),
+        static_cast<float>(conics[idx * 3 + 2])
+    };
+    const float opacity = static_cast<float>(opacities[idx]);
+    const float power = -0.5f * log(opacity + 1e-8f);  // Convert to power parameter
```
clockmh, line 231: Why power calculated in this way which is different from that in flashgs

PR #247, opened 2025-08-03
Bounty submission: 70% speedup

darshanmakwana412:
This is the submission for bounty 001. This achieves a speedup of 70% on a 4090 compared to the baseline. The improvements done are as follows

- Caches the images dynamically on gpu and cpu
- Removes the cat of the sh
- Adds taminggs kernels + FlashGS accurate tile intersection
- uses the fused adam with a bias corrected kernel without the visibility mask
- precomp and cache the projmat and viewmats
- (optional) adds support for udpating the shN after every 16 steps (but degrades performance)
- (optional) Implements [DashGS](https://dashgaussian.github.io/) dft based frequency scheduler (but metrics are slightly worse)

optional implementations were turned off while computing the 70% speedup

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,6 +68,7 @@ configure_file(
 )
 
 add_subdirectory(gsplat)
+add_subdirectory(taminggs)
 
 # =============================================================================
 # HOST LIBRARY - Compiled with g++ (fast!)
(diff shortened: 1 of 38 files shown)


MrNeRF:

MrNeRF:
This pr contains INRIA code which is a disqualification according to the rules. I am sorry :(

darshanmakwana412:
Hello, I apologize for a misunderstanding on my part. I have just realized that the [license](https://github.com/humansensinglab/taming-3dgs/blob/main/submodules/diff-gaussian-rasterization/LICENSE.md) governing the kernels is different from the [license](https://github.com/humansensinglab/taming-3dgs/blob/main/LICENSE.md) of the repo which is MIT license

MrNeRF:
I admit it is really confusing and unfortunate if you never deal with licenses. 

The repo owner can only relicense the diff but not the base code. That means, the diff must be applied on top of a free licensed implementation. In general, that requires the modification of the gsplat code base.

PR #263, opened 2025-08-06
Start looking at current folder for the parameters (better deployment in Windows)

MotivaCG:
Instead of looking at the parent level it searches in the current folder and goes up level after level as before.

It allows the user to place the .xml in the same folder as the executable without breaking compatibility with the current folder tree.


MrNeRF:
lgtm. Ty!

PR #276, opened 2025-08-08
fix compile error if using ENABLE_CUDA_GL_INTEROP=OFF

robinerd:
Fix for std pointer types if running cmake with -DENABLE_CUDA_GL_INTEROP=OFF They are probably included indirectly otherwise.

For me I were already disabling the interop earlier in the cmake file manually since I had issues with it on WSL (windows + Ubuntu 24).
This flag from earlier PR makes it easier but made it fail to compile.
This fixes it for me, hope it helps others too.

### DIFF
--- a/src/visualizer/rendering/renderer.hpp
+++ b/src/visualizer/rendering/renderer.hpp
@@ -4,6 +4,7 @@
 #include "internal/viewport.hpp"
 #include "rendering/framebuffer.hpp"
 #include "rendering/shader.hpp"
+#include <memory>
 
 #ifdef CUDA_GL_INTEROP_ENABLED
 #include "rendering/cuda_gl_interop.hpp" // Add path and extension


MrNeRF:
Alright, thank you!
Btw, do you have issue with the camera controls when interop is disabled? I just noticed something. The orbiting (middle button pressed) does not work for me when interop is disabled.

robinerd:
> Alright, thank you! Btw, do you have issue with the camera controls when interop is disabled? I just noticed something. The orbiting (middle button pressed) does not work for me when interop is disabled.

Camera controls felt extremely weird for me since I tried the viewer first time. I have not tried it  since the bounty branch though and there's been some work on master too so I'll  get back to you after trying that. But yeah earlier mouse panning was super slow, and the rotation felt inverted. My dataset also has a different up vector so that added to the rotation weirdness. Was a little hard to narrow it down to exactly what. Middle mouse orbiting was not in that version I thihk, or just didn't work.

MrNeRF:
Yeah, the initial control were weird. Still, they need some more love.

PR #277, opened 2025-08-08
Fixing ADC with new rasterizer

MrNeRF:


MrNeRF:
TODOs: 
- use empty tensor for mcmc (done)
- use fused adam in default strategy (done)
- disable densification_info after densification stop (done)
- bring back commented out blending (done)
- cleanup (done)

PR #278, opened 2025-08-08
Add differentiable depth outputs to fast rasterizer

fhahlbohm:
See title.

Not tested but the math should be correct.
Will probably cause a slowdown and also increases VRAM usage.

Similar to gsplat, the depth gradient is not propagated back onto the Gaussian means. Could add this in the future.


MrNeRF:
Finally closing, it is implemented :)

PR #279, opened 2025-08-08
Scene panel improvements

dozeri83:
1. show dataset images in scene panel instead of directory tree.
2. implemented right click + go to cam.


MrNeRF:
Lgtm, Thank you!

PR #283, opened 2025-08-10
Data config bug fix

dozeri83:
1. fixed dead reference bug that cause --resolution not to work
2. added options for resolution since load_image only handle few options

resolution is a bad name btw (this parameter only responsible for scaling - maybe resolution scale ?)


MrNeRF:
How about downsample_factor, resize_factor, or resize_factor. 
I think Downsample_factor is the best term because it is not upscaling but the word is very long.

src/argument_parser.cpp
```diff
@@ -74,6 +73,17 @@ namespace {
             ::args::Flag save_depth(parser, "save_depth", "Save depth maps during training", {"save-depth"});
             ::args::Flag skip_intermediate_saving(parser, "skip_intermediate", "Skip saving intermediate results and only save final output", {"skip-intermediate"});
 
+            ::args::MapFlag<std::string, int> resolution(parser, "resolution",
+                                                         "Scale resolution by this factor. Options: auto, -1, 2, 4, 8 (default: auto)",
```
MrNeRF: Why not just use simply 1 instead of auto?
The options would be 2,4,8 -> default is 1.

dozeri83:
> How about downsample_factor, resize_factor, or resize_factor. I think Downsample_factor is the best term because it is not upscaling but the word is very long.

I will change it to resize_factor

dozeri83:
> > How about downsample_factor, resize_factor, or resize_factor. I think Downsample_factor is the best term because it is not upscaling but the word is very long.
> 
> I will change it to resize_factor

@MrNeRF  - renaming it changes alot of files. should i still do it?

I will only change it in the interface.

MrNeRF:
That's fine if there are many file changes. Should be preferably consistent, isn't it?

dozeri83:
> That's fine if there are many file changes. Should be preferably consistent, isn't it?

finished. tested it partially on some dataset.

PR #287, opened 2025-08-11
background color tool

MrNeRF:
Add background color tool to viewer. 
Important: No impact on training, alter simply the background color in the viewer!

Screenshot from 2025-08-11 14-20-57

### DIFF
--- a/src/visualizer/gui/panels/main_panel.cpp
+++ b/src/visualizer/gui/panels/main_panel.cpp
@@ -76,7 +76,8 @@ namespace gs::gui::panels {
             events::ui::RenderSettingsChanged{
                 .fov = std::nullopt,
                 .scaling_modifier = config->scaling_modifier,
-                .antialiasing = std::nullopt}
+                .antialiasing = std::nullopt,
+                .background_color = std::nullopt}
                 .emit();
         }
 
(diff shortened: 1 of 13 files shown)


src/visualizer/gui/panels/main_panel.cpp
```diff
@@ -76,15 +76,17 @@ namespace gs::gui::panels {
             events::ui::RenderSettingsChanged{
                 .fov = std::nullopt,
                 .scaling_modifier = config->scaling_modifier,
-                .antialiasing = std::nullopt}
+                .antialiasing = std::nullopt,
+                .background_color = std::nullopt}
```
dozeri83, line 80: when naming background_color - should we distinguish between training and rendering ?

dozeri83:
tested it - feels pretty good. 
my only problem - 
1. if the user enable bg tool, changed the color, and then disabled bg tool - I think we want to keep the new bg_colour
and not reset. 
2.  There should be reset button. ( else I have to manually change it which is annoying)

MrNeRF:
Makes sense. Will add. Ty!

src/visualizer/gui/panels/main_panel.cpp
MrNeRF: Don't think so.  I is clearly separated by namespaces. The training will be also moved into a training module.

MrNeRF:
Reset and persistent bg colors are added!

PR #290, opened 2025-08-12
Point cloud rendering

MrNeRF:
https://github.com/user-attachments/assets/10377d1b-a68d-4b12-861b-73be4cc0864a

### DIFF
--- a/src/visualizer/rendering/rendering_manager.cpp
+++ b/src/visualizer/rendering/rendering_manager.cpp
@@ -235,7 +235,8 @@ namespace gs::visualizer {
                 static_cast<int>(context.viewport_region->height));
         }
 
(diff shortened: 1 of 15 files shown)


src/visualizer/rendering/rendering_manager.cpp
```diff
@@ -404,6 +407,17 @@ namespace gs::visualizer {
             }
         }
 
+        // Check if point cloud mode or voxel size changed
+        static bool prev_point_cloud_mode = false;
```
dozeri83: tracking all those GUI changes sucks ...
 
lets create an event instead -
(just like sceneload event - create sceneChanged event)

MrNeRF: done! Thx!

PR #299, opened 2025-08-13
Fix memory leak on windows when it is not longer refining

MotivaCG:
In the previous method after stopping refining a memory leak was happening in windows.

It also avoids per-step cache flush; minor increase in memory between trims is expected while avoiding the overhead of clearing on every step.

### DIFF
--- a/src/default_strategy.cpp
+++ b/src/default_strategy.cpp
@@ -395,13 +395,17 @@ void DefaultStrategy::post_backward(int iter, gs::RenderOutput& render_output) {
         if (_params->stop_refine_scale2d > 0) {
             _radii.zero_();
         }
-
-        c10::cuda::CUDACachingAllocator::emptyCache();
     }
 
     if (iter % _params->reset_every == 0 && iter > 0) {
         reset_opacity();
     }
+
+#ifdef _WIN32
+    // Windows doesn't support CUDACachingAllocator expandable_segments
+    if (iter % 10 == 0 || iter == _params->stop_refine || iter == _params->iterations)
+        c10::cuda::CUDACachingAllocator::emptyCache();
+#endif
 }
 
 void DefaultStrategy::step(int iter) {
--- a/src/mcmc.cpp
+++ b/src/mcmc.cpp
@@ -375,15 +375,16 @@ void MCMC::post_backward(int iter, gs::RenderOutput& render_output) {
 
         // Add new Gaussians
         add_new_gs();
-
-#ifdef _WIN32
-        // Windows doesn't support CUDACachingAllocator expandable_segments
-        c10::cuda::CUDACachingAllocator::emptyCache();
-#endif
     }
 
     // Inject noise to positions
     inject_noise();
+
+#ifdef _WIN32
+    // Windows doesn't support CUDACachingAllocator expandable_segments
+    if (iter % 10 == 0 || iter == _params->stop_refine || iter == _params->iterations)
+        c10::cuda::CUDACachingAllocator::emptyCache();
+#endif
 }
 
 void MCMC::step(int iter) {


src/mcmc.cpp
```diff
@@ -375,15 +375,16 @@ void MCMC::post_backward(int iter, gs::RenderOutput& render_output) {
...
-        c10::cuda::CUDACachingAllocator::emptyCache();
-#endif
     }
 
     // Inject noise to positions
     inject_noise();
+
+#ifdef _WIN32
+    // Windows doesn't support CUDACachingAllocator expandable_segments
+    if (iter % 10 == 0 || iter == _params->stop_refine || iter == _params->iterations)
```
MrNeRF: Why would the last condition make sense: iter == _params->iterations make sense?
iter == _params->stop_refine  makes also no sense if you empty the cache every ten iterations anyway.

src/default_strategy.cpp
```diff
@@ -395,13 +395,17 @@ void DefaultStrategy::post_backward(int iter, gs::RenderOutput& render_output) {
...
-        c10::cuda::CUDACachingAllocator::emptyCache();
     }
 
     if (iter % _params->reset_every == 0 && iter > 0) {
         reset_opacity();
     }
+
+#ifdef _WIN32
+    // Windows doesn't support CUDACachingAllocator expandable_segments
+    if (iter % 10 == 0 || iter == _params->stop_refine || iter == _params->iterations)
```
MrNeRF: Why would the last condition make sense: iter == _params->iterations make sense?
iter == _params->stop_refine  makes also no sense if you empty the cache every ten iterations anyway.

MotivaCG: The idea is that it forces a cleanup every time its changes its "state" as well as with the latest step even if the current step % 10 != 0

MrNeRF: Yeah, but in fact every of these steps is an multiple of 10.  And note, this has nothing to do with a memory leak!

MotivaCG: No, the fix is not related to being a multiple of 10, but placing the cache clean out of the if(refining).

The multiple of 10 check is introduced to avoid cleaning the cache on every single iteration since it is not required.

Concerning iterations using default settings, the stop conditions will always be multiples of 1,0 but obviously any user could set the stop condition to 29997 iterations, it is not a high priority, but an existing possibility, since it is a simple or condition, it is not harmful at all.

MrNeRF: Sorry. Either you clean that up or I don't merge it. This pointless discussion keeps me just busy.

MotivaCG: Sorry probably I didn't explain it properly. Let me clarify the logic behind the cache cleanup condition.

1. Why iter % 10 == 0?
The cache is cleared every 10 iterations to balance performance and memory usage. Clearing the CUDA cache on every single step would introduce significant overhead, slowing down the training process. By clearing it less frequently, we allow for minor memory increases between trims while avoiding the performance penalty of a per-step flush.

2. Why || iter == _params->stop_refine || iter == _params->iterations?
These conditions are crucial for robustness. While the iter % 10 == 0 rule works for most steps, the stop_refine and iterations values are user-configurable and may not always be a multiple of 10.

The stop_refine condition ensures that the cache is cleaned at the exact moment the refinement phase ends, freeing up memory used by those specific operations.

The iterations condition guarantees a final cleanup at the very end of the training process, regardless of the final iteration number. So with that triple condition it is a robust solution that accounts for all possible user configurations.

MrNeRF: % 10 is fine! Everything else is not!

dozeri83: 1. first - i agree that memory leak is a bad name - maybe -"poor memory management"  

2.  @MotivaCG I dont know about  "|| iter == _params->stop_refine || iter == _params->iterations" part 
( empty cache is just a hint to libtorch to clean the cache, not empying the cache at the end is OK)

3. @MrNeRF  - I agree that the cache needs to be emptied in windows in this method , but I tested that cleaning it in every iter reduces the performance. so this less frequent cleaning makes sense.

MotivaCG:
Changes done

MrNeRF:
Thx! LGTM!

PR #300, opened 2025-08-13
Masking

MotivaCG:
Masking involves some different steps:
· Giving priority to some areas in the loss function.
· Assigning alpha penalty for inside/outside mask areas
· Cleaning up gaussians after the full training.

### DIFF
--- a/src/training/rasterization/fast_rasterizer.cpp
+++ b/src/training/rasterization/fast_rasterizer.cpp
@@ -60,7 +60,22 @@ namespace gs::training {
         output.image = raster_outputs[0];
         output.alpha = raster_outputs[1];
 
-        output.image = output.image + (1.0f - output.alpha) * bg_color.unsqueeze(-1).unsqueeze(-1);
+        // Apply background - handle both single color [3] and full image [3, H, W]
+        if (bg_color.numel() == 3) {
+            // Single color case - expand to image dimensions
+            output.image = output.image + (1.0f - output.alpha) * bg_color.unsqueeze(-1).unsqueeze(-1);
+        } else if (bg_color.dim() == 3 && bg_color.size(0) == 3) {
+            // Full image case [3, H, W] - already correct size
+            TORCH_CHECK(bg_color.size(1) == height && bg_color.size(2) == width,
+                        "bg_color full image size mismatch in fast_rasterizer. Expected [3, ", height, ", ", width,
+                        "], got ", bg_color.sizes());
+
+            // Apply background: rendered + (1 - alpha) * background
+            output.image = output.image + (1.0f - output.alpha) * bg_color;
+        } else {
+            TORCH_CHECK(false, "bg_color must be either [3] (single color) or [3, H, W] (full image), got ",
+                        bg_color.sizes());
+        }
 
         return output;
     }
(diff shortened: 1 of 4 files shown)


GibsDev (commented):
Thank you for your work on this! I've been having good success with this masking implementation.

I have a couple of suggestions that might help it align better with the open issue for masking as well as some more selfish suggestions that make it more convenient for my particular use case that I hope will be useful to others:

- https://github.com/MrNeRF/gaussian-splatting-cuda/issues/114

I think it would be nice if the mask filenames also supported the same format that colmap uses by default (https://colmap.github.io/faq.html#mask-image-regions). Colmap uses a `.png` extension on the end of the original filename instead of the exact same filename extension as the image. For example: `<image/some/name/img_0.jpg>.png`

I also think it would be nice to add an option to disable the masked based splat pruning in the configuration. I think the pruning makes sense for training objects that are masked out of a scene, but the pruning has a negative effect when trying to train an environment while masking out specific objects. I've found that applying the masks to the loss function alone is sufficient at removing masked objects from a scene (at least with the iteration counts that I'm using). Otherwise objects in the distance obfuscated by a masked region tend to turn into a blob in the final result.

I hope this makes it into the main branch at some point!

MrNeRF:
I will check the pr carefully next week. I think at this point it is to invasive or rather it is very specific for a certain use case. At this point the build seems also broken. Regardless, mask support is certainly important.

lyehe:
I checked the masking code today. I think the way of implementing mask is hardcoded? (luckily our use case)

I kinda see mask more generally useful in removing objects and removing background. The implementations are different.
I looked at the PostShot's mask implementation and I think that can be specified.


> 
>https://www.jawset.com/docs/d/Postshot+User+Guide/Interface/Training+Configuration#Image_Masks
> Mask Mode
> The Mask Mode specifies how the masks are applied to the input images.
> 
> Remove Occluders: the active (white) regions of the mask images will be ignored from the associated input image only. That is, the trained model may still reconstruct pixels in the masked space, but learn it only from other images.
> 
> This mode allows removal of unwanted objects in the images.
> 
> Remove Background: the inactive (black) regions of the mask images are considered background and must not be reconstructed. That is, the model will be trained to reconstruct empty space (fully transparent pixels) for these regions.
> 
> This mode can be used to isolate objects from their background.
> 
> Note: Using image masks to remove backgrounds may result in artefacts and holes in some cases. Training without masks and then cropping the model in 3D will produce higher quality results.
> However, image masking is necessary to remove the background when training turntable shots.



I see two design decisions:
1. Follow the two mode method. Basically a mode to remove everything outside mask (the current one) and a mode to ignore occasional Occluders like the pedestrians. 

2. Specify what to do in/outside mask.  Provide options like prune/low-alpha outside, ignore outside, force high alpha inside, etc. This is more expansible in the long term but can be confusing.

MrNeRF:
Yeah, I opt for 1 for the moment as it appears to be the broader use case

MotivaCG:
Yes, it is currently designed to pay much more attention during training in the masked areas, and it also introduces a penalty on the alpha of the non-masked areas.

MrNeRF:
I will close this as I won't merge it anyway.

PR #301, opened 2025-08-13
Frame controller bug fix

dozeri83:
another  bug fix in frame control where crop box was involved. 
@MrNeRF  - sorry - I hope it will be the last


MrNeRF:
thank you!

PR #304, opened 2025-08-14
Allow switching between gsplat and the fast rasterizer 

calvin-laurenson:


MrNeRF:
@calvin-laurenson 
What is the purpose of this pr? The idea is to remove gsplat soon. Default strategy will crash with gsplat anyway.

I want to just keep the parts that are required for 3DGut. They should be ideally rewritten in a similar way like the new rasterizer. As long as the tool is only used by "experts", it would be great to identify and fix bugs in the new rasterizer (if any). If there are issues, someone might switch to this option without reporting the problem. 

This needs to be fixed https://github.com/MrNeRF/gaussian-splatting-cuda/pull/187 first however.

PR #317, opened 2025-08-20
github actions to run CI for Ubuntu

miriameng:
need this in order for dependabot updates to be meaningful, but this should also be generally very helpful

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -1,0 +1,44 @@
+name: Ubuntu CUDA Build
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build-ubuntu:
+    runs-on: ubuntu-24.04
+    container:
+      image: nvidia/cuda:12.8.0-devel-ubuntu24.04
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


miriameng:
@MrNeRF is there a “Run workflow” button in the Actions tab that you can click? I probably don't see it due to not having write access. github actions will only be automatically triggered after this yml file is in the main branch

MrNeRF:
IMG_0281
That’s all I can see.

miriameng:
What do you see when you click the "Actions" tab?

MrNeRF:
https://github.com/MrNeRF/gaussian-splatting-cuda/pull/316#issuecomment-3207572717

miriameng:
I am not asking about the failure of dependabot, I am asking about whether you see the option to manually trigger CI, which is what I added in this PR, to see if I made any mistakes. Or can we simply merge first so CI will get triggered automatically from now on.

MrNeRF:
Nope, there is none!

miriameng:
Can you consider merging first, and if there's a failure I will follow up with another PR?

MrNeRF:
Here we go. Hope it helps.

JamieMagee:
Your `vcpkg.json` has a UTF8 Byte Order Mark (BOM) at the beginning. Remove that and Dependabot will work.

image

MrNeRF:
CC @miriameng

PR #320, opened 2025-08-20
fix github actions CI for ubuntu

miriameng:
follow up on #317


MrNeRF:
I added you as collaborator @miriameng . It is night over here so I hope it helps.

miriameng:
memory are quite limited in github actions, so had to build sequentially, can probably try to explore how many threads we can build in parallel before it OOM if we think it takes too long to build

miriameng:
@MrNeRF build passes!

MrNeRF:
Awesome! I will setup a local worker soon so that should build faster.
@miriameng this would not build simply with our docker image?

miriameng:
It looks to me that we only have the DockerFile to create the docker image ourselves, but we don't have docker image published to docker hub yet for GitHub actions to use. Is that right, @shadygm?

shadygm:
At the moment we don't push to docker hub. I think you can do it with something along [these lines](https://github.com/shadygm/groningen-hunter/blob/main/.github/workflows/docker-publish.yml). The example pushes to the github registry, but you can change it to docker hub (github might even provide a template for it, I'm not 100% sure). 

@miriameng if some help is needed, let me know!

PR #330, opened 2025-08-21
Add CUDA fat-binary build: BUILD_CUDA_ALL_SM + configurable BUILD_CUDA_ALL_MIN_SM (default 86)

MotivaCG:
This PR adds two new CMake options to enable shipping CUDA fat binaries:

BUILD_CUDA_ALL_SM (OFF by default): when ON, builds native SASS for all detected SMs ≥ a configurable minimum and appends a -virtual PTX for the highest SM for forward compatibility.

BUILD_CUDA_ALL_MIN_SM (STRING cache, default 86): minimum compute capability considered when BUILD_CUDA_ALL_SM=ON. Input is validated to be an integer ≥ 70.

Advantages when enabled:

· Avoids first-run JIT latency on supported GPUs.

· More consistent performance (native SASS per SM).

· Fewer PTX/driver toolchain mismatches.

· One binary works across heterogeneous GPU families.

Disadvantages when enabled:

· Larger binaries.

· Longer compile/link times; device linking (nvlink) also takes longer.

· Debugging/profiling can be trickier with multi-arch device code (need to select the right SM in tools).

No behavior change when disabled:
With BUILD_CUDA_ALL_SM=OFF (default), the build remains exactly as before, only the native architecture is targeted.


MotivaCG:
Testing:

`"C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v12.8\bin\cuobjdump.exe" -lelf "D:\Dependencias\gaussian-splatting-cuda\BUILD\gsplat\Release\gsplat_backend.lib"`
```

member D:\Dependencias\gaussian-splatting-cuda\BUILD\gsplat\Release\gsplat_backend.lib:gsplat_backend.dir\Release\gsplat_backend.device-link.obj:
ELF file    1: gsplat_backend.1.sm_89.cubin
ELF file    2: gsplat_backend.2.sm_86.cubin
ELF file    3: gsplat_backend.3.sm_87.cubin
ELF file    4: gsplat_backend.4.sm_90.cubin
ELF file    5: gsplat_backend.5.sm_100.cubin
ELF file    6: gsplat_backend.6.sm_101.cubin
ELF file    7: gsplat_backend.7.sm_120.cubin

... and more
```

MrNeRF (changes requested):
See the wrong naming. That needs to be corrected. Otherwise should work.

CMakeLists.txt
```diff
@@ -60,6 +68,99 @@ else()
...
+    if(BUILD_CUDA_ALL_MIN_SM LESS 70)
+        message(FATAL_ERROR "BUILD_CUDA_ALL_MIN_SM must be >= 70 when BUILD_CUDA_ALL_SM=ON.")
+    endif()
+
+    # Fat binary for all the families
+    gs_detect_supported_sms(CUDA_SMS MIN_SM ${BUILD_CUDA_ALL_MIN_SM})
+    if(CUDA_SMS)
+        # Append PTX of the highest SM for forward compatibility with future GPUs
+        list(GET CUDA_SMS -1 _max_sm)
+        set(CMAKE_CUDA_ARCHITECTURES "${CUDA_SMS};${_max_sm}-virtual" CACHE STRING "CUDA architectures" FORCE)
```
miriameng: this will potentially not work with libtorch: https://github.com/pytorch/pytorch/blob/e25ee0290ef16503f178e04890c15717f6e9ea44/cmake/public/cuda.cmake#L316-L321, I got confused by this before, you have to use `TORCH_CUDA_ARCH_LIST` instead for libtorch

MotivaCG: You were right! modified!

MotivaCG:
> See the wrong naming. That needs to be corrected. Otherwise should work.

Done too

CMakeLists.txt
```diff
@@ -106,6 +225,12 @@ find_package(spdlog REQUIRED)
 find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
 find_package(Freetype REQUIRED)
 
+# TORCH_CUDA_ARCH_LIST is already defined by Torch
+# use it here as the global default CUDA architecture list too
+if(gaussian_splatting_cuda_CUDA_ARCH)
+    set(CMAKE_CUDA_ARCHITECTURES "${gaussian_splatting_cuda_CUDA_ARCH}" CACHE STRING "CUDA architectures" FORCE)
+endif()
```
miriameng, line 232: If this is called before libtorch, it will get unset by libtorch, if this is called after libtorch, are we not afraid of something might break because libtorch wants this to be unset?

PR #331, opened 2025-08-21
Make PLY export naming user-friendly

MotivaCG:
This PR updates Trainer::save_ply so exported files are named with a 1-based iteration index. We now pass iter_num + 1 to the model’s save_ply and to the filename (splat_<iter_num+1>.ply). As a result, when a run is configured for 15000 iterations (looping 0..14999), the final export is splat_15000.ply, matching user expectations.

 Avoids confusion caused by 0-based internal counters producing splat_14999.ply for a 15000-step run.

Internal compatibility: The project timeline entry (lf_project_->addPly(...)) still stores the original iter_num to preserve zero-based indexing semantics and backward compatibility with consumers that rely on it.

Example:

Before: last file -> splat_14999.ply for a 15000-iteration run.

After: last file -> splat_15000.ply.


MrNeRF:
Lgtm

PR #335, opened 2025-08-22
fix invalid ccache path for github actions

miriameng:
lots of gotchas in github actions


miriameng:
Finally setup ccache successfully

miriameng:
But it doesn't re-cache the ccache dir, need to figure out how to do it

miriameng:
should re-cache the ccache dir now

PR #336, opened 2025-08-22
check clang format in github actions

### LINKED ISSUE #332: CI: Test formatting of the Changed Files
There is a .clang-format. We should enforce this in the CI. 
Can we do this @miriameng?

miriameng:
fixes #332

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -10,6 +10,10 @@ on:
 jobs:
   clang-format:
     runs-on: ubuntu-24.04
+    permissions:
+      contents: write
+      pull-requests: write
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -28,12 +32,14 @@ jobs:
           find . -regex '.*\.\(cpp\|hpp\|cc\|c\|h\|cu\|cuh\)' \
             -exec clang-format -i {} +
 
-      - name: Check for formatting issues
+      - name: Commit changes (if any)
         run: |
           if [[ -n "$(git status --porcelain)" ]]; then
-            echo "Some files are not formatted correctly. Please run clang-format."
-            git --no-pager diff
-            exit 1
+            git config user.name "github-actions[bot]"
+            git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+            git add .
+            git commit -m "Apply clang-format"
+            git push
           else
-            echo "All files are properly formatted."
+            echo "No formatting changes needed."
           fi
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -16,8 +16,8 @@ jobs:
 
       - name: Install clang-format
         run: |
-          apt-get update
-          apt-get install -y clang-format
+          sudo apt-get update
+          sudo apt-get install -y clang-format
 
       - name: Run clang-format
         run: |
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -41,7 +41,14 @@ jobs:
             git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
             git add .
             git commit -m "Apply clang-format"
-            git push
+
+            if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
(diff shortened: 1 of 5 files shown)


miriameng:
@MrNeRF there are tons of existing clang-format issues, is this expected?

MrNeRF:
I haven't applied to the fastgs/gsplat folder. Will do so. Seems to work :)

MrNeRF:
@miriameng Can you please sync with master?

miriameng:
Yes but it seems like there are still a few more formatting issues

MrNeRF:
ah damn, my command didn't include the cu/cuh files. Anyway, I am getting now a build error. Need to fix this first.

MrNeRF:
Used now the same command:
`find . -path ./external -prune -o -regex '.*\.\(cpp\|cc\|c\|h\|cu\|cuh\|hpp\)' -exec clang-format -i {} +   `
After you sync, it should hopefully work!

MrNeRF:
Now that's weird. It should pass the clang-format test.

miriameng:
Could it be due to a difference in the version of the clang format between yours and the one on GitHub actions?

MrNeRF:
Isn't the .clang-format the important part?

```
> clang-format --version
Ubuntu clang-format version 14.0.0-1ubuntu1.1
```

MrNeRF:
can you try to format, push, and see what happens?

miriameng:
Yours seems quite old?

CI shows Ubuntu clang-format version 18.1.3 (1ubuntu1)

MrNeRF:
I will try to update!

MrNeRF:
I switched to 15 and it changes things actually, lol. But still not correct. I have to try later another higher version.

miriameng:
Just curious, how were you installing your clang-format? Is it not apt-get install clang-format?

MrNeRF:
It is. Just used apt-get

MrNeRF:
Yes it was regular apt-get. However it is on ubuntu 22.4

miriameng:
Do you want me to update CI to use 22.04 instead?

MrNeRF:
No because the build requires >=gcc++14 which is available as package starting from ubuntu 24.4. This issue might also impact others, right?

miriameng:
I guess it's time for you to update your OS then!

MrNeRF:
It might solve the problem for me but should I ask every contributor to update their os?

Let me experiment with it later. There must be a proper solution.

miriameng:
One idea is let github actions auto push the clang format changes based on its own version. Benefit is developers don't have to deal with clang format themselves anymore locally, what do you think?

MrNeRF:
Sounds pretty convenient. How hard would that be to setup? Likely easy?

MrNeRF:
Deleted the cache just to make sure that nothing weird is cached. I believe that's truly best when that's automatically assured. Otherwise formatting can get very annoying if there will be random struggle.

miriameng:
Sorry not understanding. What cache are you referring to?

MrNeRF:
> Sorry not understanding. What cache are you referring to?
The caching that is done by github actions. I cleared that to see if there is something weird going on. Don't mind, didn't help at all.

I still find it wild that there are differences between version. What's there reasoning behind it? In consequence requiring everybody to run the identical version of clang-format.

miriameng:
doesn't seem possible to commit and push to a fork unfortunately

miriameng:
another idea is auto apply clang format when a PR is merged or pushed to master, let me try that

miriameng:
I guess the problem is we will only know if it works after this PR is merged.

MrNeRF:
> I guess the problem is we will only know if it works after this PR is merged.

yolo! Merging!

miriameng:
```
remote: error: GH006: Protected branch update failed for refs/heads/master.        
remote: 
remote: - Changes must be made through a pull request.        
To https://github.com/MrNeRF/gaussian-splatting-cuda.git
 ! [remote rejected] HEAD -> master (protected branch hook declined)
error: failed to push some refs to 'https://github.com/MrNeRF/gaussian-splatting-cuda.git
```
I see you pushing to master directly all the time, did you turn on the "Protected branch" but made yourself an exception?

PR #337, opened 2025-08-22
restore-keys with rolling cache for ccache in github actions

miriameng:
doesn't seem possible to overwrite the ccache when it's in use, try using rolling cache this time and see if it works


miriameng:
Looks like it's working based on the build time of my manual rebuild!

MrNeRF:
That was fast. Yes!

PR #345, opened 2025-08-24
Add timelapse image saving

calvin-laurenson:
Saves a render of specified images every N steps in training. I had to do some unpleasant stuff so the cameras were initialized correctly at the start of training but the fix seems to work. I also have an ffmpeg command that converts the images into a video but I am not sure where the best place to display the command is.

### DIFF
--- a/include/core/camera.hpp
+++ b/include/core/camera.hpp
@@ -37,6 +37,9 @@ namespace gs {
         // Load image from disk and return it
         torch::Tensor load_and_get_image(int resize_factor = -1);
 
+        // Load image from disk just to populate _image_width/_image_height
+        void load_image_size(int resize_factor = -1);
+
         // Get number of bytes in the image file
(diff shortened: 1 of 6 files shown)


calvin-laurenson:
@miriameng seems like there is a problem with the windows CUDA toolkit cache

miriameng:
Let me try to figure out what's causing the problem

MrNeRF:
For the gui, how about a checkbox and a numeric field for the interval? Can be done in a separate pr

MrNeRF:
> render of specified images every N steps in training. I had to do some unpleasant stuff so the cameras were initialized correctly at the start of training but the fix seems to work. 

Yeah, this can be made more elegant but not in this pr.

> I also have an ffmpeg command that converts the images into a video but I am not sure where the best place to display the command is.

Let's do this in another pr!

PR #346, opened 2025-08-24
fix windows github actions when cache is hit

miriameng:


miriameng:
removing caching of the cuda install, whatever workaround proposed by LLMs does not work, taking a closer look at the cmake source code, apparently cuda installation in windows is not equivalent to a simple unzip, there are registry changes and visual studio changes made by the cuda installer

PR #358, opened 2025-08-28
World transform +/-  

dozeri83:
brought back +/- arrow for world transform. 

the current mechanism was drag input and it was not good since the the input bar were too short


MrNeRF:
tested. Looks good to me!

PR #360, opened 2025-08-28
Pad distortion parameters for 3DGUT

### LINKED ISSUE #355: GUT - deal with different cases of distortion
training with GUT with "Fern" dataset:
got - 

(guessing there is only k1, but we should deal with other scenarios)

"Training error: Training step failed: radial_coeffs last dimension must be 4 or 6, got 1"

calvin-laurenson:
Closes #355. Untested because I had to update discord and it imploded my COLMAP build because arch linux. Someone should probably test with a RADIAL or SIMPLE_RADIAL dataset.


MrNeRF:
tested with SIMPLE_RADIAL and OPENCV_FISHEYE. Works!

PR #365, opened 2025-08-29
License add CUDA Linking Exception (GPLv3 Section 7)

MotivaCG:
Adds a GPLv3 section 7 Additional Permission to the end of LICENSE, allowing this project to link with NVIDIA’s proprietary CUDA-family libraries (CUDA/cuDNN/TensorRT/NCCL/NVRTC, etc.) and distribute the combined work under GPLv3.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -672,3 +672,38 @@ may consider it more useful to permit linking proprietary applications with
 the library.  If this is what you want to do, use the GNU Lesser General
 Public License instead of this License.  But first, please read
 <http://www.gnu.org/philosophy/why-not-lgpl.html>.
+
+
+18. Additional permission: CUDA Linking Exception (GPLv3 section 7)
+
+CUDA Linking Exception (GPLv3 section 7 - Additional Permission)
+
+As an additional permission under Section 7 of the GNU General Public License
+version 3, you are allowed to link this work, and works based on it, statically
+or dynamically, with the unmodified proprietary NVIDIA GPU libraries and runtime
+components required to build and run CUDA-based software, and to copy and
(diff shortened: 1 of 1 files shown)


MrNeRF:
Where do you have this section from?

MotivaCG:
> Where do you have this section from?

I knew about this section 7
_"Additional permissions" are terms that supplement the terms of this License by making exceptions from one or more of its conditions. Additional permissions that are applicable to the entire Program shall be treated as though they were included in this License, to the extent that they are valid under applicable law. If additional permissions apply only to part of the Program, that part may be used separately under those permissions, but the entire Program remains governed by this License without regard to the additional permissions.

When you convey a copy of a covered work, you may at your option remove any additional permissions from that copy, or from any part of it. (Additional permissions may be written to require their own removal in certain cases when you modify the work.) You may place additional permissions on material, added by you to a covered work, for which you have or can give appropriate copyright permission._

More info: https://libreguide.org/gplv3_section_7.html

It was designed for this kind of cases so I asked a LLM to create a section on Lichfeld license that uses that GPL section to grant the usage of the CUDA related libraries, based on that FSF section.

MrNeRF:
I see. I will have to check carefully. I don't think a llm is the best source to extend a license. I will need some time to verify if that's required at all.

MotivaCG:
> I see. I will have to check carefully. I don't think a llm is the best source to extend a license. I will need some time to verify if that's required at all.

Technically it is required since GPL projects should not be linked to closed libraries as CUDA ones. Others like Blender make a more complicated approach with multi modules keeping Nvidia privative libraries out of the core.

PR #367, opened 2025-08-29
vcpkg hardcode approach alternative

wh1t3lord:
Hello it is me ~mario~ maybe you didn't forget about our convo from like yesterday (?), so yeah I would like to fix the https://github.com/MrNeRF/gaussian-splatting-cuda/issues/357

But for now I don't like the approach for hardcoding vcpkg using env variables, so it is really better to clone vcpkg to the root folder of repo project.

So it is really nice to specify cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE="./vcpkg/scripts/buildsystems/vcpkg.cmake" because we could possibly have a case when user nuked its vcpkg and a scenario IT was used for different projects (not just for gaussian splatting cuda) and since it was ruined he can't just easily resolve the dependencies (or remember them well) that were used and handled by single vcpkg instance that was specified and used by these projects just because it was used the route through env variables it might be a not good situation for you to receive comments about bad user experience just because they got not properly working vcpkg and now they can't build their projects...

Yeah, isolating vcpkg and providing ability to manually specifying own instance of vcpkg is a quite nice feature :D
I hope you find this request as reasonable.

So if you don't mind I just made in CMake a default handling if/endif for a case when user went to default route like using env variables (backward compatibility) BUT if it was specified like through  -DCMAKE_TOOLCHAIN_FILE then all is fine and we do as is like env variable route.

Maybe my update of readme.md is not beauty or how you would expect to see so I hope you can fix it as you wish or like.

Cheers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ data
 test.py
 *.nsys-rep
 testdata
+vcpkg/
+vcpkg/**
 
 *.pt
 !weights/*.pt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,11 @@
 # windows/linux platform independent
 set(RAW_VCPKG_PATH $ENV{VCPKG_ROOT})
 file(TO_CMAKE_PATH "${RAW_VCPKG_PATH}" VCPKG_ROOT)
-set(CMAKE_TOOLCHAIN_FILE "${VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake")
+
+if ("${CMAKE_TOOLCHAIN_FILE}" STREQUAL "")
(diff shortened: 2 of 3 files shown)


MrNeRF:
Haha, I do remember 🙂
The change makes sense to me and is useful. I don't know if it contributes to make this repo less scary to novel users though. They want binaries anyway.

We will need proper docs on the homepage.

Lgtm. Approved. Will merge after CI is done.

MrNeRF (approved):

MrNeRF:
Btw, not sure if you are on the discord but there was some discussion how to implement this feature best in case you need some ideas.

PR #368, opened 2025-08-30
prioritize colmap .bin over .txt if both exists

### LINKED ISSUE #315: Will not start training if both .bin and .txt files are found
If the COLMAP files exist in both formats in the sparse/0 folder then the program does not proceed; no error or other message shown.

miriameng:
attempt to fix #315


MrNeRF:
Lgtm!

PR #369, opened 2025-08-30
initialize training in setup

dozeri83:
currently you don't see the images list in the scene panel when loading a scene, because the trainer is only initialized
in the start training. I think it is ok to call initialization on "set trainer".
it is either that or separate the loader from the initialization.

### DIFF
--- a/src/visualizer/training/training_manager.cpp
+++ b/src/visualizer/training/training_manager.cpp
@@ -35,6 +35,17 @@ namespace gs {
 
             setState(State::Ready);
 
+            // Initialize trainer if not already initialized
+            if (!trainer_->isInitialized()) {
+                LOG_INFO("Initializing trainer");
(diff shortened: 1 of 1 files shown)


dozeri83:
now that I think of it - this causes a problem with the "changing of training params" feature. 
so we have to think of different fix

MrNeRF:
I think the data loading must be disentangled because it has the same path as in headless mode.

The training manager should manage the training data and pass it to the trainer after the start button was hit. Something like this.

Now that some parameters are changeable, I want also be able to select the strategy from the gui.

dozeri83:
> I think the data loading must be disentangled because it has the same path as in headless mode.
> 
> The training manager should manage the training data and pass it to the trainer after the start button was hit. Something like this.
> 
> Now that some parameters are changeable, I want also be able to select the strategy from the gui.

ok. closing this...

PR #371, opened 2025-08-31
fixed images not loading in scene panel.

dozeri83:
images not showing in scene panel + 
fixed project not updated in trainer - this caused plys not to load in the end of the training and saved in project file.

@MrNeRF  - fix was easy since the dataset in loaded in the CTOR, in init there was only the split.


MrNeRF:
Yup, works! Lgtm. Thank you!

PR #382, opened 2025-09-03
resize factor in gui fix

dozeri83:
1. resize factor in gui - present user with options.
2. seep resize factor into trainer correctly.


MrNeRF:
That's nice! Thank you!

PR #386, opened 2025-09-03
Fix for correct linking python libraries based on different build configurations in Visual Studio for Windows (and other platforms); Ignoring additional build folder namings

wh1t3lord:
@MrNeRF hi you asked for fixing python3xx_d.lib and as you requested I responded :) 

I tested just generated project as usual we do and there indeed a correct python3xx_d.lib as expected (before my fixes it was linking release version of python instead of debug version of library like *_d.lib)


MrNeRF:
ok, that's great. Thanks!

PR #397, opened 2025-09-05
Command prompt update

dozeri83:
change developer command prompt to x64 command prompt,
and added ninja generator in windows README

PR #398, opened 2025-09-05
update cmake/scripts/readme for the repo name change

### LINKED ISSUE #396: change the executable name to LichtFeld-Studio?
Now that the repo is renamed to `LichtFeld-Studio`, should we also rename the executable from `gaussian_splatting_cuda` to `LichtFeld-Studio` for consistency?

miriameng:
fixes #396

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ if ("${CMAKE_TOOLCHAIN_FILE}" STREQUAL "")
 endif()
 
 cmake_minimum_required(VERSION 3.24)
-project(gaussian_splatting_cuda LANGUAGES CUDA CXX C)
+project(LichtFeld-Studio LANGUAGES CUDA CXX C)
 
 # Setup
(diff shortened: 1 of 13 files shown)


MrNeRF:
Awesome! Looking forward! The CI seems to work as well greatly!
Haven't checked but do you think this https://github.com/MrNeRF/LichtFeld-Studio/pull/397 can have an impact on the speed of the Windows build?

miriameng:
possibly yes, ninja is known to be fast and msbuild is known to be slow, but I don't know whether there's a working setup for using the debugger perhaps with vscode + ninja + windows. I only know on Ubuntu vscode + ninja + gdb works sufficiently well

MrNeRF:
Sorry about the confusion. So it has actually a large speedup on Windows which was reported by multiple users. 

My question is rather with respect to the CI pipeline Windows build. Can something be transfered to the CI from this guide to improve the runtime of the windows build.

PR #399, opened 2025-09-06
use ninja for CI

miriameng:


miriameng:
doesn't look like there's any improvement at all in the build time by switching to ninja ..

MrNeRF:
Unfortunately. I will merge anyway.

PR #400, opened 2025-09-06
Documentation update for the new name + some improvements for the windows build 

kmeirlaen:
Documentation update for the new name + some improvements for the windows build that were discussed in the #dev discord channel: 

do not use SET to set the environment variable for vcpkm, but just pass the entire path in the command. as users will be using copy/paste, this now takes less steps again and is more clear on what to copy/paste. also avoids errors when recreating the build files and forgetting to use "SET"


MrNeRF (approved):
Awesome! Thank you!

PR #401, opened 2025-09-06
Remove Python dependencies

otri:
I noticed we only had one real place where Python3 headers were needed to do a build, that's due to "torch/extensions.h".  That was just replaceable with "torch/torch.h".

- Drop find_package(Python3) and all Python include/link usage in CMake
- Replace torch/extension.h with torch/torch.h to avoid Python.h
- Remove Python dev header note from README

So, no more python dependency needed at all anymore.
Happy Building!


MrNeRF:
Awesome! Thank you!

dozeri83:
OMG - that PR is desperately  needed, thank you!

PR #404, opened 2025-09-07
ci: cache vcpkg binary archives

miriameng:
Hopefully to cut the windows CI runtime by 20 mins

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -61,6 +61,14 @@ jobs:
           restore-keys: |
             ccache-${{ runner.os }}-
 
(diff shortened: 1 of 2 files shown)


MrNeRF:
My impression is that the Linux build becomes faster and windows gets slower?

miriameng:
Windows build can't be slower. At most the build time will stay the same if the cache isn't working. Let me manually trigger the build a few times if you want to see the cache working before merging

miriameng:
@MrNeRF for the manual rebuild, the cmake time drops from the usual 20+ minutes to 1+ minutes, please take a look

MrNeRF:
Awesome! I appreciate?

PR #405, opened 2025-09-07
Add Docusaurus

calvin-laurenson:
I'm guessing this could be a contentious topic so don't feel compelled to merge soon. Still needs a favicon and logo


simonbethke:
I think this has to get properly integrated in this repository:
https://github.com/MrNeRF/lichtfeld-studio-web
And it needs to work together with Astro, not against it. Also it is important that in the end there is only a single command to build Astro and Docusaurus

docs/README.md
```diff
@@ -0,0 +1,20 @@
...
+```bash
+pnpm start
+```
+
+This command starts a local development server and opens up a browser window. Most changes are reflected live without having to restart the server.
+
+## Build
+
+```bash
+pnpm build
```
MrNeRF, line 27: Build intstructions are  not clear. Must be run in this folder!
npm install -g pnpm
pnpm install
pnpm build

PR #409, opened 2025-09-07
cmake: bump 3.30 and use CXX_COMPILER_FRONTEND_VARIANT

kzubatov:
Suddenly I have to use clang-cl now, but it treats -Wall as -Weverything so clang-cl users have a lot of "not compatible with c++98" warnings

CXX_COMPILER_FRONTEND_VARIANT was added in version 3.30, which allows to separate clang-cl (MSVC frontend)
and clang (GNU frontend) to use different flags.
The other compilers (MSVC, GNU) can only have one frontend, so their flags remain unchanged.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -28,6 +28,22 @@ jobs:
       - name: Install GCC 14 and dependencies
         run: |
           apt-get update
+
+          # install cmake from https://apt.kitware.com to get cmake 3.30+
+          
+          apt-get install -y ca-certificates gpg wget 
+
+          test -f /usr/share/doc/kitware-archive-keyring/copyright ||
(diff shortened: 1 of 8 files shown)


miriameng:
CI failed because apt-get cmake is cmake 3.28, but I guess if we are willing to force people to use gcc 14 we should be willing to force people to use newer cmake, but ultimately it depends on @MrNeRF

miriameng:
Maybe also update the README about the new cmake requirements

MrNeRF:
@miriameng I am good with the update!
Let's please update the CI and the Readme in the commit.
The windows build instructions also require an update: https://github.com/MrNeRF/LichtFeld-Studio/blob/master/docs/windows_build_instructions.md

kzubatov:
Well, it's alive https://github.com/kzubatov/LichtFeld-Studio/actions/runs/17534548380/job/49795914358

miriameng (approved):

MrNeRF:
Lgtm! Ty!

PR #416, opened 2025-09-08
Enable / Disable image through the GUI

kmeirlaen:
* added event for enable/disable image
* new: right click on image shows "enable/disable image" (depending on the current status), when clicked, it will fire the event
* new: when scene is selected, pressing "SPACE" will enable/disable the last "selected for view" image using the event
* new: when image preview is open, pressing "SPACE" will enable/disable the current preview image using the event
* when loading image set, all images are enabled
* when pressing "refresh", image are reloaded and status is reset to enabled
* show strikethrough when image is not enabled
* add counter for enabled images and display it in the image list (images enabled/total) 

Todo:
* use only selected images for training


MrNeRF:
This is a bit on halt because it requires changes in the backend. I will finish and merge this pr after the libtorch free version has landed in master.

MrNeRF:
I am closing this. We have to redo this pr on dev or rather after dev is in master.

PR #423, opened 2025-09-10
move camera up/down with Q and E keys

kmeirlaen:
* Added movement using the Q and E key
* removed processWASDMovement() as it was not being used by the code and is probably something historic - all movement is handled by the update() function


MrNeRF:
Removing code is always awesome. Less code is more code :)

MrNeRF:
tested. LGTM. Thank you!

PR #431, opened 2025-09-11
Use executable relative paths for asset loading

MotivaCG:
Assets ( shaders and TTF fonts) are now searched as a fallback from a 'resources' subdirectory relative to the executable.

This enables portable deployment where the executable and resources folder can be distributed together as a self-contained package, independent of installation path or working directory.


MrNeRF:
lgtm

PR #432, opened 2025-09-12
gui and project bug fixes

dozeri83:
1. fixed a bug when app crash when loading project because of mismatch in field name.
2. fixed a bug where rotation around center was around the wrong axis.  

@MrNeRF  - we needed to rotate always around local axes not global, if we rotate global we get unintuitive rotation in the wrong direction


MrNeRF:
ad 2. The rotation should work in the global coordinate system as, i.e. opengl.
This change breaks the rotation totally. The issue is not the coordinate system, I think it is rather important to finish the reorientation of the scene so that everything acts in one common coordinate system.

ad 1. That's great. Thx!

dozeri83:
> ad 2. The rotation should work in the global coordinate system as, i.e. opengl. This change breaks the rotation totally. The issue is not the coordinate system, I think it is rather important to finish the reorientation of the scene so that everything acts in one common coordinate system.
> 
> ad 1. That's great. Thx!

I disagree and I have several arguments.. but here two:

1. if you are correct why there is no symmetry between pitch rotation (which is local) and yaw rotation? 
2. just play with the "rotation around center" and you see that there is a bug. I suggest taking blender dataset, and moving the viewport camera 45 deg down in the pitch direction, (since the **initial** viewport coord system is the same as the global so you can not tell the difference initially).

MrNeRF:
This change does not make the user controls any better. It is true that the controls are not very good atm.

Potentially, there can be a distinction between global and local coordinate that can be toggled or switched. The base should be control in the global system and the user should be able to align the scene with world coordinates.

PR #433, opened 2025-09-12
fixed loading json parameter

kmeirlaen:
the json parameter for num_workers was loaded using the wrong json index, causing a crash when loading a .ls file


MrNeRF:
Please see here: https://github.com/MrNeRF/LichtFeld-Studio/pull/432

MrNeRF:
Ok, I merge this one because the other one does a little bit more. 
Thx!

PR #436, opened 2025-09-13
Application icon

kmeirlaen:
This change adds an icon to the .exe under windows.

image

image

~~Note: application is started as a console app, so the icon is not visible in the taskbar of windows yet.~~
Now also sets the application icon in the window and taskbar (windows only)

image

image


MrNeRF:
Cool. Not sure if that works on linux. Will try later.

Can we put that in a resources folder?

MrNeRF:
Awesome! Thank you!

PR #440, opened 2025-09-14
Expose --gut (GUT) and --rc (Reality Capture) in GUI

kmeirlaen:
This change adds 2 checkboxes to to GUI:
* under dataset, we add "import from Reality Capture"
* under optimization, we add "GUT"

These checkboxes can be used to enable training of distorted images through the GUI without the need to specify their startup parameters.

image

We also change the error message that is displayed when training with distorted images is detected.
Finally, we allow a "reset" of the dataset in case a training error occurs. This allows the user to restart training with adjusted parameters without having to close the application.

image

### DIFF
--- a/src/training/trainer.cpp
+++ b/src/training/trainer.cpp
@@ -586,7 +586,7 @@ namespace gs::training {
                     return std::unexpected("Training on cameras with ortho model is not supported yet.");
                 }
             } else {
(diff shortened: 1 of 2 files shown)


MrNeRF:
lgtm. This is uncritical. If the Linux builds succeeds here, the Windows build will as well!

PR #444, opened 2025-09-14
fix for colmap import from RC + remove --rc flag and GUI checkbox

kmeirlaen:
More correct colmap import - based on discussion in  https://discord.com/channels/1293599119989805087/1388829600427020318/1416865330969575425

* in colmap import, do not set radial distortion if k=0
* removed RC flag
* show a warning if there is radial_distortion -> stop training and suggest to enable 3DGUT


src/loader/formats/colmap.cpp
simonbethke, line 1: You can only set the camera_model to pinhole if you didnt enter the if block

kmeirlaen: @simonbethke what should be the camera_model in that case?

simonbethke: simple_radial

kmeirlaen: it can be one of these: 
image
there is no code that sets ORTO. It is not "FISHEYE". So PINHOLE seems correct

simonbethke: or better, the one that is named in the file :)

simonbethke: This doesn't fit. There is no enum element for the datatype that you found. SIMPLE_RADIAL (Brown3 params) dont fit on any of those

kmeirlaen: I dont feel adding camera types is in scope of this change

simonbethke: its the unsupported --gut situation. Setting a wrong model type is really bad too. What about setting NULL or add a type UNSUPPORTED?

kmeirlaen: i hear you. but out of scope for this change and we go back to what it was before the --rc flag.
if changes are needed to how GUT is detected, we can look into that too...

simonbethke: okay

MrNeRF:
Ok, once that's build we can merge!

PR #446, opened 2025-09-15
Interop: clear stale CUDA errors

otri:
What: Add a helper `clear_cuda_error_if_present()` calling `cudaGetLastError()` (guarded by `CUDA_GL_INTEROP_ENABLED`) and invoke it at interop failure points to clear sticky CUDA error states.

Why: Prevent cascading failures where subsequent CUDA/Torch calls observe a stale error after a CUDA-GL interop failure, ensuring a clean fallback to the CPU copy path.

Scope: Only `src/rendering/cuda_gl_interop.cpp` (13 insertions, 1 deletion). Behavior unchanged on successful interop; affects only error-handling paths.

### DIFF
--- a/src/rendering/cuda_gl_interop.cpp
+++ b/src/rendering/cuda_gl_interop.cpp
@@ -21,6 +21,14 @@
 
 namespace gs::rendering {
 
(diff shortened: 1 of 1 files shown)


MrNeRF:
Seems uncritical!

PR #448, opened 2025-09-16
Changes to camera rotation

simonbethke:
- A checkbox to set a gimbal lock
- Inverted rotation axes

image

How to test:
Right-click drag to rotate the camera. When moving the mouse in circles on the screen this usually leads to a tilted horizon.

With the "Lock Gimbal" checkbox enabled, the horizon should stay leveled.
When starting a rotation with a tilted horizon but the checkbox checked, the camera will correct its up-direction in the first frame.

### DIFF
--- a/include/core/events.hpp
+++ b/include/core/events.hpp
@@ -56,6 +56,7 @@ namespace gs {
             EVENT(CyclePLY, );
             EVENT(ToggleSplitView, );
             EVENT(ToggleGTComparison, );
+            EVENT(ToggleGimbalLock, bool locked;);
         } // namespace cmd
 
         // ============================================================================
--- a/src/visualizer/gui/panels/main_panel.cpp
+++ b/src/visualizer/gui/panels/main_panel.cpp
@@ -173,6 +173,9 @@ namespace gs::gui::panels {
         // Camera Rotation
         if (ImGui::Checkbox("Lock Gimbal", &settings.lock_gimbal)) {
             settings_changed = true;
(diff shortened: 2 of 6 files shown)


kmeirlaen:
If this works as i think it works, it should be on by default :)
I'll give it a test when i master!

MrNeRF:
Works good, could be a bit smoother on the lock position. We can work later on it.
Code: events from event.hpp in include/core are meant for gui checkboxes etc. I think this will also make the transition to another framework easier in the future.

simonbethke:
I am not sure about enabling it by default, because when you jump to a camera pose, this pose will most likely not be gimbal locked. I personally dont like that the camera realigns on start-rotate. 
There are several ways to work around it:
1) you cannot jump to a camera with gimbal lock on
2) you disable gimbal lock automatically when jumping to a camera
3) keep the current and inconsistent behavior

Currently missing is that when checking the checkbox, the view should automatically get adjusted on click.
Also missing is, that pitch needs to be clamped so you cannout look beyond vertical.

simonbethke:
Furthermore missing is:
When rotating (look around) the mouse-cursor should disappear and reappear at the same position when releasing. This is to ensure that you're mouse doesnt leave the viewport area when looking around.

I tried to implement this, but failed already when trying to hide the cursor. The approached gemini/grok suggested didnt do anything

MrNeRF (approved):
lgtm. Let's wait for the build and merge! Thx!

PR #449, opened 2025-09-16
fix memory leak bugs and get_image_info usage

dozeri83:
1. fixed some memory leak bugs related to not freeing image after load_image
2. switch load_image with get_image_info when load_image was called only for the image size


MrNeRF:
Awesome! Thank you!

PR #450, opened 2025-09-16
Native windows file select

kmeirlaen:
This change replaces the dialog boxes from ImgUI with the native File Dialogs for Windows (Linux not supported).
See video https://www.youtube.com/watch?v=OeK13GDlEow

Replace the default "Open File Browser" with 3 different buttons that are dedicated to their use case.
On Linux, the buttons all do the same (open "file dialog" as before).
On windows, each button will open the native windows dialog and have specific filters (.ply, .lfs or map selection) applied to them.

Also makes .lfs the default file extention for projects.

### DIFF
--- a/src/visualizer/gui/windows/save_project_browser.cpp
+++ b/src/visualizer/gui/windows/save_project_browser.cpp
@@ -23,7 +23,7 @@ namespace gs::gui {
             *p_open = false;
             return true;
         }
-
+        *p_open = false;
         return false;
     }
 #endif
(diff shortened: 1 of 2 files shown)


kmeirlaen:
I'm sure copilot will have lots of suggestions!

MrNeRF:
Breaking change here, but we not that :) Let me know when it is ready for review.

kmeirlaen:
@MrNeRF 

So code is ready and works but could use a review/refactor.

As the "selectFileNative" routine is needed in 3 different panels, i had to define that routine 3 times.  As copilot suggests, this is not very good code practice.
If you can rework it so that we can have a good place to put that routine, that would be great.
Maybe its already possibele to define that routine and all supporting function sin guimanager?  And then call from the panels? But i have not been able to do it.

Or if you can define a routine somewhere that I can call from those 3 places, I can then build on that and finish the PR.

PR #457, opened 2025-09-20
Abort training when an image is detected that is too large to process

kmeirlaen:
Based on feedback from "Flo", we now abort training if the maximum size for an image is exceeded. This avoids that LFS and possibly also the user OS "hangs", forcing an application exit or reboot.

Reference:
> Remember that the fastgs rasterizer implementation is currently limited to images with a maximum of TILE_SIZE * UINT16_MAX = 256 * 2^16 = 16,777,216 pixels which is, e.g., 4096x4096 or 5461x3072 resolution.

When using headless, the folllowing is displayed (depending on the number of worker threads):
image

When using GUI, the following is displayed:
image

### DIFF
--- a/src/core/camera.cpp
+++ b/src/core/camera.cpp
@@ -115,12 +115,6 @@ namespace gs {
         _image_width = w;
         _image_height = h;
 
-        if (w * h > 4096*4096) {
-            std::string error_msg = std::format("Failed to load image - image size too large {} - use a resize factor for this dataset", _image_path.filename().string());
-            LOG_ERROR("Failed to load image - image size too large: (image: {} - w: {} - h: {})", _image_path.filename().string(), w, h);
-            throw std::runtime_error(error_msg);
-        }
-
         // Create tensor from pinned memory and transfer asynchronously
         torch::Tensor image = torch::from_blob(
             data,
--- a/src/training/trainer.cpp
+++ b/src/training/trainer.cpp
@@ -314,6 +314,13 @@ namespace gs::training {
             // Setup camera cache
             for (const auto& cam : base_dataset_->get_cameras()) {
                 m_cam_id_to_cam[cam->uid()] = cam;
+                cam->load_image_size(params.dataset.resize_factor);
+
+                if (cam->image_width() * cam->image_height() > 4096 * 4096) {
+                    std::string error_msg = std::format("image size too large {} - use a resize factor for this dataset", cam->image_name());
+                    LOG_ERROR("Image size too large: (image: {} - w: {} - h: {})", cam->image_name(), cam->camera_width(), cam->camera_height());
+                    throw std::runtime_error(error_msg);                                                       
+                }
             }
             LOG_DEBUG("Camera cache initialized with {} cameras", m_cam_id_to_cam.size());
 


kmeirlaen:
Test: 
* image 4236 x 2833 -> training succeeds
* image 5999 x 3990-> training is aborted

MrNeRF:
I wonder if we should offer to downscale the images instead of throwing an exception.

kmeirlaen:
@MrNeRF as a suggestion (like  we suggest to use --gut when we detect a distorted image) -> eg, "please use --resize_factor" ?
or actually offer and perform a downscale in the same execution?

I do think current behaviour (stopping training) is acceptable - there are other, less severe ways where the application just terminates.  This is before training, no time has been spent by the user and they can easily adjust it when run from the command line or in the GUI

kmeirlaen:
i added a suggestion to use the resize factor

MrNeRF:
Not meant to block this but I think this should be done before the training starts. 
I will look into this.

Btw, image_io.hpp should have a function that let's you ask just the metadata of an image. Width and height is also coming from Colmap data.

kmeirlaen:
I thought about having it in the respective loaders, but I'm unable to test each loader as I dont have data for them.  at training level seemed "convenient" as it was a central place.

no worries about "blocking". still making "small" patches so no biggy if something is not accepted if you dont see it fit

MrNeRF:
@kmeirlaen  I went through the code and I thing it should be checked here:
`    std::expected<void, std::string> Trainer::initialize(const param::TrainingParameters& params) `

You have basically all information in this function including the resize factor and the image height and width. This is much more cleaner than just throwing in each worker which can be may! This initialize function is called on every training start.

kmeirlaen:
@MrNeRF ready for review again!

kmeirlaen:
@MrNeRF as an alternative, we could introduce a new parameter --max-width, which defaults to 3840.
So by default, we would rescale the longest side to 3840.  This way, we would only return an error if the user (accidentally) sets a max width too high.

MrNeRF:
I don't think max_width should be exposed. I think the behavior is not satisfying in general when I think more about it.

Currently all images get downscaled, right? What if you have many lower resolution but only 50 high-res. Only the high res cannot be processed properly. Now you either downscale all or none. 

I would not even throw, just inform the user, and downscale those that need to be! I don't think it is an advantage to cancel the training at this point.

kmeirlaen:
@MrNeRF with a max-width you could mix images
- images below the max width would retain their size
- image above the max width would be resized to that widthµ

For example
max-width=3840 (4K size)
images that are 1920x1080 would be used for training at 1920x1080
images that are 6000x4000 would be used for trining at 3840x2560

i totally dont like the resize-factor, that one indeed scales all the images

resize factor=2
images that are 1920x1080 would be used for training at 960x540
images that are 6000x4000 would be used for trining at 3000x2000

with max-width you have so much more control

btw, this is also what postshot does :)

MrNeRF:
But why expose it. This can be hardcoded. Anyway, if you like to implement it that's fine. Please just make sure that max_width is capped so that the rasterizer does not get in trouble :)

MrNeRF:
Btw, you can get the width and height from the camera directly as they get this per initialization. You just need to apply the scale_factor. This is much cheaper then reading every image meta data from disk.

MrNeRF:
I am closing this as said on discord!

PR #462, opened 2025-09-21
Add Checkbox to enable/disable background modulation to GUI

kmeirlaen:
Added checkbox to visualize and allow acting on the --bg-modulation parameter

image

Also removed the [X] next to the "Settings" and "Training" panels

image


MrNeRF:
lgtm! Thx!

PR #465, opened 2025-09-22
New parameter: --max-width

kmeirlaen:
The max width will limit the size of the images to the given max-width.
When an image is greater than the max width, it will be rescaled before training.  The longest side of the image will match the max-width.

max-width default is 3840
max-width cannot be greater than 4096
max-width must be greater than 0

max-width exposed in GUI under "dataset".
max-width is used for the output of the evaluation images
max-width is saved in project config

### DIFF
--- a/include/core/camera.hpp
+++ b/include/core/camera.hpp
@@ -40,13 +40,13 @@ namespace gs {
         void initialize_cuda_tensors();
 
         // Load image from disk and return it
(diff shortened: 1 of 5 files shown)


MrNeRF:
Ok, looks lgtm! Thank you!
We should create some constants in a follow up pr to not hardcode each number somewhere. It becomes harder to track when you need to change it.

PR #468, opened 2025-09-25
DPI Scaling for GUI Font loading

kmeirlaen:
This change will detect the OS dpi scaling and adjust the font used for displaying accordingly.

Also tried this, but the quality of the font was not very good when upscaling using this method
````
        // Scale fonts according to monitor DPI -> causes issues with pixel-perfect rendering
        // we will scale the font size instead
        // ImGui::GetIO().FontGlobalScale = xscale; 
        // ImGui::GetStyle().ScaleAllSizes(2.0f);
````


MrNeRF:
Awesome. Thanks!

PR #469, opened 2025-09-26
Ply management

dozeri83:
1. added project to scene_manager, added croped ply to project
2. implemented ply renaming

### DIFF
--- a/include/project/project.hpp
+++ b/include/project/project.hpp
@@ -144,6 +144,7 @@ namespace gs::management {
         bool addPly(bool imported, const std::filesystem::path& path, int iter, const std::string& _ply_name);
         void removePly(size_t index);
         void renamePly(const std::string& old_name, const std::string& new_name);
+        bool updatePlyPath(const std::string& ply_name, const std::filesystem::path& new_path);
 
         void clearPlys();
         [[nodiscard]] std::vector<PlyData> getPlys() const;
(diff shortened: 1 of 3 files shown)


MrNeRF:
Tested: 
1. Would be cool to use f2 to change the name.
2. I have to press two times esc to abort the name changing
3. Why is the renaming not propagated to the ply? That's actually the most useful feature.
4. Please remove the llm comments :)

dozeri83:
> Tested:
> 
> 1. Would be cool to use f2 to change the name.
> 2. I have to press two times esc to abort the name changing
> 3. Why is the renaming not propagated to the ply? That's actually the most useful feature.
> 4. Please remove the llm comments :)

1. what do you mean by f2? can you elaborate? I try my best.
2. do you want to rename the stem of the ply file ? so the ply_name will always match the stem ?

MrNeRF:
>     1. what do you mean by f2? can you elaborate? I try my best. 

With f2 I mean the F2 key on the keyboard. That's a standard shortcut for rename files on linux. Isn't this on Windows as well?

>     2. do you want to rename the stem of the ply file ? so the ply_name will always match the stem ?
Exactly. I feel this will be otherwise a bug report because users will expect to see this on disk.

dozeri83:
@MrNeRF  - finished with the fixes.  changing the file name feels a little intrusive, but i guess we will get feedbacks about it in the future.

MrNeRF:
@dozeri83 Tested. Feels very good. I think the renaming is an very nice convenience feature! I will use this a lot definitively!

Issue #456, opened 2025-09-20
Difference in viewing splat after training vs when opening splat


kmeirlaen:
First picture is the view AFTER training
Second picture is the view when opening that same position, but opening the project / .PLY
the quality of the viewer with the .ply is significantly lower
* less contrast
* less sharp (notice the yellow label top left)

Some discussion lichtfeld-sw-dev suggested it might be that the SH are not loaded when a .ply is viewed - this has not been confirmed

<img width="2559" height="1439" alt="Image" src="https://github.com/user-attachments/assets/a8be7583-3532-44f2-b4f2-052fab34d6df" />

<img width="2559" height="1439" alt="Image" src="https://github.com/user-attachments/assets/ded3acf3-ec53-4775-9fcd-7627bde15aa2" />

https://github.com/user-attachments/assets/f5eab2b0-6289-48dd-8df5-e5ab0cb136b9

kmeirlaen:
@MrNeRF any update on this?
I did a test and imported a splat in supersplat and exported it with SH0
I then imported both the original splat (SH3) and exported splat (SH0) into LFS and compared them using the split viewer.  Result: they looked exactly the same in the split viewer.

MrNeRF:
Haven't looked into it yet as working on higher priority. It is not nice but no show stopper imho.
I tagged it. It could be a perfect first good issue for someone starting out. Otherwise I will take care later.

MrNeRF:
Solved with: https://github.com/MrNeRF/LichtFeld-Studio/pull/470

PR #470, opened 2025-09-27
Allow selectino of SH degree in viewport

kmeirlaen:
This change adds a setting for selecting the SH degree when viewing gaussians.
image

SH0:
image

SH3:
image

This also solves
https://github.com/MrNeRF/LichtFeld-Studio/issues/456

### DIFF
--- a/include/core/events.hpp
+++ b/include/core/events.hpp
@@ -137,6 +137,7 @@ namespace gs {
             EVENT(CameraMove, glm::mat3 rotation; glm::vec3 translation;);
             EVENT(SpeedChanged, float current_speed; float max_speed;);
             EVENT(RenderSettingsChanged,
+                  std::optional<int> sh_degree;
                   std::optional<float> fov;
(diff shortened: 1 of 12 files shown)


MrNeRF:
Wouldn’t be a slider better? Btw, how is the bug solved. Does that mean that it simply doesn’t show full shs degree in the viewer after ply loading as default?

kmeirlaen:
@MrNeRF its a nice suggestion for a slider, but drop down will do for me for now... might make a change later...

And yes, loading ply was loading the PLY with SH3, but only displaying on SH0.  I noticed the "_active_sh_degree" when I was looking at the cropbox code and did some digging and then I realized it is never set for PLY that are loaded (and default was 0).  I could have made a simple change and hardcode the "active_sh_degree" to the max_sh the splat has, but I guess that was too easy :)

MrNeRF:
ok great. Ideally we design something that is generally favorable :) But I don't know if a slider is preferred.

PR #475, opened 2025-09-30
add gpu usage to training panel

kmeirlaen:
* add gpu usage to training panel
* added label to the "loss" graph

before:
image

after:
image


MrNeRF:
That’s nice. Looking at the images, there is the loss two times shown. Once should be enough and it is fine to show it in the graph. I also wonder why the values are not equal?

kmeirlaen:
@MrNeRF yes, loss is redundant, i'll remove one.
graph loss is displayed every xx of a second, and the loss text is updated every iteration faster

kmeirlaen:
@MrNeRF found a bug, dont merge

kmeirlaen:
found a bug - will create a new PR when fixed

PR #478, opened 2025-10-01
crop box small modifications

dozeri83:
1. fit crop box text to match app scale resolution,
2. increased box size to 200


MrNeRF:
Thx. Lgtm. The failed pipeline makes no sense. I will retrigger.

PR #480, opened 2025-10-03
Updated CLI Helper Message

SravAku:
- Updated Helper Message for Command Line Interface.
- Added .vs to .gitignore.

### DIFF
--- a/src/core/argument_parser.cpp
+++ b/src/core/argument_parser.cpp
@@ -59,11 +59,11 @@ namespace {
 
         try {
             ::args::ArgumentParser parser(
-                "3D Gaussian Splatting CUDA Implementation\n",
-                "Lightning-fast CUDA implementation of 3D Gaussian Splatting algorithm.\n\n"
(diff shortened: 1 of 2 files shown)


src/core/argument_parser.cpp
```diff
@@ -59,11 +59,11 @@ namespace {
 
         try {
             ::args::ArgumentParser parser(
-                "3D Gaussian Splatting CUDA Implementation\n",
-                "Lightning-fast CUDA implementation of 3D Gaussian Splatting algorithm.\n\n"
+                "LichtFeld Studio: High-performance 3D Gaussian Splatting. \n",
+                "High-performance CUDA implementation of 3D Gaussian Splatting algorithm.\n\n"
```
MrNeRF: Can you please merge these sentences because it is a bit redundant.

MrNeRF (changes requested):

src/core/argument_parser.cpp
SravAku: Updated. Please check the newer commit.

PR #481, opened 2025-10-04
Menu bar

dozeri83:
1. added menu_bar to gui.
2. move all IO operations to File (open project, import dataset, import ply, save)
3. added help item - with "about" and "shortcuts".

also did some refactoring and bug fixes - (added sog to project)

tested in windows.
we need to check that the openUrl works in linux and that i did not broke anything in linux.

image

### DIFF
--- a/include/core/splat_data.hpp
+++ b/include/core/splat_data.hpp
@@ -85,7 +85,7 @@ namespace gs {
         // Export methods - join_threads controls sync vs async
         // if stem is not empty save splat as stem.ply
         void save_ply(const std::filesystem::path& root, int iteration, bool join_threads = true, std::string stem = "") const;
-        void save_sog(const std::filesystem::path& root, int iteration, int kmeans_iterations = 10, bool join_threads = true) const;
+        std::filesystem::path save_sog(const std::filesystem::path& root, int iteration, int kmeans_iterations = 10, bool join_threads = true) const;
 
         // Get attribute names for the PLY format
         std::vector<std::string> get_attribute_names() const;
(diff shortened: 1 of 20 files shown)


MrNeRF:
Thank you. Can't test before Sunday evening but will do it on Sunday.

dozeri83:
> Thank you. Can't test before Sunday evening but will do it on Sunday.

np.  also - you should look at the "about"  window. I am guessing you will want to change/rephrase it.

MrNeRF:
Creator can be in about similar to gplv3 file header, i.e. created by LichtFeld authors.

kmeirlaen:
This is a very nice addition!  One suggestion I would make is to keep the most important actions as buttons in the scene tab, more specifically the "import dataset" button.

dozeri83:
> This is a very nice addition! One suggestion I would make is to keep the most important actions as buttons in the scene tab, more specifically the "import dataset" button.

this is quite a dilemma... I am not frontend developer. on one hand ,Import Ply and Import dataset feel natural in the scene panel and on the other hand, gui real-estate is precious and I don't like duplicated buttons. I feel like we need more feedback on this.

kmeirlaen:
If we need more feedback, then I suggest we put it back as it was before until we have that feedback :)

MrNeRF:
Yeah, maybe let’s bring them back. Sorry, had a long day. Will be back to work tomorrow and will test this pr.

dozeri83:
brought "open ply", "open data" set and "add ply" back
"open/save project" does not belong to scene/training panel

PR #483, opened 2025-10-05
Migrate docs wiki

kmeirlaen:
* Moving docs to wiki by removing key documentation (build instructions / FAQ/ Readme)
* Cleaned up readme
Info that was removed from the readme is now in the [Wiki](https://github.com/MrNeRF/LichtFeld-Studio/wiki)
Also removed info of build instructions and FAQ so that they point to the Wiki

Wiki docs WIP - but I saw that there was a PR for docs in the repo and I would like to motivate users to edit the WIKI pages from now on


MrNeRF:
The files should be removed after a period to have this all only in the wiki.

PR #490, opened 2025-10-07
New parameter --config <json>

kmeirlaen:
New parameter --config <json>
Allows passing a config file with training parameters

Command line only for now.

This change is required to be able to
- the params in the folder params can be the best one (with the lowered pos lr)
- the ability to change the default parameters that are currently used without affecting test/evaluation scripts (the eval script should have the current params and the current param jsons can be moved into the eval folder where the test scripts are located)
- having "profiles" in the GUI (select config from drop down list)

### DIFF
--- a/src/core/argument_parser.cpp
+++ b/src/core/argument_parser.cpp
@@ -454,7 +454,7 @@ gs::args::parse_args_and_params(int argc, const char* const argv[]) {
 
     std::string strategy = params->optimization.strategy; // empty when config files is used and not passed as command line argument
     std::string config_file = params->optimization.config_file;
-    std::filesystem::path config_file_to_read = config_file != "" ? config_file : get_config_path(params->optimization.strategy + "_optimization_params.json");
+    std::filesystem::path config_file_to_read = config_file != "" ? std::filesystem::u8path(config_file) : get_config_path(params->optimization.strategy + "_optimization_params.json");
 
     if (!parse_result) {
         return std::unexpected(parse_result.error());


src/core/parameters.cpp
```diff
@@ -485,6 +485,35 @@ namespace gs {
             return params;
         }
 
+        /**
+         * @brief Read optimization parameters from JSON file
+         * @param[in] path Path to the JSON file (can be absolute or relative)
+         * @return Expected OptimizationParameters or error message
+         */
+        std::expected<OptimizationParameters, std::string> read_optim_params_from_custom_json(const std::string path) {
```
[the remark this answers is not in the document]
MrNeRF: I second that!

src/core/argument_parser.cpp
```diff
@@ -422,12 +438,27 @@ gs::args::parse_args_and_params(int argc, const char* const argv[]) {
...
+        if (!opt_params_result) {
+            return std::unexpected(std::format("Failed to load optimization parameters: {}",
+                                               opt_params_result.error()));
+        }
+        params->optimization = *opt_params_result;
+    }
+
+    // if a config file was used and strategy was also passed as command line argument, ensure they match 
+    if (config_file != "" && strategy != "" && strategy != params->optimization.strategy) {
+        LOG_WARN("Conflict between strategy in config file and --strategy on command line ");
```
MrNeRF: Should be LOG_ERROR

src/core/parameters.cpp
MrNeRF: This is necessary. I recommend to reuse read_optim_params_from_json and add a name if that's the custom file

`read_optim_params_from_json(const std::string& file_name)`

So you can provide the correct filename upon calling this function. Otherwise this is simply duplicated.

src/core/argument_parser.cpp
```diff
@@ -422,12 +438,27 @@ gs::args::parse_args_and_params(int argc, const char* const argv[]) {
         std::exit(0);
     }
 
-    auto opt_params_result = gs::param::read_optim_params_from_json(params->optimization.strategy);
-    if (!opt_params_result) {
-        return std::unexpected(std::format("Failed to load optimization parameters: {}",
-                                           opt_params_result.error()));
+    if (config_file != "") {
```
MrNeRF: Given that you implement the comment below (assemble name before calling) you can remove this if-else clause.

MrNeRF (changes requested):
Please see my comments. A lot of code can be simplified.

kmeirlaen:
* moved code for getting the default param file to argument parser
* adjusted read_optim_params_from_json to take file as argument instead of strategy
* removed the read_optim_params_from_custom_json function
* removed duplicate code for loading json

MrNeRF:
Much cleaner. Thank you! Lgtm!

PR #491, opened 2025-10-07
Copy params json files to scripts / adjust scripts / adjust .json file Position Learning Rate

kmeirlaen:
* copy params json files to scripts 
* adjust scripts 
* adjust .json file Position Learning Rate

tested on Windows (powershell scripts)


scripts/benchmark_mipnerf360.sh
```diff
@@ -40,7 +40,7 @@ do
         --eval \
         --headless \
         --save-eval-images \
-        --strategy mcmc
+        --config ./scripts/mcmc_optimization_params.json
```
MrNeRF: "scripts/mcmc_optimization_params.json" should be enough

MrNeRF (commented):
How about renaming the scripts folder into eval?

eval/benchmark_mipnerf360_mcmc.sh
```diff
@@ -40,7 +40,7 @@ do
         --eval \
         --headless \
         --save-eval-images \
-        --strategy mcmc
+        --config .eval/mcmc_optimization_params.json
```
MrNeRF: The dot must be removed. Should be
        --config eval/mcmc_optimization_params.json
Otherwise it will fail to load it.

MrNeRF: @kmeirlaen This is the remaining issue. Otherwise it is good to go!

kmeirlaen: @MrNeRF oops, good catch, corrected!

miriameng:
@kmeirlaen @MrNeRF why is the `means_lr` in `eval` 0.00016 but the `means_lr` in `parameter` is 0.000016? One of them gotta be wrong, right?

MrNeRF:
The eval folder has the standard params for evaluation. The parameter folder contains better parameters for robust training. Possibly they will change.

PR #492, opened 2025-10-08
shortcuts and control additions

dozeri83:
1. added more shortcuts and controls. added "state" column to shortcut.
2. fixed a log bug (print was done after the remove)


MrNeRF:
lgtm! Thx!

PR #496, opened 2025-10-15
Update index on README.md to match current page layout

SharkWipf:
Another really minor documentation patch I just happened across, on the current README.md half the index anchors don't exist anymore. This updates the list.


MrNeRF:
Thx!

PR #498, opened 2025-10-16
hierarchical cache 

dozeri83:
1. made hierarchical cache RAM -> FS. loaded images are stored in cpu memory, if no memory available they are stored in temporary file sytem.

2.  memory cache has safety thresholds to not get close to overflowing RAM:
       float min_cpu_free_memory_ratio_ = 0.1f; // make sure at least 10% RAM is free
        std::size_t min_cpu_free_GB_ = 1;        // min GB we want to be free 
3. cpu_cache and fs_cache flags are exposed in arg parser and gui.
4. expected imporvements:
        if the images are not scaled or preprocessed - got x3 improvements on varaious dataset from cpu cache.
        if the images are scale - x4-5 is expected.
        using only files system (and scale)- x2 improvement is expected. 
        
  5. file system cache is emptied each time the training is starting

### DIFF
--- a/src/loader/cache_image_loader.cpp
+++ b/src/loader/cache_image_loader.cpp
@@ -422,7 +422,12 @@ namespace gs::loader {
                 size_t giga_to_bytes = 1024ULL * 1024 * 1024;
                 double required_GB = static_cast<double>(required_bytes) / static_cast<double>(giga_to_bytes);
                 double available_GB = static_cast<double>(get_available_physical_memory()) / static_cast<double>(giga_to_bytes);
-                LOG_INFO("not all images fit to cpu memory required_GB {:.2f}. available {:.2f}", required_GB, available_GB);
+
+                if (use_cpu_memory_) {
+                    LOG_INFO("not all images fit to cpu memory required_GB {:.2f}. available {:.2f}", required_GB, available_GB);
+                } else {
+                    LOG_INFO("skip cpu memory cache by user");
+                }
 
                 auto [org_width, org_height, org_channels] = get_image_info(path);
                 if (use_fs_cache_ && (params.resize_factor > 1 || params.max_width < org_width)) {


include/loader/cache_image_loader.hpp
```diff
@@ -0,0 +1,128 @@
...
+#pragma once
+
+#include <filesystem>
+#include <memory>
+#include <mutex>
+#include <set>
+#include <vector>
+#include <unordered_map>
+
+namespace gs::system {
```
MrNeRF: Why system? This should be gs::loader

src/core/parameters.cpp
```diff
@@ -505,12 +505,7 @@ namespace gs {
                 nlohmann::json json;
 
                 // Dataset configuration
-                json["dataset"]["data_path"] = params.dataset.data_path.string();
-                json["dataset"]["output_path"] = params.dataset.output_path.string();
-                json["dataset"]["images"] = params.dataset.images;
-                json["dataset"]["resize_factor"] = params.dataset.resize_factor;
-                json["dataset"]["test_every"] = params.dataset.test_every;
-                json["dataset"]["max_width"] = params.dataset.max_width;
+                json["dataset"]= params.dataset.to_json();
```
MrNeRF: Why is this not in the parameters?

src/loader/cache_image_loader.cpp
```diff
@@ -0,0 +1,463 @@
...
+#define NOMINMAX
+#include <windows.h>
+#elif defined(__APPLE__)
+#include <mach/mach.h>
+#include <mach/mach_host.h>
+#include <mach/vm_statistics.h>
+#include <sys/sysctl.h>
+#endif
+
+namespace gs::system {
```
MrNeRF: gs::loader

src/loader/cache_image_loader.cpp
```diff
@@ -0,0 +1,463 @@
...
+
+#include <fstream>
+
+// Platform-specific includes
+#ifdef __linux__
+#include <sys/sysinfo.h>
+#elif defined(_WIN32)
+#define NOMINMAX
+#include <windows.h>
+#elif defined(__APPLE__)
```
MrNeRF: Why  apple?

src/loader/CMakeLists.txt
```diff
@@ -45,28 +46,28 @@ target_include_directories(gs_loader
 )
 
 # Find the actual WebP library target
-if(TARGET WebP::webpdecoder)
```
MrNeRF, line 48: Can you please revert the subsequent changes?

include/loader/cache_image_loader.hpp
dozeri83: i think those methods are more general than loader and might be used by other modules. 
I will try to put them in a more general place. any suggestions?

src/core/parameters.cpp
dozeri83: not sure what you mean by this

dozeri83:
> There is a lot AI slop. I don't have anything against using llms but it should be cleaned like all the apple stuff, etc which is none sense.
> 
> More severe, the behavior should be:
> 
> 1. Do the images fit into ram -> cache in ram
> 2. Don't fit, only cache on fs if and only if they are resized otherwise don't cache anything.
> 
> If the images don't fit into ram, you will have constant trashing by eviction which makes it even worse than now. If the images are not preprocessed, the behavior will waste disk space.
> 
> Something like this: const bool use_cpu_cache = (estimated_dataset_size < available_ram * 0.9 (maybe 0.8?)); const bool use_fs_cache = (resize_factor > 1 || max_width < max supported resolution);
> 
> This should be set based on the auto detection in gui.

dont you want a partial use of the RAM even if not all images fit? we could benefit from this as well even if 50% of images are cached.
(think of edge cases where 99% fit)

include/loader/cache_image_loader.hpp
MrNeRF: They should be in loader. Loader is already quite general.

src/core/parameters.cpp
MrNeRF: Should go into https://github.com/MrNeRF/LichtFeld-Studio/tree/master/parameter
and here https://github.com/MrNeRF/LichtFeld-Studio/tree/master/parameter

MrNeRF:
> dont you want a partial use of the RAM even if not all images fit? we could benefit from this as well even if 50% of images are cached.
> (think of edge cases where 99% fit)

Makes no sense because you will get the eviction overhead on top if not all images fit into ram. The trainer consumes images in complete random order over every batch size which gains nothing in this case.

dozeri83:
> > dont you want a partial use of the RAM even if not all images fit? we could benefit from this as well even if 50% of images are cached.
> > (think of edge cases where 99% fit)
> 
> Makes no sense because you will get the eviction overhead on top if not all images fit into ram. The trainer consumes images in complete random order over every batch size which gains nothing in this case.

i believe that above certain occupancy (which should be small) the price you pay for eviction is small relatively to the gain from 
cache hits. 
the price of cache miss is loading + allocation, and you have to do the loading anyway (allocation should be fast).

For now i will check if all images fit to memory in each iteration.
 If yes - cacheloader will be in memory state - If no we will be in fs state.  
 But I think this is sub optimal.

src/core/parameters.cpp
dozeri83: i just made datasetConfig json serializable in this code.
the folder you referred to  is related to optimizationParams. 

if what you are asking about is loadingParams, I felt that the natural place for LoadingParams is datasetConfig  and not optimizationParams.

include/loader/cache_image_loader.hpp
dozeri83: fixed it

dozeri83:
@MrNeRF  - resolved your required remarked. (besides the parameter one which is still not clear - lets talk in discord if you want to clarify)

introduce cache status which is determined in the start of the training.  if in the middle of the training memory will exceed 
thresholds images will be removed from cache but we will still be in cpu memory mode, since this change is assumed to be temporary (also - change this mode adaptively makes things ugly)

PR #499, opened 2025-10-16
fix compilation for CUDA 13 + libtorch 2.9

miriameng:
In CUDA 13, cub::Max() no longer exists, need to use cub::Max{}, because Max is now a struct template with a default constructor


MrNeRF:
Works also with CUDA 2.8. Alright! Thx!

PR #512, opened 2025-10-23
Lfs core

MrNeRF:
Migration to new tensor lib 1.
Old code still exists.


MrNeRF:
@miriameng Can you help with this build? There is some include or so that Windows doesn't like and I have no clue. The linux build is fine but the pipeline fails because it has not enough space. Let's ignore that because I am currently removing libtorch and switching to my tensor lib. This will solve many problems.

PR #517, opened 2025-10-24
nvidia_smi detection issue

lyehe:
When CMake runs as a 32-bit process, WOW64 file system redirection can prevent it from finding `nvidia-smi.exe` in `System32`, causing CMake to incorrectly assume no NVIDIA GPU is present and default to compute capability 8.6 and cause build failure.


MrNeRF:
Thx!

PR #525, opened 2025-11-08
Mask implementation and some feature improvements

lyehe:
# Mask Support, Config Export/Import, and Enhanced Parameters

## Summary

Adds mask support for training, configuration export/import, enhanced parameter configurability, and automatic 3DGUT detection.

### Mask System
4 mask modes: None, Segment (soft penalty), Ignore (exclude from loss), AlphaConsistent (exact alpha). Configurable inversion, threshold clamping, and opacity penalty. GUI includes mask mode selector, real-time preview with overlay, and parameter controls.

New flags: --mask-mode, --invert-masks, --mask-threshold, --mask-opacity-penalty-weight, --mask-opacity-penalty-power
You sure this is the right branch to pr?

### Configuration Export/Import
Export/import optimization parameters to/from JSON via File menu (Ctrl+E/Ctrl+I). Cross-platform file dialogs with 3-second status feedback.

### Enhanced Parameters
Now configurable: --prune-opacity, --gut, --bg-modulation, --gpu-id, --final-lr-fraction, --bilateral-grid-warmup-steps, --bilateral-grid-warmup-start-lr

Fixed missing JSON serialization for gut and bg_modulation. All 68 optimization parameters now properly exported/imported. Updated all 4 default config JSONs.

### Automatic 3DGUT Detection
Detects camera distortion (radial/tangential) when loading datasets. Auto-enables 3DGUT and shows warning popup immediately on dataset load (not at training start). Updates project parameters automatically.

### Cross-Platform File Dialogs
Native dialogs on Windows, ImGui-based on Linux. Unified interface for config, project, PLY, and dataset operations.

## Technical Changes

Moved 3DGUT detection from trainer.cpp::initialize() to training_manager.cpp::setTrainer() - now triggers on dataset load for immediate feedback.

Optimized mask validation from full loading to existence checks.

36 files modified, 2 new files: src/visualizer/gui/utils/file_dialogs.{cpp,hpp}

## Breaking Changes
None - fully backwards compatible.

## Testing
Test all mask modes, config export/import, 3DGUT auto-detection with distorted datasets, new CLI parameters, cross-platform file dialogs, and parameter serialization roundtrip.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,7 @@
-<div align="center"><img src="src/visualizer/gui/assets/logo/lichtfeld-logo-white.svg" alt="LichtFeld Studio" height="60"></div>
+<div align="center"><picture>
+  <source media="(prefers-color-scheme: dark)" srcset="src/visualizer/gui/assets/logo/lichtfeld-logo-white.svg">
+  <img src="src/visualizer/gui/assets/logo/lichtfeld-logo.svg" alt="LichtFeld Studio" height="60">
+</picture></div>
 
 <div align="center">
 **A high-performance C++ and CUDA implementation of 3D Gaussian Splatting**
@@ -20,6 +23,18 @@
 [**Acknowledgments**](#acknowledgments) •
 [**Citation**](#citation) •
 [**License**](#license)
+
+</div>
+
+---
+
+## Sponsors
+
+<div align="center">
+
+<a href="https://www.core11.eu/">
+  <img src="docs/media/core11_multi.svg" alt="Core 11" height="60">
+</a>
 
 </div>
 
--- a/docs/media/core11_multi.svg
+++ b/docs/media/core11_multi.svg
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Generator: Adobe Illustrator 27.9.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+
+<svg
+   version="1.1"
+   id="Layer_1"
+   x="0px"
+   y="0px"
+   viewBox="0 0 331.53999 61.950003"
+   xml:space="preserve"
+   sodipodi:docname="core11.svg"
+   width="331.54001"
+   height="61.950001"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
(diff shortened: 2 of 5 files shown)


kmeirlaen:
some nice changes - more than what is described:
- also exposes num workers in the config

some things I could not confirm
- CTR+E / CTR+I for the import / export (file export/import did work)

something i would like to see different, but not a showstopper imo, more a suggestion

- have the default location of the import/export config to the "parameters" directory
- show the "config loaded/saved" confirmation somewhere else.  now it is at the bottom of the "training" panel, where it even might be off-screen or not shown as the tab might not be selected - also, the light green color is not very readable on the grey background
 
no showstoppers

kmeirlaen:
I'm not a fan of enabling GUT by default.  The user should actively do this as from what I understand, viewers will not be compatible with the .ply that is generated using GUT

lyehe:
> I'm not a fan of enabling GUT by default.  The user should actively do this as from what I understand, viewers will not be compatible with the .ply that is generated using GUT

This is on enabling for images with distortion. The trainer does not work without-gut flag anyway. It does not enable gut for undistorted images. I think the proper way is to undistort on the play like gsplat.

kmeirlaen:
> > I'm not a fan of enabling GUT by default.  The user should actively do this as from what I understand, viewers will not be compatible with the .ply that is generated using GUT
> 
> This is on enabling for images with distortion. The trainer does not work without-gut flag anyway. It does not enable gut for undistorted images. I think the proper way is to undistort on the play like gsplat.

Yes, I understand that it would only be enabled when LFS detects images with distortion. Still, not a fan of doing that automatically...

opsiclear-admin:
> > > I'm not a fan of enabling GUT by default.  The user should actively do this as from what I understand, viewers will not be compatible with the .ply that is generated using GUT
> > 
> > 
> > This is on enabling for images with distortion. The trainer does not work without-gut flag anyway. It does not enable gut for undistorted images. I think the proper way is to undistort on the play like gsplat.
> 
> Yes, I understand that it would only be enabled when LFS detects images with distortion. Still, not a fan of doing that automatically...


Btw, there is a misconception. --gut is fully compatible with all viewers.. 

3dgrut is producing different output, but gut is basically distort the regular 3dgs output (during rendering, not after rendering), so there should not be any compatibility issue

MrNeRF:
Sorry for the delay. I promise I will test and review it over the weekend.

MrNeRF:
I just yolo it. Thank you :)

PR #533, opened 2025-11-13
Fix COLMAP flat-structure detection for paths with trailing slashes

SharkWipf:
There was a bug where a trailing `/` after a dataset path would trip up the flat tree check, due to the paths not evaluating as equal.
This fixes that, plus an unverified side-issue with symlinks.
I do not know if this works on Windows, it should, but I haven't tested. 

Full disclosure: This was written by Codex.
I've verified and tested it, it looks alright to me, but my C++ knowledge is limited to non-existent, I don't know if the way it handled things here is the optimal way.

In Codex' own words:

>   - Problem: flat-structure detection failed when the dataset path ended with a trailing slash, so installs reported “Images directory 'images' not found” even though images lived at the dataset root.                                                                
>   - Solution: compare camera-parent directories to the dataset root via a new is_dataset_root helper that uses std::filesystem::equivalent (with a safe fallback), ensuring both paths are treated as the same location regardless of trailing separators.

### DIFF
--- a/src/loader_new/loaders/colmap_loader.cpp
+++ b/src/loader_new/loaders/colmap_loader.cpp
@@ -14,6 +14,7 @@
 #include <chrono>
 #include <filesystem>
 #include <format>
+#include <system_error>
 
 namespace lfs::loader {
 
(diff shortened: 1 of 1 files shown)


MrNeRF:
Thx. I promise to review it over the weekend!

SharkWipf:
It's the lowest of low priorities, I'm in no rush 😄

PR #534, opened 2025-11-14
Fix dataset load crash when loading unsupported formats from GUI

### LINKED ISSUE #471: LichtFeld Studio exits on unknown data
When attempting to load a wrong folder as colmap data, lichtfeld-studio just exits. I assume this is behavior which was implemented and fine for CLI usage, because in the CLI you get valueable feedback. But in GUI without any console visible there is no feedback.

Best would be an alert box with a message "Unrecognized dataformat".

francescofugazzi:
## Summary

This PR fixes a crash that occurred when loading unsupported dataset formats from the GUI.  
Instead of allowing the exception to terminate the application, the GUI now displays a modal error popup.

## Changes

- Added a try catch block in the dataset path of `DataLoadingService::handleLoadFileCommand`
- On failure, the code emits `events::state::DatasetLoadCompleted` with `success = false` and the error message
- `GuiManager` now listens to `DatasetLoadCompleted` and shows a modal ImGui popup when loading fails
- Introduced a simple user friendly error dialog without modifying the existing dataset loading logic
- Console and CLI dataset loading behavior remains unchanged

## Notes

The popup uses the same modal style already used in the application and does not affect normal loading flows.

Fixes #471

### DIFF
--- a/src/visualizer/core/data_loading_service.cpp
+++ b/src/visualizer/core/data_loading_service.cpp
@@ -28,7 +28,19 @@ namespace gs::visualizer {
 
     void DataLoadingService::handleLoadFileCommand(bool is_dataset, const std::filesystem::path& path) {
         if (is_dataset) {
-            loadDataset(path);
+            // Prevent GUI crash on unsupported dataset formats
+            try {
+                loadDataset(path);
+            } catch (const std::exception& e) {
+                events::state::DatasetLoadCompleted{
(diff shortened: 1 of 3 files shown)


MrNeRF:
@francescofugazzi Thank you! I have looked at the code and @kmeirlaen has tested this pr. It looks good!

PR #539, opened 2025-11-27
Windows one shot build script

lyehe:
tested on 2 pcs


MrNeRF:
That's sounds like a big enough focus group. Thx 🙂

PR #550, opened 2025-12-15
remove libtorch and a ton of improvements

MrNeRF:


miriameng:
Impressed by the new UI!

PR #555, opened 2025-12-20
Windows: fix cl.exe not in PATH when running nvcc directly with execute_process

miriameng:
cmake won't work without this PR on my windows environment


miriameng:
Also cl.exe not being in PATH I think is by design, so multiple different versions of visual studio can co-exist.

miriameng:
Not sure what happened to the Ubuntu CI, it was still functioning like half a day ago..

MrNeRF:
just sync, it should work.

MrNeRF:
You should sync. CI should work again.

miriameng:
Thanks, CI passing now

MrNeRF:
thx, lgtm

PR #556, opened 2025-12-20
add --bg-noise flag for per-pixel random background

MotivaCG:
This commit introduces the `--bg-noise` flag, which applies random noise to the background on a per-pixel basis.

Key behaviors:
- Enabling `--bg-noise` overrides and disables `--bg-modulation`.
- If the flag is not set, the pipeline remains unchanged.

Use cases:
- `--bg-modulation` is typically preferred for interior scenes.
- `--bg-noise` is often better for masked objects to improve separation from the background.


MrNeRF:
can you sync with master? I can test this later.

MotivaCG:
Done

MrNeRF:
We had a lengthy discussion on discord about this pr. Unfortunately this is not correctly implemented and doesn't do anything. I will add a correct implementation later on.

PR #567, opened 2025-12-21
Fix regex to take only first GPU if multiple GPUs present

ege-adam:
Fix regex to take only first GPU if multiple GPUs present, so we can find compute capability


MrNeRF:
Lgtm! Thx

PR #571, opened 2025-12-22
Add --skip-intermediate flag to disable intermediate saves

MotivaCG:
Add new command-line flag --skip-intermediate that disables all intermediate PLY and checkpoint saves during training. When enabled:

- Intermediate saves at save_steps are skipped
- Checkpoint (.resume) files are never written
- Only the final PLY is saved at the end of training

This is useful for batch and CLI training runs where intermediate outputs are not needed, reducing disk I/O and storage usage.


MrNeRF:
This was removed on purpose. You can specify the save interval either in the params or via GUI. I close this to prevent flag creep.

PR #572, opened 2025-12-22
Auto-create cache base directory if missing

MotivaCG:
Instead of failing with an error when the LichtFeld temp folder doesn't exist, automatically create it. This fixes the "Cache base path missing" error that occurs on fresh installations or when the temp directory has been cleaned.


MrNeRF:
This one sneaks in again skip_intermediate. Please clean it

MotivaCG:
Yes, my bad, done!

PR #581, opened 2025-12-23
mip 2d filter

MrNeRF:
improved mip filter provided by flo


miriameng:
Is mip filter from https://arxiv.org/abs/2311.16493?

MrNeRF:
> Is mip filter from https://arxiv.org/abs/2311.16493?

Yep. Just a bit more stable go train implementation

PR #584, opened 2025-12-24
Button states

kmeirlaen:
Small change to logic of how buttons are displayed:
* allow clear in initial state
* in "paused", remove clear, replace with "stop training"
* in "stopped", add "clear" button

this allows the user to stop the training and go to "edit" mode at any point in the training process.
Use case: user sets 30K seps, but after 15K wants to stop, edit and save the splat at that point.


MrNeRF:
Lgtm, thx!

PR #585, opened 2025-12-24
Make scene context take training state into account

kmeirlaen:
When in "dataset" mode, the scene menu shows "Add Ply..." and "New Group".  Both these actions are not useful in case we are not in editing mode:
* add ply does not work as no .ply should be added when in dataset mode
* add group has no use as no .ply can be added in the group and no merging can be performed

This PR checks if we are in training mode or editing mode and shows/hides the menu items so that a user can only perform the action when it actually is possible

### DIFF
--- a/src/visualizer/gui/panels/scene_panel.cpp
+++ b/src/visualizer/gui/panels/scene_panel.cpp
@@ -270,17 +270,19 @@ namespace lfs::vis::gui {
         const bool blnTraininMode = m_trainerManager->hasTrainer();
 
         if (ImGui::BeginPopupContextItem("##ModelsMenu")) {
-            if (ImGui::MenuItem("Add PLY...", nullptr, false, !blnTraininMode)) {
-                cmd::ShowWindow{.window_name = "file_browser", .show = true}.emit();
+            if (!blnTraininMode) {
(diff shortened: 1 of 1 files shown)


kmeirlaen:
seems my commits are bit fucked up since i reverted a commit - will try to fix for next PR

src/visualizer/gui/panels/scene_panel.cpp
```diff
@@ -266,18 +266,23 @@ namespace lfs::vis::gui {
 
         // Context menu for folder
         theme().pushContextMenuStyle();
+
+        const bool blnTraininMode = m_trainerManager->hasTrainer();
+
         if (ImGui::BeginPopupContextItem("##ModelsMenu")) {
-            if (ImGui::MenuItem("Add PLY...")) {
-                cmd::ShowWindow{.window_name = "file_browser", .show = true}.emit();
+            if (!blnTraininMode) {
```
MrNeRF, line 273: You could just use here  m_trainerManager->hasTrainer();
so this never get queried         const bool blnTraininMode = m_trainerManager->hasTrainer();
when  if (ImGui::BeginPopupContextItem("##ModelsMenu")) { should be false

PR #672, opened 2026-01-02
Fix info panel overlapping image in preview (fixes #670)

### LINKED ISSUE #670: image preview is overlapped with image
That needs to be fixed
Image

jacobvanbeets:
1. Reserve space for the info panel when it's visible (subtracting the panel width + margins from available width)
2. Center the image in the remaining space (not the full window)
3. Prevent overlap at any window size
4. The window can be resized.
Closes issue: https://github.com/MrNeRF/LichtFeld-Studio/issues/670


MrNeRF:
lgtm! thx!

PR #685, opened 2026-01-03
BUILD_TESTS=ON on CI

### LINKED ISSUE #677: BUILD_TESTS=ON in CI?
Although we can't run tests on CI due to no GPU, but I think it will be a good idea to at least check whether the tests compile. Do we agree?

miriameng:
resolve #677

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -68,6 +68,21 @@ jobs:
           set(VCPKG_MAX_CONCURRENCY 2)
           EOF
 
+      - name: Cache LibTorch
+        id: cache-libtorch
+        uses: actions/cache@v4
+        with:
+          path: external/libtorch
+          key: libtorch-2.7.0-cu128-${{ runner.os }}
+
+      - name: Download LibTorch 2.7.0 (CUDA 12.8 build)
+        if: steps.cache-libtorch.outputs.cache-hit != 'true'
+        run: |
+          mkdir -p external
+          cd ..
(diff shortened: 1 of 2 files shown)


miriameng:
hmm no space left on device ... why didn't we encounter this issue before when libtorch was a required dependency?

MrNeRF:
Right, forgot. That was a problem on linux builds before. It ocassionally failed running out space.

All of it could be freed from libtorch. However, it is convenient to make sure the tensor implementation works correctly.

dozeri83:
just letting you know - on my windows desktop, I got int /size_t casting  errors from Tensor::from_vector when i try to build tests.
After fixing them, i got load dll error code from lichtfeld_tests.exe ...

miriameng:
Run out of space in linux CI, and failed on windows CI because we have one single test exe for all GPU and non GPU tests and we are running the test exe on Windows CI.

@dozeri83 the problem you described should be fixed in the latest master

MrNeRF:
Ok, I am not sure how feasible building the tests in the limited CI pipeline is? Wdyt?
There are some separate unicode tests that only run on windows.

miriameng:
@MrNeRF how about splitting the test exe into 2, one GPU and one non GPU? At least we can still compile all the tests on windows CI and only run the non GPU unicode tests like usual.

MrNeRF:
@miriameng Not sure what value that brings. Building would be some value but then again it has libtorch issues. Pretty much everything critical runs on GPU. I would like to run the evals for every pr that touches the training to make sure that that it is not broken. Apparently, this is not doable with github. I will try to get an infrastructure sponsor over the next months which will hopefully get us a better build node. I think we should postpone this until I have something.

MrNeRF:
I will enable some cpu  tests for now.

Issue #690, opened 2026-01-04
I like this project and I want to add Chinese language support to it. How can I do that?


hanlanyimeng:
I like this project and I want to add Chinese language support to it. How can I do that?

MrNeRF:
Hey! Here is a generated translation https://github.com/MrNeRF/LichtFeld-Studio/blob/master/src%2Fvisualizer%2Fgui%2Fresources%2Flocales%2Fzh.json
Awhich you can activate in the gui? Is it wrong or not good? You can contribute by improving it.

todashuta:
FYI, the zh.json file has been present since last week, but its JSON syntax was broken, so it didn’t appear in the GUI.

The JSON syntax issue was fixed in this commit:
https://github.com/MrNeRF/LichtFeld-Studio/commit/516a720b73fc19f2802768ae088005f4e35c6e79

<img width="550" height="273" alt="Image" src="https://github.com/user-attachments/assets/c5fef1f6-a120-4b6e-acd7-f6ba706b6e9b" />

Therefore, translations provided by zh.json should be available starting with the nightly build `LichtFeld‑Studio‑windows‑nightly‑2026‑01‑04‑8bbb720.zip` and later.

hanlanyimeng:
Thanks
Yes, the current version in use is LichtFeld-Studio-windows-nightly-2026-01-03-f06ce6b. I have downloaded the zip package LichtFeld‑Studio‑windows‑nightly‑2026‑01‑04‑8bbb720.zip, which comes with Chinese language support, but there is garbled text as shown in the figure.

<img width="1247" height="1302" alt="Image" src="https://github.com/user-attachments/assets/f8a726c2-d4c2-4dad-a48e-f3980c73c77d" />
@MrNeRF @todashuta

MrNeRF:
If you are able to fix that it would be very helpful.

hanlanyimeng:
@MrNeRF 
I’m not familiar with the code, but I’ve found the root cause and relevant references – it seems to be a problem with the ImGui version.
<img width="1568" height="981" alt="Image" src="https://github.com/user-attachments/assets/96eedc78-6518-475f-8b5f-612beaa3c6e5" />
here the link:https://github.com/ocornut/imgui/blob/25158fe33bd22871eab47b4a06f951fd6655fad8/docs/FAQ.md#q-how-can-i-display-and-input-non-latin-characters-such-as-chinese-japanese-korean-cyrillic

---
[...]

PR #697, opened 2026-01-05
Improve Chinese language support

todashuta:
Improved Chinese language support.

See #690


MrNeRF:
@todashuta Thank you for your fix!
I just figured out the Japanese fonts already includes all Chinese character. So I changed it to add Chinese glyph ranges to the existing Japanese font load - no need for a separate 11MB font file. Hope that works.

todashuta:
@MrNeRF Thank you for the review and the merge!
I built and tested 17a7f46 on my own machine, and everything works fine.

Issue #699, opened 2026-01-05
Sparsification steps added twice when using GUI


kmeirlaen:
Issues with dynamically calculated sparsification steps:
1. steps are added 2x 
* open LFS
* load dataset
* from training, basic settings, enable sparsification
** steps is changed from 30000 to 45000
* start training
** traing progress shows 60000 steps

2. incosistent removal of steps
* open LFS
* load dataset
* from training, basic settings, enable sparsification
** steps is changed from 30000 to 45000
* scale down the iterations to 10000 using +/- buttons in GUI
* disable sparsify
** iterations REMAIN at 10000
* enable sparsfiy
** iterations increased with 15000 (to 25000)
* disable sparsify
** iterations CHANGE to 10000 (inconsistent with previous behaviour)

3. incosistant calculation of steps
* open LFS
* load dataset
* from training, basic settings, enable sparsification
** steps is changed from 30000 to 45000
* scale down the iterations to 1000 using +/- buttons in GUI
* manually enter the number "200" in sparsfication steps
** iterations change to 1198

Solution: dont increase/decrease iterations when changing sparsification steps

PR #700, opened 2026-01-05
remove dynamic calculation of steps when chaning sparsity parameters

kmeirlaen:
Fix for issue #699


kmeirlaen:
There is still an issue when importing a .resume file -> the iterations will be set to 45000 and when sparsify is unchecked, it remains 45000.
Should be iterations - sparsify steps

kmeirlaen:
.resume file handled properly now

Issue #595, opened 2025-12-25
Embedded python


MrNeRF:

shadygm:
#716 Is a first step. I created this PR because in the coming month I will be quite busy so I am unsure when I will be able to finish the work. Its a stepping stone in case anyone else wants to keep working on it in the meantime.

PR #716, opened 2026-01-06
#595 Python Integration

shadygm:
This is a temporary PR for the python integration until it gets completed (I'll add details explaining the features/how it works as soon as I'm done with it).


MrNeRF:
ok thank you a lot! I will take a look likely not before next week

PR #720, opened 2026-01-07
Clean up compiler warnings (primarily on Windows)

MotivaCG:
## PROBLEM
The build was generating hundreds of warnings on MSVC, creating so much noise that actual issues could easily go unnoticed during compilation. 

This PR eliminates the excessive warning clutter, particularly on Windows, while maintaining cross-platform compatibility. While some warnings (shadowing, nodiscard, unused parameters) would also appear with strict GCC/Clang flags, others are MSVC only due to stricter conversion checks and CUDA stdlib implementation differences.

Additionally RelWithDebInfo and MinSizeRel build configurations were broken on Windows (Release and Debug were ok).

### Build System
- Fix RelWithDebInfo and MinSizeRel CMake targets on Windows
- Suppress benign warnings from CUDA/MSVC standard library headers (Windows only)

### Code Fixes
- Fix LOG_TIMER macro name collisions using __ COUNTER __ (cross-platform) instead of __ LINE __
- Add explicit type casts for conversions (helps on all platforms)
- Explicitly discard [[nodiscard]] return values (cross-platform)
- Mark intentionally unused parameters (cross-platform)

### DIFF
--- a/src/io/formats/sogs.cpp
+++ b/src/io/formats/sogs.cpp
@@ -1085,7 +1085,7 @@ namespace lfs::io {
         auto scales = splat_data.scaling_raw().cpu();
         const auto* scales_ptr = scales.ptr<float>();
 
-        auto scale_result = cluster1d(scales_ptr, num_rows, 3, options.kmeans_iterations);
+        auto scale_result = cluster1d(scales_ptr, static_cast<int>(num_rows), 3, options.kmeans_iterations);
 
         std::vector<uint8_t> scales_data(width * height * CHANNELS, 0);
         for (int64_t i = 0; i < num_rows; ++i) {
@@ -1109,7 +1109,7 @@ namespace lfs::io {
         auto sh0 = splat_data.sh0_raw().cpu();
         const auto* sh0_ptr = sh0.ptr<float>();
 
-        auto color_result = cluster1d(sh0_ptr, num_rows, 3, options.kmeans_iterations);
+        auto color_result = cluster1d(sh0_ptr, static_cast<int>(num_rows), 3, options.kmeans_iterations);
 
         auto opacity = splat_data.opacity_raw().cpu();
         const auto* opacity_ptr = opacity.ptr<float>();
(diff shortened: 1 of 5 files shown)


MrNeRF:
Just looked at it. This pr has many problems:
- Silencing warnings with cstyle casts is not the correct way to handle this. In essence, the return values must be checked and not silenced.
- Not sure why so many variables are rename and Hungarian notation was introduced?
- Changes to the tensor lib are not accepted without running the tests

What's ok:
CMakelists ok
static_cast are also ok

MotivaCG:
> * Silencing warnings with cstyle casts is not the correct way to handle this. In essence, the return values must be checked and not silenced.

Oh I thought that they were intentionally ignored

> * Not sure why so many variables are rename and Hungarian notation was introduced?

No hungarian notation was used at all, I think that the confussion comes from this line change:
const float t = 1.0f - m_selectionFlashIntensity;
to
const float ft = 1.0f - m_selectionFlashIntensity;

it doesnt mean float t, the issue was that there was another variable named t in the higher scope and it triggers an warning (and a potential error if someone confuses them, ft is the t from the Flash.

For example the same thing happens here:

(parent_node == nullptr || other->name != parent_node->name)) {
to
(scn_parent_node == nullptr || other->name != scn_parent_node->name)) {

Anyway I've renamed it to invFlash so it will be clearer.


> * Changes to the tensor lib are not accepted without running the tests

This one shouldn't affect the performance at all, in Windows INFINITY value exceeds the Cuda float capacity that's why FLT_MAX is heavily preferred, actually the compiler can handle that limit but it could be better to be on the safe side.


What kind of test should I run? (as I said no changes should be expected at all)

MotivaCG:
updated @MrNeRF can you take a look at that? What tests should I run?

Have a great weekend.

MrNeRF:
Spend time with other stuff. Haven't looked in detail but I think it looks good. Will try to take care of all prs over the weekend. Likewise!

Issue #658, opened 2026-01-02
image input as bg for rasterizer


MrNeRF:
Can be used to create per pixel random bg color, for instance.

kmeirlaen:
would also be nice to be able to set the training background color in the training parameters, just like we can in the viewer. right now hardcoded to black.

PR #725, opened 2026-01-07
#658 Image Input for Background 

shadygm:
This introduces a image input for the background instead of just solid color. 

I would appreciate some good review on this since I'm quite unsure what I implemented makes sense or if there is a easier/smarter way to do it. 

Thanks!

#658


MrNeRF:
Will look into it either today or tomorrow! Thx a lot!

MrNeRF:
Hold on, will take later this weekend a look. Was busy with python :)

grafpez:
very interesting! would love to know if a 360 equirect can provide a 'skybox' for a cropped model, (this should also be able to be aligned), or if 'outerwordly' splats can be compressed to provide such a thing. Will have a go at this

Issue #724, opened 2026-01-07
Tile mode bug with equirectangular


gradeeterna:
Tile mode with equirectangular datasets is not working correctly. If I leave tile mode at default 1 and train on the full res equi images, it falls back to tile mode when running oom which breaks training. If I downscale the images 4x so that it doesn't run oom / use tile mode, then training looks fine.

I tested with circular fisheye images and that seemed to work fine, so I think it's specific to equirectangular rather than all GUT - https://github.com/MrNeRF/LichtFeld-Studio/commit/19b2a14bf1bc2c120172311fecb02c30cbd6686c

PR #729, opened 2026-01-07
#724 Fix: Equirectangular tile mode

gradeeterna:
Tile mode was broken for equirectangular cameras because the camera model uses image resolution directly for angle-to-pixel mapping (unlike pinhole/ fisheye which use focal length and can adjust via K matrix).

This fix:
- Passes full image resolution to equirectangular camera model
- Subtracts tile offset from projected coordinates (projection kernel)
- Adds tile offset to pixel coordinates before ray generation (rasterization)

This ensures correct angular mapping while producing tile-local coordinates.

#724


MrNeRF:
@gradeeterna I simplified the commit a bit. It should be equivalent. Please verify on master. Should be good :) Thx a lot for fixing this.

gradeeterna:
Thanks! Just tested on master and it works great :)

Issue #601, opened 2025-12-26
When using "ring selection" in combination with "depth editing", it is not possible to change the depth with ALT+scroll


kmeirlaen:
* Load a .ply
* From the toolbar, select brush editor
* press CTRL-F to active depth editing
* use alt+mouse scroll wheel to change the depth -> works ok
* from the toolbar, select "ring selection" 
* (depth mode stays active) -> ok
* use alt+mouse scroll wheel to change the depth -> does not works, instead the scroll is handled as zoom

PR #737, opened 2026-01-08
#601: Fix enable depth editing with ALT+scroll in ring selection mode.

jacobvanbeets:
The depth filter adjustment (ALT+scroll) and toggle (CTRL+F) were previously disabled when using ring selection mode. This was due to explicit checks that excluded SelectionMode::Rings from depth editing functionality.

This commit removes those restrictions, allowing users to:
- Toggle depth editing with CTRL+F in ring selection mode
- Adjust depth with ALT+scroll when depth editing is active
- Adjust width with CTRL+ALT+scroll when depth editing is active
Fix for issue: https://github.com/MrNeRF/LichtFeld-Studio/issues/601

### DIFF
--- a/src/visualizer/tools/selection_tool.cpp
+++ b/src/visualizer/tools/selection_tool.cpp
@@ -370,7 +370,7 @@ namespace lfs::vis::tools {
         const auto mode = rm ? rm->getSelectionMode() : lfs::rendering::SelectionMode::Centers;
 
         // Depth filter adjustment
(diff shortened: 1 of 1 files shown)


kmeirlaen:
Tested, works as expected.
@MrNeRF I noticed that there was some code to explicit exclude this from RING selection mode.  Was this intentional?

This is good for me to merge...

MrNeRF:
Yes, it was. I thought it was useless. No reason to exclude though.

MrNeRF:
Thx for closing another issue :muscle:

PR #772, opened 2026-01-15
add version info to console log and GUI

kmeirlaen:
When users report issues and post screenshots or log files, it is not always clear what version of LFS they are using.
This PR adds the commit hash to the GUI. We also log the version and commit has to the console upon startup

image

image


MrNeRF:
Great idea!

PR #782, opened 2026-01-18
Fix Python crash when pyd overwrites ensure_initialized callback

opsiclear-admin:
## Summary
- Fixes crash on Windows when Python bindings are enabled
- Root cause: `lfs_python_utils` statically linked into both exe and pyd creates duplicate `g_py_init_once` flags
- When pyd loads, its `g_registrar` overwrites the callback, and later calls to `ensure_initialized()` run with a fresh `call_once` flag, eventually calling `PyEval_SaveThread()` without holding the GIL

## Fix
Add early `Py_IsInitialized()` check at the start of `ensure_initialized()` to skip entirely if Python is already running.

## Test plan
- [x] Clean build from scratch with Python bindings enabled
- [x] Application starts without crash
- [x] Python console functional (`import lichtfeld as lf` works)

### DIFF
--- a/src/python/runner.cpp
+++ b/src/python/runner.cpp
@@ -244,6 +244,16 @@ _add_dll_dirs()
 
     void ensure_initialized() {
 #ifdef LFS_BUILD_PYTHON_BINDINGS
+        // Early exit if Python is already initialized by another binary.
+        // When lfs_python_utils is statically linked into both exe and pyd,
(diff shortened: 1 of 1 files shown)


opsiclear-admin:
Upstream commit 0f001c85 ('move static vars to shared lib') already fixes this issue by moving static variables to a shared library.

PR #803, opened 2026-01-24
#798 - add init_path to params upon reset

### LINKED ISSUE #798: Reset Training should not discard a specified init model
Currently if one chooses to restart a run, then any initialisation from a ply file is lost - I feel that parameter should be retained, in the same way as other settings.

kmeirlaen:
In case a --init-ply was used at the start of the training, when pressing reset after a training, the init_path was copied over to the parameters for loading the dataset. 

This PR adds the missing init_path for loading the dataset after reset.

fixes #798


MrNeRF:
Thx. Lgtm!

PR #815, opened 2026-01-28
Correct aspect ratio when ground truth comparison is toggled

miriameng:
Currently after clicking g, the ground truth image is stretched to match the current viewport, the stretching is especially bad when the images are captured vertically by a phone and when we are viewing on an ultra wide monitor. This PR resizes the window to the correct aspect ratio instead when g is clicked.

### DIFF
--- a/src/visualizer/visualizer_impl.cpp
+++ b/src/visualizer/visualizer_impl.cpp
@@ -214,45 +214,34 @@ namespace lfs::vis {
         });
 
         cmd::ResizeWindow::when([this](const auto& cmd) {
-            if (window_manager_) {
-                const int img_w = cmd.width;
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thx. I did some simpifications of the code. Can be merged after successful CI run

MrNeRF (changes requested):
Ok, after discussing we think the Window should not resize but the rendering should be adopted to the aspect ratio.

MrNeRF:
I added this letterbox approach in favor of this pr: https://github.com/MrNeRF/LichtFeld-Studio/commit/35dd9a361e019bf445adc77f120d9e3d0ae56a14

I will close it because I would have to revert everything.

PR #827, opened 2026-01-31
Selecetion tool fixes

dozeri83:
I removed updateBrushPreview in handlemousemove - this cause the mouse to be laggy when moving it, 
especially when zooming out. it fixed this issue.

I did not noticed any functionality  problems with this change


MrNeRF:
Tested, lgtm. Thank you!

PR #830, opened 2026-02-02
Fix menu bar localization by consolidating LocalizationManager singleton

jacobvanbeets:
Moved LocalizationManager from static lfs_visualizer library to shared lfs_python_runtime.dll to fix the duplicate singleton issue that caused uninitialized menu strings in Python plugins.

The Problem:
- LocalizationManager was part of static lfs_visualizer library
- Each DLL (main exe and Python bindings) got its own copy of the singleton
- Python plugins accessed an uninitialized copy, resulting in missing menu strings

The Solution:
1. Moved localization_manager.cpp to lfs_python_runtime (shared DLL)
2. Added DLL export macros (LFS_LOCALIZATION_API) for cross-DLL visibility
3. Added dependencies (nlohmann_json, spdlog, lfs_core) to python_runtime
4. Updated lfs_visualizer to link against lfs_python_runtime

This ensures one shared LocalizationManager instance across all modules.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -16,6 +16,7 @@ find_package(nanobind CONFIG REQUIRED)
 # when linked into both the main executable and the Python module
 add_library(lfs_python_runtime SHARED
     python_runtime.cpp
+    ../visualizer/gui/localization_manager.cpp
 )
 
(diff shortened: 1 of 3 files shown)


MrNeRF:
that's fixed. Closing.

PR #838, opened 2026-02-04
windows: add /bigobj to prevent C1128 number of sections exceeded object file format limit

miriameng:
Not sure why this wasn't an issue in CI, maybe it's because CI is CUDA 12.8? Locally I have CUDA 13.0, if I don't add `/bigobj` I am seeing `fatal error C1128: number of sections exceeded object file format limit`


MrNeRF:
likely nobody has build with cuda 13 on Windows since the big merge yesterday?

PR #839, opened 2026-02-04
fix cmake to consider BUILD_PYTHON_BINDINGS=OFF

miriameng:
Fix cmake problems when `BUILD_PYTHON_BINDINGS=OFF` caused by the mega merge https://github.com/MrNeRF/LichtFeld-Studio/pull/760


MrNeRF:
Hmm, this should not be off. In general, I think it should not be able to be toggled because Python is now an integrated part.

MrNeRF:
I removed this flag entirely. I close this pr.

PR #862, opened 2026-02-05
windows: add missing dll export

miriameng:
The tests are using these symbols, so they need to be exported explicitly in windows, otherwise the tests/benchmark won't link.


MrNeRF:
great! Missed these. Thank you. Let's wait for the build and merge!

PR #872, opened 2026-02-09
PPISP: Fix CRF, fix gsplat rasterization clamp

fishcu:
Minor fixes related to PPISP, including:
- Init CRF to identity. Due to the parameterization, this requires inverting the functions applied during the forward pass. Setting all params to zero is not equivalent.
- Remove the clamp to [0, 1] in the gsplat rasterizer, which blocks HDR reconstructions with PPISP
- Add clamp after bilateral grid and PPISP post-processing
- Re-use `is_known_*` helper functions

### DIFF
--- a/src/training/kernels/ppisp.cu
+++ b/src/training/kernels/ppisp.cu
@@ -332,10 +332,7 @@ namespace lfs::training::kernels {
                 case 2:  // gamma: target=1.0, min_value=0.1
                     raw_value = bounded_positive_inverse(1.0f, 0.1f);
                     break;
-                case 3:  // center: sigmoid(0) = 0.5 (identity)
-                    raw_value = 0.0f;
-                    break;
-                default:
+                default:  // case 3: center: sigmoid(0) = 0.5 (identity)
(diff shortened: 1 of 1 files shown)


src/training/kernels/ppisp.cu
```diff
@@ -308,9 +316,30 @@ namespace lfs::training::kernels {
...
+                case 1:  // shoulder: target=1.0, min_value=0.3
+                    raw_value = bounded_positive_inverse(1.0f, 0.3f);
+                    break;
+                case 2:  // gamma: target=1.0, min_value=0.1
+                    raw_value = bounded_positive_inverse(1.0f, 0.1f);
+                    break;
+                case 3:  // center: sigmoid(0) = 0.5 (identity)
+                    raw_value = 0.0f;
+                    break;
+                default:
```
MrNeRF: nitpick: Dead code! Param_idx ranges from 0...3

fishcu: Fixed

PR #876, opened 2026-02-09
Mesh support

MrNeRF:


miriameng:
Why not directly use the openmesh from vcpkg?

MrNeRF:
> Why not directly use the openmesh from vcpkg?

I wanted to have the latest version and full control over the bindings.

PR #879, opened 2026-02-10
Mesh to splat

MrNeRF:


miriameng:
Is this https://arxiv.org/pdf/2601.21400?

MrNeRF:
> Is this https://arxiv.org/pdf/2601.21400?

No, that's https://github.com/electronicarts/mesh2splat with some changes.

PR #885, opened 2026-02-11
fix: bundle venvlauncher.exe for Windows plugin venv creation

jacobvanbeets:
Python's venv module requires venvlauncher.exe to create python.exe in the Scripts folder of new virtual environments on Windows. Without these files, venvs are created but have no Python executable, causing plugin installation to fail.


MrNeRF:
I close this one. This is not the correct fix. Please see discord.

PR #909, opened 2026-02-19
Feat: Added mask modes to the evaluations.

shadygm:
Self explanatory (:

Not sure if the image saving functionality for each of the different mask modes is what is expected. I simply assumed that the user always wants to focus on the masked-in image and wants to compute metrics on that. Can be revised easily.


MrNeRF:
I cleaned the diff quite a bit up. There was duplicated code mostly and the evaluation logic with masks was not correct. 

It should be also more robust. Still needs testing before it can be merged.

PR #911, opened 2026-02-19
Pre-compute undistortion during COLMAP camera loading

### LINKED ISSUE #162: Add undistortion functionality in colmap_reader.cpp
Pre-computes undistortion maps during camera setup (similar to OpenCV's approach but without the dependency) in colmap_reader.cpp.

Updates camera intrinsics to account for the undistorted image geometry
Reference: https://github.com/nerfstudio-project/gsplat/blob/main/examples/datasets/colmap.py

ToRainu256:
## Summary

  - Pre-compute camera undistortion during COLMAP loading instead of delaying until trainer initialization.
- Cameras now always have correct pinhole intrinsics immediately after construction, matching the gsplat reference implementation.

  ## Changes

  ### `src/io/formats/colmap.cpp` 
  Added `prepare_undistortion()` call after each Camera is created in
  `assemble_colmap_cameras()`. This pre-computes undistortion maps at load time for any camera
  with distortion coefficients.

  ### `src/training/trainer.cpp` 
  Removed the redundant undistortion preparation block from trainer initialisation. This code iterated over all train/validation cameras, calling `prepare_undistortion()`, which is now handled at load time.

  Closes #162

### DIFF
--- a/src/python/lfs_plugins/installer.py
+++ b/src/python/lfs_plugins/installer.py
@@ -5,9 +5,11 @@
 import logging
 import os
 import shutil
+import stat
 import subprocess
 import sys
 import tempfile
+import time
 from pathlib import Path
 from typing import Optional, Callable, Tuple
 from urllib.parse import urlparse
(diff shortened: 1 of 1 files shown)


MrNeRF:
@ToRainu256 Thank you for this pr!
Could you please explain the reasoning behind the changes? Afaik, this will break the following behaviour:
- User can't opt out distortion.
- 3dgut can process undistorted images. This change will break it.

ToRainu256:
@MrNeRF  Thanks for the feedback! 
You’re right — the previous commit (a3e7424) was
calling prepare_undistortion() at load time, which modified intrinsics and
broke the opt-out and 3DGUT paths. That was my mistake; I hadn’t fully traced
the downstream pipeline when making that change.

I’ve updated the approach by splitting undistortion into two phases:

- precompute_undistortion(): called at load time in the COLMAP loader.
  It only computes/caches UndistortParams (the expensive CPU work) and does
  not touch intrinsics.

- prepare_undistortion(): called by the trainer only when the undistort flag
  is true. It ensures params are available (idempotent: no-op if already
  computed) and then applies them to intrinsics.

With this split, users can still opt out of undistortion — intrinsics remain
unchanged unless the trainer explicitly applies them, and 3DGUT continues to
work as before.

MrNeRF:
Thx @ToRainu256. It was  still not 100% correct. I fixed it.

So, currently it is a bit wasteful as the undistortion is unneccessary in case someone trains with gut. However, let's see how it goes.

Will merge after the pipeline succeeded. Thx :)

PR #921, opened 2026-02-23
migrate from glfw to sdl3

MrNeRF:
towards cross-platform


miriameng:
Just curious, why is GLFW considered not cross platform?

MrNeRF:
The frontend will pivot towards platform independence over the next months and hopefully run on my ipad. Glfw doesn't support ios.

PR #927, opened 2026-02-27
fix: use DXGI for per-process VRAM reporting on Windows

jacobvanbeets:
NVML returns NVML_VALUE_NOT_AVAILABLE for usedGpuMemory on Windows WDDM, causing the status bar to always show 'LFS 0.0GB'.

Replace NVML with IDXGIAdapter3::QueryVideoMemoryInfo on Windows, which reliably reports the calling process's dedicated GPU memory. NVML is retained on Linux where it works correctly.


miriameng:
For some reason I am seeing 'LFS 0.0GB' in the latest master and I am on Windows

jacobvanbeets:
Should be fixed again soon. Thanks

PR #931, opened 2026-03-02
Docker Update

shadygm:
Update docker to support already installed plugins to be loaded/


MrNeRF:
Thx! Won't work on Windows, no?

shadygm:
> Thx! Won't work on Windows, no?

i never did :p, the paths and script don't allow it to. I could convert it if there's demand, but it seems I'm the only one using it at the moment...

Issue #746, opened 2026-01-12
Rotation gizmo - add option to snap to nearest full degree when rotating


kmeirlaen:
"Add option to snap to nearest full Degree when rotating?    some models produced in different software have Swapped xyz coordinates
I know for my model setting it to 90  exactly levels the model.  Hard to accomplish with mouse
"

PR #994, opened 2026-03-16
 Add snap option to gizmos (Ctrl+drag)

ToRainu256:
Summary

  - Add grid snapping to gizmo operations when holding Ctrl during drag
  - Rotation snaps to 5° increments, translation to 0.1 units, scale to 0.1 ratio
  - Without Ctrl, gizmo behavior is unchanged

  Details

  Introduces a computeSnapPtr() helper that returns snap values based on the active gizmo operation
  when Ctrl is held, or nullptr otherwise. Applied to all three ImGuizmo::Manipulate call sites in
  GizmoManager (node gizmo, standard gizmo, and the third gizmo path).

#746

### DIFF
--- a/src/visualizer/gui/gizmo_manager.cpp
+++ b/src/visualizer/gui/gizmo_manager.cpp
@@ -56,12 +56,12 @@ namespace lfs::vis::gui {
         inline const float* computeSnapPtr(float* buf, ImGuizmo::OPERATION op) {
             if (!ImGui::GetIO().KeyCtrl)
                 return nullptr;
-            if (op == ImGuizmo::ROTATE)
+            if (op & ImGuizmo::ROTATE)
                 buf[0] = ROTATION_SNAP_DEGREES;
             else if (op & ImGuizmo::TRANSLATE)
                 buf[0] = buf[1] = buf[2] = TRANSLATE_SNAP_UNITS;
             else if (op & ImGuizmo::SCALE)
-                buf[0] = SCALE_SNAP_RATIO;
+                buf[0] = buf[1] = buf[2] = SCALE_SNAP_RATIO;
             return buf;
         }
     } // namespace


src/visualizer/gui/gizmo_manager.cpp
```diff
@@ -64,6 +52,18 @@ namespace lfs::vis::gui {
             return glm::vec3(glm::length(glm::vec3(m[0])), glm::length(glm::vec3(m[1])),
                              glm::length(glm::vec3(m[2])));
         }
+
+        inline const float* computeSnapPtr(float* buf, ImGuizmo::OPERATION op) {
+            if (!ImGui::GetIO().KeyCtrl)
+                return nullptr;
+            if (op == ImGuizmo::ROTATE)
```
MrNeRF: should be & instead == to match individual axes.

src/visualizer/gui/gizmo_manager.cpp
```diff
@@ -64,6 +52,18 @@ namespace lfs::vis::gui {
...
+                buf[0] = ROTATION_SNAP_DEGREES;
+            else if (op & ImGuizmo::TRANSLATE)
+                buf[0] = buf[1] = buf[2] = TRANSLATE_SNAP_UNITS;
+            else if (op & ImGuizmo::SCALE)
+                buf[0] = SCALE_SNAP_RATIO;
```
MrNeRF: buf[0] = buf[1] = buf[2] = SCALE_SNAP_RATIO; //  to snap all scale axes, no?

src/visualizer/gui/gizmo_manager.cpp
ToRainu256: Fixed. Now using bitmask comparison consistent with TRANSLATE/SCALE. 36d3ad2c

src/visualizer/gui/gizmo_manager.cpp
ToRainu256: Fixed. All three axes now receive the snap value.  36d3ad2c

ToRainu256:
> Thanks for this pr. I have added two comments which make it a bit more general. Could be also 10 degrees? Wdyt?

 Thanks for the review! I hardcoded 5° to match Blender's default. Ideally this would be
  user-configurable in the UI, but that's a bigger change so I kept it simple for now. Happy to change
   it to 10° if you prefer.

MrNeRF:
> > Thanks for this pr. I have added two comments which make it a bit more general. Could be also 10 degrees? Wdyt?
> 
> Thanks for the review! I hardcoded 5° to match Blender's default. Ideally this would be user-configurable in the UI, but that's a bigger change so I kept it simple for now. Happy to change it to 10° if you prefer.

Thank you. Let's start with 5 degrees and let's see how it goes. Configuration can be added  later, too! 
Lgtm!

PR #1033, opened 2026-03-25
Syncing settings name between --help and config

ichsan2895:
`LichtFeld-Studio -h` result:
...
RENDERING OPTIONS:
        --enable-mip                      Enable mip filter (anti-aliasing)
        --bilateral-grid                  Enable bilateral grid filtering
        --ppisp                           Enable PPISP for per-camera appearance modeling
        --ppisp-controller                Enable PPISP controller for novel views
        --bg-modulation                   Enable sinusoidal background modulation
        --gut                             Enable GUT mode
...

While in the config file, it said `use_bilateral_grid`

I think we can make it same name for consistency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,7 @@ find_package(WebP REQUIRED)
 find_package(LibArchive REQUIRED)
 find_package(OpenImageIO REQUIRED)
 find_package(OpenSSL REQUIRED)
+find_package(pxr CONFIG REQUIRED)
 find_package(assimp REQUIRED)
 find_package(RmlUi CONFIG REQUIRED)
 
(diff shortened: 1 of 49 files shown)


MrNeRF:
This will break many things. Look here
https://github.com/MrNeRF/LichtFeld-Studio/blob/master/src/core/parameters.cpp#L344-L346

MrNeRF:
You changed the eval configs that's why it does not break for you when you run your training. There are many places including the python bindings that need changes to not break anything. The intent is great but it breaks LFS .

ichsan2895:
> The intent is great but it breaks LFS

Should I close this PR? I don't aware the solution is not easy as my thought.

MrNeRF:
Yeah, sorry I close. Please feel free to create a new pr if you really want to dig deep in.

PR #1048, opened 2026-03-28
Add --output-name parameter to override default PLY filename

MotivaCG:
Adds --output-name CLI flag to DatasetConfig, replacing the default splat_<iter>.ply stem on final saves (training complete, stop request, save_final_ply_and_checkpoint). 

Without this parameter, the existing default naming behaviour (splat_<iter>.ply) is preserved.

Also serializes output_name in DatasetConfig JSON for use with --config files.


src/core/include/core/parameters.hpp
```diff
@@ -260,6 +261,7 @@ namespace lfs::core {
         struct LFS_CORE_API ConvertParameters {
             std::filesystem::path input_path;
             std::filesystem::path output_path; // Empty = derive from input
+            std::filesystem::path output_name; // Empty = splat_ + number of steps
```
MrNeRF: why that's here? Dead code.

MotivaCG: You are right. I planned to support that in the converter as well, but I thought it didn't make sense and reverted all the changes except this one!

PR #1050, opened 2026-03-28
MCP: Add full UI window capture, harden viewport capture

SharkWipf:
I needed to debug the UI, and realized there was no MCP path for capturing the full window with all elements yet.
This adds that.

Disclaimer: While the machine assured me it would work on Windows, I have not confirmed this.
Disclaimer 2: While I do not see anything in the code that would cause issues on real hardware, this was written and tested on a highly unstable and experimental ZLUDA build, *not* on real CUDA hardware.

I will not take offense to any ZLUDA-originated PRs not being merged :wink:

### DIFF
--- a/src/app/include/app/mcp_app_utils.hpp
+++ b/src/app/include/app/mcp_app_utils.hpp
@@ -42,53 +42,62 @@ namespace lfs::app {
         }
     }
 
+    namespace detail {
+
+        template <typename F, typename PostFn>
+        auto post_and_wait_impl(PostFn&& post_fn, F&& fn) {
+            using R = std::invoke_result_t<F>;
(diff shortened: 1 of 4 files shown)


SharkWipf:
Added a general MCP image path bugfix. MCP call responses were always being wrapped in json, but for images to viewed by the agent directly without storing them to a temp file on disk first, they need to be sent without the json wrapper.
So this should allow more direct visual MCP use for all visual paths.

Todo: Chrome MCP accepts an optional file path for any image paths. It will inline the image if omitted, or store it to the given path if not. Currently LFS only inlines the image.

MrNeRF:
I cleaned it a bit up. Works! Thx!

PR #1057, opened 2026-03-30
Fix modal click-through on startup overlay

jacobvanbeets:
## Summary
- stop modal clicks from reaching the welcome-screen UI behind the default-application prompt
- mask underlay input while a blocking RML modal or context menu is open
- keep the active modal on the original input stream so its buttons still work normally

## Validation
- git --no-pager diff --check -- src/visualizer/gui/gui_manager.cpp


MrNeRF:
That's all good! Thank you!

PR #1062, opened 2026-03-30
Fixed the build‑failure issue on environments that do not support AVX2

todashuta:
Because the `COMPILE_OPTIONS` added to `splat_simplify.cpp` included `-mno-fma` and `-mno-avx2`, the following error occurred during compilation on systems that do not support AVX2:

```
g++-14: error: unrecognized command-line option ‘-mno-fma’
g++-14: error: unrecognized command-line option ‘-mno-avx2’
```

This PR resolves the problem described above.


MrNeRF:
Great! Thanks!

PR #1074, opened 2026-04-02
Vignette Fix

shadygm:
Allow user to customize the vignette.

ON: 
image

Off: 
image


MrNeRF:
Do you add changed shaders? And I think it should be opt-out because otherwise it can be completely removed.

shadygm:
@MrNeRF ready! vignette is opt-out

PR #1078, opened 2026-04-05
fix shutdown crash on exit

Lyapsus:
## Summary
- finalize Python before destroying the viewer/GL context
- avoid shutdown crashes when Python-owned RmlUI panels tear down OpenGL resources on exit

## Testing
- reproduced on Linux with `timeout --signal=INT 15s LichtFeld-Studio`
- before: exit `139`
- after: clean shutdown, no segfault

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -303,12 +303,13 @@ namespace lfs::app {
 
             mcp_http.stop();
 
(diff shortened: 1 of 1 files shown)


MrNeRF:
That's justified. Thanks!

PR #1080, opened 2026-04-06
fix modal overlay shutdown crash

Lyapsus:
## Summary
- destroy the modal overlay before RmlUI shutdown starts
- avoid shutdown crashes when the modal overlay tears down OpenGL resources after window close

## Testing
- reproduced by closing the window on Linux
- before: segfault in `RmlFBO::destroy()` via `RmlModalOverlay::~RmlModalOverlay()`
- after: app exits normally, no segfault


MrNeRF:
Interestingly, I don't experience any crashes but the changes make sense!

PR #1082, opened 2026-04-07
Add action-only plugin toolbar support

OlstFlow:
Adds minimal plugin toolbar API support needed for true one-click action plugins.

This patch adds:
- ToolDef.action_only
- ToolDef.selected
- plugin toolbar icon loading from plugin_path/icons/{icon}.png

This allows sync-style plugins to behave as one-shot toolbar actions instead of persistent tools, and enables plugin-defined busy highlight state and custom toolbar icons.

### DIFF
--- a/src/python/lfs_plugins/toolbar.py
+++ b/src/python/lfs_plugins/toolbar.py
@@ -2,11 +2,15 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 """Viewport toolbars rendered from a retained RmlUI data model."""
 
+from pathlib import Path
+from urllib.parse import quote
+
 from .histogram_support import histogram_mode_available
 from .tools import ToolRegistry
 
 
 _TOOLBAR_HIDDEN_STATES = ("running", "paused", "stopping", "completed")
+_RML_PATH_SAFE_CHARS = "/:._-~"
 
 _toolbar_controller = None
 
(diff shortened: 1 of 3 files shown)


src/python/lfs_plugins/toolbar.py
```diff
@@ -13,11 +14,29 @@
...
     return f"../icon/{icon_name}.png"
 
 
+def _tool_icon_src(tool_def):
+    """Resolve builtin toolbar icons or plugin-provided toolbar icons."""
+    plugin_path = getattr(tool_def, "plugin_path", "") or ""
+    if plugin_path:
+        candidate = Path(plugin_path) / "icons" / f"{tool_def.icon}.png"
+        if candidate.exists():
+            return candidate.as_posix()
```
MrNeRF: from urllib.parse import quote
    return quote(candidate.as_posix(), safe=_RML_PATH_SAFE_CHARS)

instead of:
            return candidate.as_posix()

OlstFlow:
Synced the branch with current `master` and updated the toolbar icon path handling to use `quote(candidate.as_posix(), safe=_RML_PATH_SAFE_CHARS)` for plugin icon paths. The PR branch has been force-updated accordingly.

src/python/lfs_plugins/toolbar.py
OlstFlow: Done, thanks for the suggestion. I synced with current master and updated the icon path handling to use quote(..., safe=_RML_PATH_SAFE_CHARS).

MrNeRF:
Ok, lgtm! Thank you!

PR #1100, opened 2026-04-11
add missing font license file and third party license links

todashuta:
I added the missing license files for several fonts. I also added information regarding third-party license files.


MrNeRF:
Great! Thank you!

PR #1103, opened 2026-04-12
Add Viewport Export plugin to marketplace curated list

jacobvanbeets:
This plugin allows users to export the current viewport as a JPG image at 1080p, 4K, or 8K resolution directly from the Plugin Manager.


jacobvanbeets:
Plugin to go to web instead.

Issue #1083, opened 2026-04-07
Export USD Crashing 


bgofish:
Exporting USD from V0.5.1 (Windows 11) - Progress Bar gets to 10% and then 'full crash' 

I've even started LFS from Command line to keep the console open & nothing gets reported.

MrNeRF:
Anyone can reproduce and provide extra info?

bgofish:
Crash Log from Windows: I added the .txt to get it to upload here [Report.wer.TXT](https://github.com/user-attachments/files/26551405/Report.wer.TXT)

xiaoccc4980:
Unable to export .USD; the program crashes when the import progress reaches 10%.

jacobvanbeets:
Should be fixed https://github.com/MrNeRF/LichtFeld-Studio/issues/1083

This bug was really annoying to find ☻

MrNeRF:
I close! SHould be fixed!

bgofish:
Latest NB  v0.5.2.dev20260420+29cedacd   USDZ is working :thumbsup:  & Imports back into LFS for a round trip  - but USD is still crashing the program out.

bgofish:
Confirm all working in the 24/04 build - it still  sits on 10%  for a while but then successfully completes.

MrNeRF:
this 10% progress indication should be fixed, too :)

PR #1112, opened 2026-04-15
#1083 Fix: USD export crash on Windows due to missing plugin resources (this bug was annoying!)

jacobvanbeets:
OpenUSD requires plugInfo.json and generatedSchema.usda files at runtime for plugin/schema discovery.  These were never deployed from vcpkg to the build output or install prefix, causing UsdStage::CreateNew() to hit TF_FATAL_ERROR -> abort() -> instant crash with no error message.

Changes:

- CMakeLists.txt: deploy USD plugin resources from vcpkg to <exe_dir>/../lib/usd/ via post-build commands and install() rules

- main.cpp: register USD plugins via PlugRegistry::RegisterPlugins() at the start of main(), before any USD API usage.  The env-var-only approach (PXR_PLUGINPATH_NAME) is unreliable on Windows because USD DLLs may initialise static data before main() runs.

- usd.cpp: add TfErrorMark to capture OpenUSD diagnostics and try/catch around each USD API call so failures produce error messages instead of crashes

- async_task_manager.cpp: wrap the export thread body in try/catch to prevent any uncaught exception from terminating the process

This should close issue: https://github.com/MrNeRF/LichtFeld-Studio/issues/1083

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -655,6 +655,45 @@ add_custom_command(TARGET ${PROJECT_NAME} POST_BUILD
     COMMENT "Copying locale files"
 )
 
+# =============================================================================
+# OpenUSD PLUGIN RESOURCES - Required for USD import/export at runtime.
+# OpenUSD discovers its schema and file-format plugins through plugInfo.json
+# files.  vcpkg places per-module resources under bin/usd/ and a root index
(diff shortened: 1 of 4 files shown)


MrNeRF:
Look good to me! I have a follow up cleanup that I will apply after the merge! Thank you!

PR #1128, opened 2026-04-21
Fix x264 build and GNOME window decorations, linux

selimsandal:
Ubuntu 25.10 x86_64
GNOME

### DIFF
--- a/.github/overlays/x264/allow-clang-cl.patch
+++ b/.github/overlays/x264/allow-clang-cl.patch
@@ -1,0 +1,15 @@
+diff --git a/configure b/configure
+index 6f95e2314..e677e36f4 100644
+--- a/configure
++++ b/configure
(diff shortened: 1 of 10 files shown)


MrNeRF:
Thanks for contributing.

We had some discussions here: https://github.com/MrNeRF/LichtFeld-Studio/pull/1070
I will review tomorrow

selimsandal:
> Thanks for contributing.
> 
> We had some discussions here: #1070 I will review tomorrow

Thanks! Will take a look at that discussion.

MrNeRF:
Please split this pr. The window-manager/docs change and the x264 overlay are independent. The former is fine.  

I think we should not introduce custom overlays. This issue should be fixed upstream as it is not an LFS issue.

selimsandal:
Closing in favor of #1136 and #1137

PR #1129, opened 2026-04-21
Fix potential missing-source failure of libvterm during a Windows build

kersplody:
This patch makes the Windows build reliably initialize and validate the external/libvterm submodule before CMake tries to compile it, replacing the missing-source failure with a clear fix path.


kmeirlaen:
Seems to also change the eval function?

kersplody:
my error, need to resubmit on a branch

PR #1134, opened 2026-04-22
fix usd export

### LINKED ISSUE #1083: Export USD Crashing

kmeirlaen:
fixes https://github.com/MrNeRF/LichtFeld-Studio/issues/1083

## Symptom

Clicking export to USD crashed the application with a fatal USD error:

```
usd_plug.dll!PlugRegistry::DemandPluginForType() — TF_FATAL_ERROR
usd_ar.dll!_GetAvailableResolvers()
usd_ar.dll!_DispatchingResolver::_DispatchingResolver()
usd_ar.dll!ArGetResolver()
usd_sdf.dll!SdfLayer::_CreateNew()
usd_usd.dll!UsdStage::CreateNew()
lfs_visualizer.dll!lfs::io::save_usd()
```

`TF_FATAL_ERROR` calls `ArchDebuggerTrap` → `abort()`. The error could not be caught by
`try/catch`.

## Root Cause

Two independent bugs in the OpenUSD plugin discovery setup:

### Bug 1 — Wrong plugin source directory

`CMakeLists.txt` searched for USD plugins in this order:

```cmake
"${CMAKE_BINARY_DIR}/vcpkg_installed/${VCPKG_TARGET_TRIPLET}/lib/usd"   # checked first
"${CMAKE_BINARY_DIR}/vcpkg_installed/${VCPKG_TARGET_TRIPLET}/bin/usd"   # actual plugins
```

The existence check was just `if(EXISTS "${_candidate}")`.  
`lib/usd/` **exists** but contains only a root `plugInfo.json` with `"Includes": ["*/resources/"]`
and **no actual plugin subdirectories**.  
`bin/usd/` is where the real plugin tree lives (`ar/`, `sdf/`, `usd/`, etc.).

Result: the POST_BUILD copy only placed the empty root index at the destination. USD's
`RegisterPlugins()` found zero actual plugins.

### Bug 2 — Wrong plugin destination (LibraryPath mismatch)

Even if `bin/usd/` had been chosen as the source, the destination was:

```cmake
set(_USD_PLUGIN_DEST "$<TARGET_FILE_DIR:${PROJECT_NAME}>/../lib/usd")
```

Each plugin descriptor contains a relative `LibraryPath`, e.g.:

```json
{ "LibraryPath": "../../usd_ar.dll", "Root": ".." }
```

USD resolves `LibraryPath` relative to the plugin root directory. With the destination
`build/Release/../lib/usd/`:

| Path | Resolves to | DLL actually at |
|---|---|---|
| `build/lib/usd/ar/` (root) | `build/lib/usd/ar/../../usd_ar.dll` = `build/lib/usd_ar.dll` | `build/Release/usd_ar.dll` |

`LoadLibraryW` with a full absolute path that doesn't exist fails immediately with no fallback.

The correct destination is `build/Release/usd/`, where the LibraryPaths resolve to
`build/Release/usd/ar/../../usd_ar.dll` = **`build/Release/usd_ar.dll`** ✓ — the same DLL
copies that are already loaded by Windows at process startup.

## Why It Worked on Some Machines

If USD plugins are registered from their **original vcpkg location** (`bin/usd/ar/`), the
`LibraryPath` resolves to `bin/usd_ar.dll` which exists in vcpkg. But loading that DLL a second
time (it was already loaded from `build/Release/`) creates **two separate module instances**,
breaking USD's TfType singleton system. This path was never hit in practice; the crash always
occurred due to zero plugins being registered.

### `CMakeLists.txt`

1. **Plugin source directory** — added sentinel check (`ar/resources/plugInfo.json`) to ensure
   only directories with actual plugin subdirectories are selected:

   ```cmake
   if(NOT _USD_PLUGIN_SOURCE_DIR AND EXISTS "${_candidate}/ar/resources/plugInfo.json")
   ```

2. **Plugin destination** — changed to `<exe_dir>/usd/` on Windows so that relative
   `LibraryPath` entries resolve to the DLL copies already loaded from `<exe_dir>/`:

   ```cmake
   if(WIN32)
       set(_USD_PLUGIN_DEST "$<TARGET_FILE_DIR:${PROJECT_NAME}>/usd")
   else()
       set(_USD_PLUGIN_DEST "$<TARGET_FILE_DIR:${PROJECT_NAME}>/../lib/usd")
   endif()
   ```

### `src/app/main.cpp` — `configure_usd_plugins()`

Changed the runtime lookup path to match the new destination:

```cpp
#ifdef _WIN32
    auto usd_dir = exe_dir / "usd";          // was: exe_dir / ".." / "lib" / "usd"
#else
    auto usd_dir = exe_dir / ".." / "lib" / "usd";
#endif
```

Added `std::println(stderr, ...)` diagnostics (logger not yet initialised at this point) that
show the plugin directory path and the total registered plugin count.

### `src/io/formats/usd.cpp` — `save_usd()`

- Added `#include <pxr/base/plug/registry.h>`.
- Pre-flight check: queries `PlugRegistry::GetAllPlugins().size()` before `UsdStage::CreateNew`.
  If the count is zero the function returns a proper `Result` error instead of calling
  `abort()`, giving a recoverable error message rather than a crash.
- Added `LOG_DEBUG("[USD] ...")` step logging through each stage of the save pipeline
  (stage creation, prim definition, metadata, attribute authoring, layer save).

## Post-Fix Log Output

Normal successful export (debug level, visible with `--log-level debug`):

```
[USD] registering plugins from: C:\...\build\Release\usd    (stderr, at startup)
[USD] plugin registration complete (47 plugins)             (stderr, at startup)
...
[info]  Saving USD file: C:\...\export.usd
[debug] [USD] registered plugin count: 47
[debug] [USD] creating stage
[debug] [USD] stage created — defining prim
[debug] [USD] prim defined — setting stage metadata
[debug] [USD] metadata set — authoring 1234567 gaussian attributes
[debug] [USD] attributes authored — saving layer
[debug] [USD] export complete: C:\...\export.usd
```

If plugins are missing (zero count), the error is now reported gracefully:

```
[error] [USD] No USD plugins registered — UsdStage::CreateNew will crash fatally
```

## Additional changes

Two bugs were fixed in the `cmake --install` / `BUILD_PORTABLE` workflow on Windows.

### Symptom

```
[error] [USD] No USD plugins registered — UsdStage::CreateNew will crash fatally
```

Occurred when running `LichtFeld-Studio.exe` from a portable install (`cmake --install build --prefix ./dist`).

### Root cause

The CMake install rule placed USD plugin JSON files into `${CMAKE_INSTALL_LIBDIR}/usd` = `dist/lib/usd`.

At runtime, `configure_usd_plugins()` on Windows looks for `<exe_dir>/usd`. When the exe lives at `dist/bin/LichtFeld-Studio.exe`, it searches `dist/bin/usd` — which did not exist.

The plugin `LibraryPath` entries (e.g. `"../../usd_ar.dll"`) are also layout-sensitive: they are relative to `<plugin>/resources/`, so the entire `usd/` subtree must sit two directories above the actual USD DLLs. Placing it next to the exe (`bin/usd/`) means `bin/usd/ar/resources/../../usd_ar.dll` resolves to `bin/usd_ar.dll` ✓.

### Fix

Made the USD install destination platform-conditional in `CMakeLists.txt`:

```cmake
if(WIN32)
    set(_USD_INSTALL_DEST "${CMAKE_INSTALL_BINDIR}/usd")   # dist/bin/usd
else()
    set(_USD_INSTALL_DEST "${CMAKE_INSTALL_LIBDIR}/usd")   # dist/lib/usd (Linux unchanged)
endif()

install(DIRECTORY "${_USD_PLUGIN_SOURCE_DIR}/" DESTINATION "${_USD_INSTALL_DEST}" ...)
install(FILES    "${_USD_ROOT_INDEX}"          DESTINATION "${_USD_INSTALL_DEST}")
```

---

### Symptom

After `cmake --install`, all USD DLLs (and other vcpkg DLLs) appeared in both `dist/lib/` and `dist/bin/`, roughly doubling the size of the distribution (~100 MB wasted).

### Root cause

The `bundle_runtime_deps()` helper was designed with a Unix-first layout: it staged all runtime DLLs into `${CMAKE_INSTALL_LIBDIR}` (`dist/lib/`). A subsequent Windows-only step then copied every `dist/lib/*.dll` to `dist/bin/`. This left the originals in `dist/lib/` as dead copies — Windows only searches the directory containing the exe (`dist/bin/`) for DLLs.

### Fix

1. `bundle_runtime_deps()` now writes directly to `${CMAKE_INSTALL_BINDIR}` on Windows:

```cmake
if(WIN32)
    set(_BUNDLE_DEST "${CMAKE_INSTALL_BINDIR}")
else()
    set(_BUNDLE_DEST "${CMAKE_INSTALL_LIBDIR}")
endif()
```

2. Removed the redundant `file(GLOB _dlls ... *.dll)` → `file(INSTALL ... DESTINATION bin/)` step from the Windows post-install block.

After this fix, `dist/lib/` contains only import libs (`.lib` files), and all DLLs are exclusively in `dist/bin/`.

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -54,13 +54,12 @@ namespace {
 #endif
         usd_dir = std::filesystem::canonical(usd_dir, ec);
         if (ec || !std::filesystem::is_directory(usd_dir, ec)) {
-            std::println(stderr, "[USD] plugin directory not found ({})",
-                         ec ? ec.message() : "not a directory");
+            LOG_ERROR("[USD] plugin directory not found ({})",
+                      ec ? ec.message() : "not a directory");
             return;
         }
 
         const std::string path_utf8 = lfs::core::path_to_utf8(usd_dir);
-        std::println(stderr, "[USD] registering plugins from: {}", path_utf8);
 
         // Also set the env var for any code that reads it directly.
 #ifdef _WIN32
@@ -72,8 +71,6 @@ namespace {
         // Programmatically register plugins so the Plug system finds them
         // regardless of compiled-in search paths or env var timing.
         pxr::PlugRegistry::GetInstance().RegisterPlugins(path_utf8);
-        std::println(stderr, "[USD] plugin registration complete ({} plugins)",
-                     pxr::PlugRegistry::GetInstance().GetAllPlugins().size());
     }
 } // namespace
 
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -27,7 +27,9 @@
 #endif
 
 namespace {
-    // Register OpenUSD plugin resources deployed at <exe_dir>/../lib/usd/.
+    // Register OpenUSD plugin resources deployed beside the executable.
+    // On Windows: <exe_dir>/usd/ — keeps relative LibraryPaths correct.
+    // On Linux:   <exe_dir>/../lib/usd/ — conventional layout.
     // Must be called before any USD API usage (stage creation, schema lookup).
     // The env-var approach (PXR_PLUGINPATH_NAME) does not work reliably on
     // Windows because USD DLLs may initialise before main() runs.
@@ -39,13 +41,26 @@ namespace {
             return;
         }
 
-        auto usd_dir = exe_dir / ".." / "lib" / "usd";
         std::error_code ec;
+
+        // On Windows, plugins sit next to the exe at <exe_dir>/usd/ so that
+        // relative LibraryPath entries (e.g. "../../usd_ar.dll") resolve to
+        // the DLL copies that are already loaded by Windows at startup.
+        // On Linux they follow the conventional <exe_dir>/../lib/usd/ layout.
+#ifdef _WIN32
+        auto usd_dir = exe_dir / "usd";
+#else
+        auto usd_dir = exe_dir / ".." / "lib" / "usd";
+#endif
         usd_dir = std::filesystem::canonical(usd_dir, ec);
-        if (ec || !std::filesystem::is_directory(usd_dir, ec))
+        if (ec || !std::filesystem::is_directory(usd_dir, ec)) {
+            std::println(stderr, "[USD] plugin directory not found ({})",
+                         ec ? ec.message() : "not a directory");
             return;
+        }
 
         const std::string path_utf8 = lfs::core::path_to_utf8(usd_dir);
+        std::println(stderr, "[USD] registering plugins from: {}", path_utf8);
 
         // Also set the env var for any code that reads it directly.
 #ifdef _WIN32
@@ -57,6 +72,8 @@ namespace {
         // Programmatically register plugins so the Plug system finds them
         // regardless of compiled-in search paths or env var timing.
         pxr::PlugRegistry::GetInstance().RegisterPlugins(path_utf8);
+        std::println(stderr, "[USD] plugin registration complete ({} plugins)",
+                     pxr::PlugRegistry::GetInstance().GetAllPlugins().size());
     }
 } // namespace
 
--- a/src/io/formats/usd.cpp
+++ b/src/io/formats/usd.cpp
@@ -757,7 +757,6 @@ namespace lfs::io {
         // Pre-flight check: ArResolver requires USD plugins to be registered.
         // If none are registered, the USD plugin path may not be configured.
         const auto plugin_count = pxr::PlugRegistry::GetInstance().GetAllPlugins().size();
-        LOG_DEBUG("[USD] registered plugin count: {}", plugin_count);
         if (plugin_count == 0) {
             LOG_ERROR("[USD] No USD plugins registered — UsdStage::CreateNew will crash fatally");
             return make_error(ErrorCode::WRITE_FAILURE,
@@ -792,7 +791,6 @@ namespace lfs::io {
         // TfErrorMark captures OpenUSD diagnostic errors that don't throw.
         pxr::TfErrorMark error_mark;
 
-        LOG_DEBUG("[USD] creating stage");
         pxr::UsdStageRefPtr stage;
         try {
             stage = pxr::UsdStage::CreateNew(lfs::core::path_to_utf8(options.output_path));
@@ -814,7 +812,6 @@ namespace lfs::io {
                               options.output_path);
         }
 
-        LOG_DEBUG("[USD] stage created — defining prim");
         const pxr::SdfPath prim_path("/GaussianSplats");
         pxr::UsdVolParticleField3DGaussianSplat splat_prim;
         try {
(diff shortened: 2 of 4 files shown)


src/io/formats/usd.cpp
```diff
@@ -24,6 +24,7 @@
 #include <pxr/base/gf/quath.h>
 #include <pxr/base/gf/vec3f.h>
 #include <pxr/base/gf/vec3h.h>
+#include <pxr/base/plug/registry.h>
```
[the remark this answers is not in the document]
kmeirlaen: this has been handled

src/app/main.cpp
```diff
@@ -39,13 +41,26 @@ namespace {
...
+        // On Linux they follow the conventional <exe_dir>/../lib/usd/ layout.
+#ifdef _WIN32
+        auto usd_dir = exe_dir / "usd";
+#else
+        auto usd_dir = exe_dir / ".." / "lib" / "usd";
+#endif
         usd_dir = std::filesystem::canonical(usd_dir, ec);
-        if (ec || !std::filesystem::is_directory(usd_dir, ec))
+        if (ec || !std::filesystem::is_directory(usd_dir, ec)) {
+            std::println(stderr, "[USD] plugin directory not found ({})",
```
MrNeRF: use the logger please

src/app/main.cpp
```diff
@@ -39,13 +41,26 @@ namespace {
...
+                         ec ? ec.message() : "not a directory");
             return;
+        }
 
         const std::string path_utf8 = lfs::core::path_to_utf8(usd_dir);
+        std::println(stderr, "[USD] registering plugins from: {}", path_utf8);
```
MrNeRF: logger

src/app/main.cpp
```diff
@@ -57,6 +72,8 @@ namespace {
         // Programmatically register plugins so the Plug system finds them
         // regardless of compiled-in search paths or env var timing.
         pxr::PlugRegistry::GetInstance().RegisterPlugins(path_utf8);
+        std::println(stderr, "[USD] plugin registration complete ({} plugins)",
```
MrNeRF: logger

src/io/formats/usd.cpp
```diff
@@ -772,6 +792,7 @@ namespace lfs::io {
         // TfErrorMark captures OpenUSD diagnostic errors that don't throw.
         pxr::TfErrorMark error_mark;
 
+        LOG_DEBUG("[USD] creating stage");
```
MrNeRF: do we still need the debug logs?

src/python/CMakeLists.txt
```diff
@@ -486,6 +486,7 @@ add_custom_target(lfs_plugins_copy ALL
 )
 
 add_custom_target(lfs_python_support_copy ALL
+    COMMAND ${CMAKE_COMMAND} -E make_directory $<TARGET_FILE_DIR:lfs_py>
```
MrNeRF: that's from the last pr?

MrNeRF:
Thx!

PR #1137, opened 2026-04-22
Split/x264 overlay workaround

selimsandal:


MrNeRF:
See my comment https://github.com/MrNeRF/LichtFeld-Studio/pull/1128#issuecomment-4296297455

shadygm:
@selimsandal, can you check the previous comment. Thanks (:

selimsandal:
@MrNeRF @shadygm i’ve checked the upstream for x264, seems like they host on a private git server and the account I tried to register with 2 weeks ago is still not verified by the administrators of videolan so I cannot send a patch to that repo. If you are in contact with videolan maybe we can ask directly?

PR #1150, opened 2026-04-25
automatic centralization

dozeri83:
This PR implements automatic centralization feature.
Uppon opening a dataset - the data set import pop will have a new "centralize dataset" drop down options

None, Auto (default), By Points, by Cameras.

None - will do nothing (mainly for headless)
Auto - will automatically decide if to centralize dataset.
by Points - will calculate the geometric median of the points and centralize it to the origin
By cameras - will calculate the geo median of the cameras and centralize it to the origin.

Auto decision will be done in the following way: 
  calculate the geometric median, geometric M.A.D ( median absolute differences) and will centralized iff  |origin -median| > 2mad
 
 
 I tested feature on most of the "usual" datasets and on most of them auto mode did not centralized.
 on aerial datasets - the centralization result was as I expected.
 
 
98% of the code was written by claude code agent.

image

image


MrNeRF:
I will check later today or tomorrow. 
Don't know what it is initially doing. I would suggest it is per default disabled and can be toggled by the user to not change the current behavior.

dozeri83:
> I will check later today or tomorrow. Don't know what it is initially doing. I would suggest it is per default disabled and can be toggled by the user to not change the current behavior.

I will change default to None  but can you reconsider? i think that making the decision automatically can make a good user experience. It will only centralize in extreme cases ( i can increase the m.a.d  threshold) . Unfortunately most users don’t tend to change default.

MrNeRF:
> > I will check later today or tomorrow. Don't know what it is initially doing. I would suggest it is per default disabled and can be toggled by the user to not change the current behavior.
> 
> I will change default to None but can you reconsider? i think that making the decision automatically can make a good user experience. It will only centralize in extreme cases ( i can increase the m.a.d threshold) . Unfortunately most users don’t tend to change default.

Yes but many rely now on LFS in their workflow and you don't know how they will be impacted. I wouldn't  want to impose it. I can think of a preference menu to configure such things. Once that's available this could be changed.

dozeri83:
@MrNeRF  - i made None default (and change it to "OFF").   
also i removed "Auto"  options from drop down menu since it is kind a pointless.

kmeirlaen:
i see this more useful as a "one click" action after importing. personally, i align the dataset before importing in LFS and I would not want LFS to automatically change the point cloud position as it could be aligned towards 3rd party data

dozeri83:
> i see this more useful as a "one click" action after importing. personally, i align the dataset before importing in LFS and I would not want LFS to automatically change the point cloud position as it could be aligned towards 3rd party data

I normalize that data too before processing but I have to use external python script for it (so i see this as a more power user thing). I see this is a “quality of life” feature for beginners.
Regarding your one click suggestion - i thought about it but the design was unnatural since we load the dataset from two different modules (the ui and the training). Also i had no idea where to put the centralize button in the gui… 😅

src/io/include/io/loader.hpp
```diff
@@ -44,12 +44,21 @@ namespace lfs::io {
         Transforms
     };
 
+    // Centralize dataset enum
+    enum class CentralizeDataset {
+        Off,
+        Auto,
```
MrNeRF: I think auto is gone, no?

src/geometry/CMakeLists.txt
```diff
@@ -32,4 +33,4 @@ message(STATUS "╔════════════════════
 message(STATUS "║  geometry - LibTorch-free Geometry Module                 ║")
 message(STATUS "╚════════════════════════════════════════════════════════════════╝")
 message(STATUS "  • Namespace: lfs::geometry")
-message(STATUS "  • Contains: EuclideanTransform, BoundingBox")
+message(STATUS "  • Contains: EuclideanTransform, BoundingBox, geometry_misc")
```
MrNeRF: Does this bring any value?

src/core/include/core/events.hpp
```diff
@@ -57,7 +57,7 @@ namespace lfs::core {
             EVENT(ResetTraining, );
             EVENT(SwitchToLatestCheckpoint, );
             EVENT(SaveCheckpoint, std::optional<int> iteration;);
-            EVENT(LoadFile, std::filesystem::path path; bool is_dataset; std::filesystem::path output_path; std::filesystem::path init_path;);
+            EVENT(LoadFile, std::filesystem::path path; bool is_dataset; std::filesystem::path output_path; std::filesystem::path init_path; std::string centralize_dataset;);
```
MrNeRF, line 60: Why not use the enum here?

src/io/loaders/loader_utils.hpp
```diff
@@ -45,4 +49,104 @@ namespace lfs::io {
...
+            pts.reserve(cameras.size());
+            for (const auto& cam : cameras) {
+                auto pos = cam->cam_position().cpu();
+                const float* p = pos.ptr<float>();
+                pts.push_back({p[0], p[1], p[2]});
+            }
+            return pts;
+        };
+
+        if (mode == CentralizeDataset::Auto && point_cloud && point_cloud->size() > 0) {
```
MrNeRF: why not go with switch case?

src/geometry/geometry_misc.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "geometry/geometry_misc.hpp"
+#include <algorithm>
+#include <glm/geometric.hpp>
+#include <numeric>
+#include <random>
+
+namespace lfs {
+    namespace geometry {
+
+        glm::vec3 geometric_median(std::span<const glm::vec3> points,
```
MrNeRF, line 14: These functions are used in a single place. I would argue to move them where they are used without adding them here.

src/io/include/io/loader.hpp
MrNeRF: now auto is dead, no? But it is sprinkled all over in the code

src/visualizer/gui/rmlui/resources/dataset_import_panel.rml
```diff
@@ -76,6 +76,19 @@
       </div>
     </div>
 
+    <div class="impdlg-control-group">
+      <span class="impdlg-field-label text-muted">Centralize Dataset</span>
+      <div class="impdlg-control-main">
+        <div class="impdlg-control-row">
+          <select class="impdlg-input" data-value="centralize_dataset">
```
MrNeRF, line 83: this need proper localization and not hardcoded strings.

MrNeRF (changes requested):
Please see my comments. After removing auto now it is dead code everywhere.

I think that the enum should be used whereever possible and not the string representation.

The localization needs to be added. I also don't think that a single function has to go into a lib if that's used at a single place. 

There is quite a bit of ai slop in here.

src/io/include/io/loader.hpp
dozeri83: i was think maybe save it for the future. but you are right. removing it

src/geometry/geometry_misc.cpp
dozeri83: do you think that they can be used by other modules in the future? if so, i think that they belong to geometry lib

src/io/include/io/loader.hpp
dozeri83: yes. removing it.

src/core/include/core/events.hpp
dozeri83: events.hpp seems like a core file and CentralizeDataset enum belongs to loader. 
adding this enum to events.hpp mean that all files that include events.hpp will be exposed to loader.hpp, 
so i chose to go with the basic string.  

do you still  prefer the enum ?

src/geometry/CMakeLists.txt
dozeri83: removing

dozeri83:
@MrNeRF 
resolved most of conversations, pending your answer on the last two.

MrNeRF:
Ok, thank you! I think it is good. I would remove that from the geometry module. Can be cleaned later. I will merge now!

Issue #1159, opened 2026-04-27
Color issues on mesh to splats converter


Vilour:
This is the splats produced by mesh2splat [repo](https://github.com/electronicarts/mesh2splat):

<img width="361" height="741" alt="Image" src="https://github.com/user-attachments/assets/1bac7eac-9f6a-4fe3-bd41-b0441755d3f8" />

And this is from LichtFeld Studio:

<img width="336" height="728" alt="Image" src="https://github.com/user-attachments/assets/560dbab6-3ef4-44d5-b4e2-6a192be7a844" />

The color of the mesh is similar to the first one in Blender, and the mesh in LichtFeld renders similarly to the second one. I guess there is some issues on mesh rendering or something. The color should be more vivid.

PR #1163, opened 2026-04-28
fix: resolve washed out colors and double gamma in mesh2splat (#1159)

Vilour:
This PR fixes the issue #1159 where splats converted from meshes appear severely washed out and desaturated.

### Root Cause
1. **Texture Upload:** Albedo textures were uniformly uploaded as `GL_RGBA8` instead of `GL_SRGB8`/`GL_SRGB8_ALPHA8` in both `mesh_renderer.cpp` and `mesh2splat.cpp`, causing linear math to be performed on gamma-encoded textures.
2. **Double Gamma & Baked Lighting:** `converterFS.glsl` was calculating unnecessary PBR lighting (adding white specular/ambient wash) and explicitly applying a gamma curve `pow(color, 1.0/2.2)`. However, the C++ pipeline `build_splat_data` expects a linear color to compute the SH0 coefficients. This resulted in baked lighting and a double-gamma effect.

### Solution
- Added an `is_srgb` flag to `create_gl_texture` and `upload_texture`. Albedo maps are now correctly uploaded as `GL_SRGB8_ALPHA8`.
- Stripped the hardcoded lighting calculation in `converterFS.glsl`. It now passes the pure, unlit albedo color and correctly applies the standard gamma encoding for the SH0 representation.

### Visual Proof
**Before (Original mesh2splat):**
image

**After (With this PR):**
image

### DIFF
--- a/src/rendering/mesh2splat.cpp
+++ b/src/rendering/mesh2splat.cpp
@@ -431,7 +431,7 @@ namespace lfs::rendering {
                 break;
             case 3:
                 format = GL_RGB;
-                internal_format = GL_RGB8;
+                internal_format = is_srgb ? GL_SRGB8 : GL_RGB8;
                 break;
             case 4:
                 format = GL_RGBA;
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thx for the fix. Will review it later today!

src/rendering/mesh2splat.cpp
MrNeRF: can be also fixed, no? 
is_srgb ? GL_SRGB8 : GL_RGB8;

src/rendering/resources/shaders/mesh2splat/converterFS.glsl
```diff
@@ -130,11 +130,7 @@ void main() {
     vec3 diffuse = kD * albedo / PI;
     vec3 spec = (NDF * G * F) / (4.0 * max(dot(N, V), 0.0) * NdotL + 0.0001);
```
MrNeRF: this is all dead now? Could be removed including the  vars.

MrNeRF (commented):
Please see my comments

Vilour:
Thanks for the review! I've removed the dead lighting code and updated the case 3 texture format.

MrNeRF:
Thx! Merged!

PR #1175, opened 2026-05-01
Csv file browser

dozeri83:
added csv file dialog (need it for a plugin)


MrNeRF:
Let's add that. I think for the future this should be injectable by a plugin so not every file format needs to be added to the core.

PR #1193, opened 2026-05-05
Fix compiler and linker warnings across the build

MotivaCG:
A large number of warnings were being emitted on every build, making it difficult to spot genuine issues in the output. When the warning log is hundreds of lines long, real regressions — type mismatches, uninitialized variables, ABI breaks — get buried and go unnoticed. This change brings the build to a clean state so that any new warning stands out immediately.

Fixes by category:

- Discarded [[nodiscard]] results (C4834 / #3004-D): Return values marked [[nodiscard]] were being silently dropped. Each call site is now wrapped in an if check that either propagates the error or logs it, depending on whether the surrounding function can return a failure.
- Narrowing conversions (C4244 / C4267): Implicit truncations from double to float and from size_t to int are made explicit with static_cast, clarifying intent and silencing the warning.
- Shadowed declarations (C4458 / C4456): Local variables or parameters sharing a name with a class member or an outer-scope variable are renamed to remove the ambiguity.
- Printf format mismatch (C4477): %ld used for int64_t arguments is replaced with PRId64 from <cinttypes>, which resolves correctly on both Windows and Linux.
- GLM compiler-detection macro redefinition (C4005): GLM_COMPILER was being set to 0 as a compile definition on CUDA targets, conflicting with GLM's own platform.h which unconditionally redefines it. The external definition is removed; GLM's autodetection is left to run unobstructed.
- Unsafe strerror (C4996): Replaced with a portable conditional that uses strerror_s on MSVC and std::strerror elsewhere.
- Third-party warning suppression (indicators, nanoflann, nvImageCodec): External headers and sources were being compiled with full project warning levels, flooding the log with noise from code we do not own and cannot fix. These are now isolated via CMake SYSTEM include directories and NVCC --diag-suppress flags, without modifying anything under external/.
- Linker warnings (LNK2005 / LNK4217): A duplicate-symbol conflict caused by an incorrect export macro on a CUDA static library is fixed, and a harmless dllimport diagnostic on a locally-defined symbol is suppressed at the linker level.
- Unicode box-drawing characters in CMake messages: Several message(STATUS ...) calls used Unicode box-drawing characters (╔, ═, ║, etc.) for decorative banners that rendered as garbage on Windows terminals not running in UTF-8 mode. These are replaced with plain ASCII equivalents so the CMake configuration output is actually readable on any platform.
- 

The expected outcome is a build output clean enough that any new warning is immediately visible and actionable — the goal being that a warning in CI or a local build reliably signals something worth investigating, rather than being dismissed as background noise.

Regards, Víctor


shadygm:
@MotivaCG thanks for the PR, can you check the ubuntu builds and fix them, thanks!

MotivaCG:
> @MotivaCG thanks for the PR, can you check the ubuntu builds and fix them, thanks!

Sure, I'll will take a couple of days probably since last week was mad for me and I'm updating several things!

shadygm:
CI passed and works fine locally. Feel free to merge @MrNeRF

PR #1209, opened 2026-05-09
Fix VRAM leak on PLY/SOG export by trimming CUDA memory pool

### LINKED ISSUE #1196: VRAM usage increases on PLY export
Anytime a PLY or SOG is exported from LFS, the VRAM usage increases, on larger models this can be a significant increase.

jacobvanbeets:
## Problem

VRAM usage increases every time a PLY or SOG is exported from LFS. On larger models the increase is significant and accumulates with each export.

Fixes #1196

## Root Cause

During export, temporary CUDA tensors are allocated for the merged splat data, Morton encoding, and k-means clustering. When these tensors are freed, their memory goes into the \CudaMemoryPool\'s \SizeBucketedPool\ cache (up to 4 entries per bucket) and the CUDA async memory pool (whose release threshold is \UINT64_MAX\). Neither export path called \	rim_cached_memory()\ afterward, so the cached VRAM was never returned to the GPU.

## Fix

Add \Tensor::trim_memory_pool()\ calls after both export paths:
- **Synchronous MCP export** (\xport_scene_nodes\ in \mcp_gui_tools.cpp\): reset the merged \SplatData\ unique_ptr then trim.
- **Async UI export** (\startAsyncExport\ in \sync_task_manager.cpp\): reset the shared_ptr then trim after the export try/catch block.

This follows the same pattern already used after training densification in \improved_gs_plus.cpp\ (lines 652 and 674).

---

### DIFF
--- a/src/app/mcp_gui_tools.cpp
+++ b/src/app/mcp_gui_tools.cpp
@@ -1405,6 +1405,11 @@ namespace lfs::app {
             }
             }
 
+            // Release merged CUDA tensors and trim the memory pool so that
+            // temporary VRAM allocated for the export is returned to the GPU.
+            merged.reset();
+            lfs::core::Tensor::trim_memory_pool();
+
             return {};
         }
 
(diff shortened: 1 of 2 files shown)


MrNeRF:
I am closing as explained on discord. It should be fixed  with https://github.com/MrNeRF/LichtFeld-Studio/commit/a418f2f8e325af5b3a178e62705006ba11257763 and https://github.com/MrNeRF/LichtFeld-Studio/commit/2363bddc81ef33d563dd48d1eed6fc2208a7daab

PR #1222, opened 2026-05-16
Plugin List View Load on Startup

### LINKED ISSUE #1221: 'Load at Startup' option in List View
The "load at startup" option is only visible in card view. 

It can also be shown in list view, since the number of plugins are increasing , its easier to view them in list view, thus all options relating to it.. should be visible.

shadygm:
Resolve #1221


MrNeRF:
lgtm! Thanks!

PR #1229, opened 2026-05-19
First and Second Moment Quantization

shadygm:


shadygm:
Closing since the new swizzle SH layout saved a bit more vram, will look into it in the future.

PR #1231, opened 2026-05-20
Adding a TCP server to broadcast and share events [Volinga]

Ediolot:
We propose a basic and simple way to externalize most of the events produced by LichtFeld-Studio in a way that any number of external applications may connect and track progress and logs while the application is running. 

When in headless mode, the flag `--tcp-connection` enables two tcp servers. The responder server answers requests in JSON format, like a pause request: `{"command":  "pause"}` with response `{"command": "pause", "success": true}`. The publisher server constantly broadcasts events and logs to all connected clients: 

```
{
  "command": "event",
  "data": {
    "level": "info",
    "message": "VMM initialized: device=0, virtual=32 GB"
  },
  "event_type": "log"
}
{
  "command": "event",
  "data": {
    "is_refining": false,
    "iteration": 1,
    "loss": 0.2827635407447815,
    "num_gaussians": 182686
  },
  "event_type": "TrainingProgress"
}
{
  "command": "event",
  "data": {
    "is_refining": false,
    "iteration": 10,
    "loss": 0.2985736131668091,
    "num_gaussians": 182686
  },
  "event_type": "TrainingProgress"
}
...
```

TCP servers use the generic ZMQ library which is available to work with in different languages.

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -17,7 +17,10 @@
 #include "io/cache_image_loader.hpp"
 #include "training/trainer.hpp"
 #include "training/training_setup.hpp"
+#include "visualizer/training/training_manager.hpp"
 #include "visualizer/visualizer.hpp"
+#include "tcp/include/tcp_publisher.hpp"
+#include "tcp/include/tcp_responder.hpp"
 
 #include "app/mcp_gui_tools.hpp"
 #include "io/video/video_encoder.hpp"
@@ -173,19 +176,40 @@ namespace lfs::app {
                     }
 
                     auto trainer = std::make_unique<training::Trainer>(scene);
+                    auto manager = std::make_shared<vis::TrainerManager>();
 
                     if (!params->python_scripts.empty()) {
                         trainer->set_python_scripts(params->python_scripts);
                         vis::gui::panels::PythonScriptManagerState::getInstance().setScripts(params->python_scripts);
                     }
 
-                    if (const auto result = trainer->initialize(*params); !result) {
-                        LOG_ERROR("Failed to initialize trainer: {}", result.error());
-                        return 1;
-                    }
+                    trainer->setParams(*params);
+                    manager->setTrainer(std::move(trainer));
 
                     core::Tensor::trim_memory_pool();
 
+                    if (params->optimization.tcp_connection) {
+                        tcp::ResponderServer responder(params->optimization.tcp_server_connection_port, manager);
+                        tcp::PublisherServer publisher(params->optimization.tcp_broadcast_connection_port, manager);
+
+                        responder.start();
+                        publisher.start();
+                        LOG_INFO("Responder server listening on {}", responder.getEndpoint());
+                        LOG_INFO("Publisher server listening on {}", publisher.getEndpoint());
+
+                        manager->startTraining();
+                        manager->waitForCompletion();
+
+                        publisher.stop();
+                        responder.stop();
+                        responder.join();
+                    }
+                    else {
+                        manager->startTraining();
+                        manager->waitForCompletion();
+                    }
+
+                    // TODO
                     if (const auto result = trainer->train(); !result) {
                         LOG_ERROR("Training error: {}", result.error());
                         if (!params->python_scripts.empty()) {
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -175,16 +175,19 @@ namespace lfs::app {
                         return 1;
                     }
 
-                    auto trainer = std::make_unique<training::Trainer>(scene);
                     auto manager = std::make_shared<vis::TrainerManager>();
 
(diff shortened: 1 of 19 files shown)


src/tcp/tcp_publisher.cpp
```diff
@@ -0,0 +1,156 @@
...
+        }
+    );
+
+    SUBSCRIBE_EVENT(TrainingStarted);
+    SUBSCRIBE_EVENT(TrainingProgress);
+    SUBSCRIBE_EVENT(TrainingPaused);
+    SUBSCRIBE_EVENT(TrainingResumed);
+    SUBSCRIBE_EVENT(TrainingCompleted);
+    SUBSCRIBE_EVENT(TrainingStopped);
+    //SUBSCRIBE_EVENT(SceneLoaded);
```
MrNeRF: Commented out code should be simply removed

src/tcp/tcp_server.cpp
```diff
@@ -0,0 +1,66 @@
...
+{
+    port_ = std::max(port_, 0); // Port == 0 sets automatic port
+    socket_.bind("tcp://*:" + std::to_string(port_));
+    auto str_endpoint = getEndpoint();
+    auto str_port = str_endpoint.substr(str_endpoint.find_last_of(':') + 1);
+    port_ = std::stoi(str_port);
+}
+
+void lfs::tcp::TCPServer::stop() {
+    ;
```
MrNeRF: this is unnecessary and should be removed.

src/tcp/tcp_server.cpp
```diff
@@ -0,0 +1,66 @@
...
+}
+
+void lfs::tcp::TCPServer::join() {
+    ;
```
MrNeRF: same here and all other cases that have that pattern.

src/tcp/tcp_publisher.cpp
```diff
@@ -0,0 +1,156 @@
...
+    ENABLE_TO_JSON(ExportCompleted, path, format);
+    ENABLE_TO_JSON(ExportFailed, error);
+    ENABLE_TO_JSON(VideoExportCompleted, path, total_frames);
+    ENABLE_TO_JSON(VideoExportFailed, error);
+
+    // CUDA version check
+    ENABLE_TO_JSON(CudaVersionUnsupported, major, minor, min_major, min_minor);
+}
+
+lfs::tcp::PublisherServer::PublisherServer(int port, std::shared_ptr<lfs::vis::TrainerManager> trainer_manager, core::LogLevel level, bool warm_up)
```
MrNeRF: the pattern in the cpp files should be 

namespace lfs::tcp {
}
instead of fully qualified so you can wrap all code in there.

src/tcp/include/tcp_server.hpp
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include <memory>
+#include <zmq.hpp>
+#include <visualizer/training/training_manager.hpp>
+#include <nlohmann/json.hpp>
+
+namespace lfs::tcp {
+    class TCPServer {
+        static constexpr int kNumberOfThreads = 2;
```
MrNeRF: why is that? Short comment?

src/training/checkpoint.cpp
```diff
@@ -292,6 +293,9 @@ namespace lfs::training {
                     if (params_json.contains("dataset")) {
                         params.dataset = lfs::core::param::DatasetConfig::from_json(params_json["dataset"]);
                     }
+                    if (params_json.contains("server")) {
```
MrNeRF: why do the checkpoints get a server flag?

src/app/application.cpp
MrNeRF: I think the pr should not change the headless path. It should have its own tcp server path. Please restore the headless path.

MrNeRF (changes requested):
Hey, thanks for the pr.
I think the biggest blocker here is the change of the headless path. That should be restored. I would suggest to copy that over into its own tcp path which gets' only triggered on the specified flag. 

Please see also my other comments.

src/tcp/include/tcp_server.hpp
Ediolot: The number of threads used for ZMQ tcp network, I'll add the comment

src/training/checkpoint.cpp
Ediolot: This was a mistake 👍

src/app/application.cpp
Ediolot: I have modified the PR to add its own path for the TCP+headless, originally, we just leveraged the headless path because both are very similar use cases

Ediolot:
I have updated the PR. Dont hesitate to request any other changes you may see fit !

MrNeRF:
Ok, this looks good to me!

PR #1233, opened 2026-05-20
Fix masks not loading or applying (#1198)

### LINKED ISSUE #1198: Masks not loading
Hello,

I carefully followed the video about masking convention (https://www.youtube.com/watch?v=956qR8N3Xk4).
When I import my project, the masks are loaded (mask icon is visible) but they appear in the image preview as alway white when I press show mask overlay.
The mask are in png with only pure white and pure black, with the same resolution. I also tried jpeg (same error)

Here is the error for one image mask:
[11:26:43.426] [warn] rmlui_render_interface.cpp:897  RmlUI LoadTexture failed: F:/LichtFeld-Studio-windows-v0.5.2/share/LichtFeld-Studio/assets/rmlui/\2026_4GS\TextModel2400\masks\50\03_27\27_00243.png

Can you help ?
Thanks

parodiniccolo:
## Summary

Fixes #1198 — masks not loading / not applied during training.

Datasets that ship **RGBA images alongside real per-pixel mask files** were silently ignoring the sidecar masks. This is common with Reality Scan and Unreal Engine exports, where the image's embedded alpha channel is a degenerate constant `255` (fully opaque) but the actual segmentation lives in `masks/*.png`. Reproduced on two such datasets.

Three independent issues are fixed:

### 1. Training & eval ignored sidecar masks (alpha-as-mask took precedence)
`use_alpha_as_mask` (default `true`) won over the sidecar mask file. With a constant alpha=255 the mask covered the whole frame, so `Segment`/`Ignore` modes had no effect and gaussians kept spawning in masked-out regions. PR #832 introduced this precedence; before it, sidecar masks were always preferred.

**Fix:** sidecar mask wins when present; alpha-as-mask is only used as a fallback for cameras without a sidecar. Applied in the training dataloader (`dataset.hpp`) and the eval/metrics input loader (`trainer.cpp`). The startup log now reports the actual source.

### 2. Preview mask overlay never loaded the texture
The image-preview panel passes an absolute mask path to RmlUI percent-encoded (`C:%5CUsers%5C...png`). `pathReferenceToFilesystemPath` didn't recognise the encoded form as absolute, so RmlUI joined it against the document directory and produced a bogus `assets/rmlui/\Users\...` path — the original symptom reported in #1198 (`RmlUI LoadTexture failed`).

**Fix:** percent-decode and re-test for absolute before giving up (`rml_path_utils.hpp`).

### 3. Loaded mask overlay tinted the whole image
A single-channel mask loaded via `stbi_load(req_comp=4)` becomes `(gray,gray,gray,255)` — fully opaque. The RmlUI backend blends with **premultiplied alpha** (`ONE / ONE_MINUS_SRC_ALPHA`), so even alpha=0 pixels add their colour.

**Fix:** remap masks to a premultiplied `(gray,gray,gray,gray)` texture so the gray value drives transparency; the `image-color` CSS tint then paints only the foreground (`rmlui_vk_backend.cpp`).

## Testing

Built on Windows (MSVC) and run on the affected datasets:

- **Training:** log now shows `Found 160 sidecar masks` (was `Using alpha channel as mask source`); Segment mode correctly suppresses background, confirmed on a 30k-iter run.
- **Preview overlay:** the red mask overlay now highlights only the foreground object instead of washing the entire frame (verified with an opaque-tint test isolating foreground vs background).

## Notes

- No behaviour change for alpha-only datasets (no sidecar): alpha-as-mask still applies via the fallback.
- The eval-path `MetricsEvaluator::load_eval_mask` was already sidecar-first; only the metrics *input* loader needed the same guard.

### DIFF
--- a/src/training/dataset.hpp
+++ b/src/training/dataset.hpp
@@ -626,14 +626,14 @@ namespace lfs::training {
                     request.params.undistort = request.undistort;
                 }
 
-                if (mask_config_.use_alpha_as_mask && cam->has_alpha()) {
-                    request.extract_alpha_as_mask = true;
-                    request.alpha_mask_params.invert = mask_config_.invert_masks;
-                    request.alpha_mask_params.threshold = mask_config_.mask_threshold;
-                } else if (mask_config_.load_masks && cam->has_mask()) {
+                if (mask_config_.load_masks && cam->has_mask()) {
                     request.mask_path = cam->mask_path();
                     request.mask_params.invert = mask_config_.invert_masks;
                     request.mask_params.threshold = mask_config_.mask_threshold;
+                } else if (mask_config_.use_alpha_as_mask && cam->has_alpha()) {
+                    request.extract_alpha_as_mask = true;
+                    request.alpha_mask_params.invert = mask_config_.invert_masks;
+                    request.alpha_mask_params.threshold = mask_config_.mask_threshold;
                 }
 
                 loader_->prefetch({request});
--- a/src/training/trainer.cpp
+++ b/src/training/trainer.cpp
@@ -287,7 +287,11 @@ namespace lfs::training {
                 mask_mode == lfs::core::param::MaskMode::Segment ||
                 mask_mode == lfs::core::param::MaskMode::Ignore;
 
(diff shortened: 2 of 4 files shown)


MrNeRF:
That's great! Thank you!

PR #1243, opened 2026-05-24
set home position using SHIFT-H

### LINKED ISSUE #1216: Wishlist: Be able to set Home position
Pretty straightforward ... when navigating, be able to use a key to specify that current viewpoint is to be used as Home from now on.

kmeirlaen:
Use Shift-H to set home position for current session.

Fixes https://github.com/MrNeRF/LichtFeld-Studio/issues/1216


kmeirlaen:
after building this branch, i cannot use any selection tool any more..

kmeirlaen:
> after building this branch, i cannot use any selection tool any more..

fixed

MrNeRF:
ok, lgtm. Nice minimal impl!

PR #1254, opened 2026-05-26
RAD LOD Viewer

shadygm:


MrNeRF:
Still at functionality level
1. Importing rad via drag and drop does not activate lod
2. It is static, i.e. it seems it simply constantly renders in the viewport the max number that is set (in my case 1.5m)
3. Given 2, there is no lod rendering nor can I see any hud or whatever that shows me how many are rendered, etc.

PR #1268, opened 2026-06-02
fix(scene): render plugin-inserted splats (add_splat → Vulkan-external storage)

lyehe:
## Summary

`scene.add_splat` clones its input tensors into plain CUDA storage, so the Vulkan splat renderer — which refuses the input-copy fallback and requires `vulkan_external_buffer` storage — can't bind them. A splat inserted by a Python plugin therefore renders nothing:

> `VkSplat refusing full input-copy fallback; model tensors must use Vulkan-external storage`

The file loader already guarantees renderer-ready models via `migrateSplatTensorsToAllocator`. This routes `PyScene::add_splat` through the same path:

- **io**: expose `migrateSplatTensorsToAllocator` (was file-local in `loader_service.cpp`)
- **`SceneManager::migrateSplatToRendererStorage`**: migrate a model through the viewer's Vulkan-external allocator (no-op when interop is unavailable / headless)
- **`PyScene::add_splat`**: migrate the cloned model before handing it to the scene

Un-breaks rendering for plugins that insert via `scene.add_splat` (TripoSplat, HY-World-Mirror, ml-sharp, …).

## Scope / notes

- **Single inserted model only.** A separate gap remains for the **combined model** of 2+ visible splats: `combined_model_allocator_` is only installed on file/dataset load, so plugin-built multi-splat scenes still fall back to plain storage. Intentionally left out of this PR.
- `loader_service.cpp`'s large diff is just the migrate helper moving out of the file-local anonymous namespace into `lfs::io` (clang-format re-indent); no logic change.

## Test plan

- [ ] Insert a splat from a Python plugin (`scene.add_splat`) → renders in the Vulkan viewport (was blank with the VkSplat error).
- [ ] Load a `.ply` / checkpoint → unchanged (migrate is a no-op for already-external models).
- [ ] Headless / no Vulkan-external interop → no crash (migration no-ops, logs a warning).

### DIFF
--- a/src/io/include/io/loader.hpp
+++ b/src/io/include/io/loader.hpp
@@ -42,6 +42,12 @@ namespace lfs::io {
     using ProgressCallback = std::function<void(float percentage, const std::string& message)>;
     using CancelCallback = std::function<bool()>;
 
+    // Re-home a model's parameter tensors into the given allocator (Vulkan-external interop)
+    // so the splat renderer can bind them. No-op when tensors are already allocator-backed or
+    // when the allocator is empty. Used by the loader and by Scene::add_splat (plugin inserts).
+    [[nodiscard]] Result<void> migrateSplatTensorsToAllocator(SplatData& model,
+                                                              const SplatTensorAllocator& allocator);
+
     // Dataset type enum
     enum class DatasetType {
         Unknown,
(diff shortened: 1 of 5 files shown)


lyehe:
Superseded by the maintainer's dceb8f8d ("Fix ply loading via api"), which routes `PyScene::add_splat` through `io::migrateSplatTensorsToAllocator` (same fix as this PR) and additionally installs the combined-model allocator (`setCombinedModelAllocator`) for the 2+-visible-splat case. Closing in favor of that. Thanks!

PR #1286, opened 2026-06-08
feat(python): clipboard image read API for plugins

lyehe:
## What

Adds a Python clipboard **image read** API so plugins can pull an image straight off the system clipboard:

- `lf.ui.has_clipboard_image() -> bool`
- `lf.ui.get_clipboard_image_texture() -> (texture_id, w, h)`
- `lf.ui.save_clipboard_image(path) -> bool`

These complement the existing write-only `set_clipboard_text` (the only clipboard binding today).

## How

- Enumerates SDL's advertised clipboard MIME types and takes any `image/*`. Browsers expose `image/png`; native Windows copies expose `image/bmp` (SDL maps `CF_DIB`/`CF_DIBV5` → `image/bmp`), so it works across sources without hard-coding a list.
- Decodes the bytes through the existing `core::load_image_from_memory` (OpenImageIO).
- `get_clipboard_image_texture` uploads via the same `create_ui_texture_from_data` path as `load_image_texture`; `save_clipboard_image` writes the decoded image via `core::save_img_data`.
- Owning image buffers freed with `core::free_image`; SDL allocations with `SDL_free`.
- Helpers factored into the existing anonymous namespace alongside `create_ui_texture_from_data`; no new includes required.

## Notes

- Decoded as RGB (alpha dropped), matching `load_image_from_memory`.
- New stubs added to `lichtfeld/ui/__init__.pyi`.
- Verified: full build links (`lichtfeld.*.pyd` + `LichtFeld-Studio.exe`), and live end-to-end on Windows (clipboard image → `save_clipboard_image` wrote a valid PNG; `has_clipboard_image` gated the action correctly).

## Use case

Lets plugins offer a "Paste image" action (e.g. single-image 3D generators) instead of only a file dialog.

### DIFF
--- a/src/python/lfs/py_ui.cpp
+++ b/src/python/lfs/py_ui.cpp
@@ -583,6 +583,61 @@ namespace lfs::python {
             return {static_cast<uint64_t>(result.texture_id), result.width, result.height};
         }
 
+        // SDL returns clipboard payloads keyed by MIME type, chosen by the source app
+        // (browsers use image/png, native copies may use image/bmp, etc.), so scan every
+        // advertised image/* type rather than guessing a fixed set.
+        bool clipboard_has_image() {
+            size_t count = 0;
+            char** mimes = SDL_GetClipboardMimeTypes(&count);
+            if (!mimes)
+                return false;
+            bool found = false;
+            for (size_t i = 0; i < count; ++i) {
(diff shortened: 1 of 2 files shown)


lyehe:
Slop upgrade

MrNeRF:
Nice. I should remove the std::tuple from the api as it is a performance sink. Other than that it looks good. All mem freed :)

PR #1287, opened 2026-06-08
Docked Asset Manager

shadygm:
To be merged after #1254 

image


MrNeRF:
The resize of this area behaves very weird. 
image

MrNeRF:
More issues. That's just functional for now:
- Undocking does not work. It stays docked. If that's docked on the size, not sure whether the docking handle should be at the top
image

- Why is there a max width for resizing it?
- add tag is broken
- long list scrolling is laggy
- what else?

MrNeRF (approved):

PR #1289, opened 2026-06-09
Add SPHERICAL (equirectangular) camera model to the COLMAP loader

zalo:
Dependent on https://github.com/colmap/colmap/pull/4441

LichtFeld already supports the EQUIRECTANGULAR camera model natively in the CUDA/Vulkan rasterizers, UI, and the transforms.json loader, but the COLMAP loader did not recognize the SPHERICAL model emitted by equirectangular COLMAP forks (e.g. colmap/colmap#4441), so 360 panorama reconstructions could not be loaded from COLMAP output.

src/io/formats/colmap.cpp:
- Register SPHERICAL (COLMAP model id 17, params [width, height]) in the CAMERA_MODEL enum, id/name maps, name lookup, and intrinsic scaling.
- Handle it in both the binary and text camera parse paths, mapping it to CameraModelType::EQUIRECTANGULAR. "EQUIRECTANGULAR" is accepted as an alias name. A placeholder focal length is set (the EQUIRECTANGULAR rasterizer derives the projection from the image dimensions), mirroring the transforms.json convention.

src/core/camera.cpp:
- Exclude EQUIRECTANGULAR from has_distortion(): it is a complete projection model handled natively by the rasterizer, with no pinhole image to undistort to.

Equirectangular COLMAP reconstructions train with --gut (3DGUT).

### DIFF
--- a/src/core/camera.cpp
+++ b/src/core/camera.cpp
@@ -589,6 +589,11 @@ namespace lfs::core {
             return true;
         if (_tangential_distortion.is_valid() && _tangential_distortion.numel() > 0)
             return true;
+        // Equirectangular is a complete projection model handled natively by the
+        // rasterizer; there is no pinhole image to "undistort" to, so it must not
+        // be routed through the undistortion path.
+        if (_camera_model_type == CameraModelType::EQUIRECTANGULAR)
+            return false;
         if (_camera_model_type != CameraModelType::PINHOLE)
             return true;
         return false;
--- a/src/io/formats/colmap.cpp
+++ b/src/io/formats/colmap.cpp
@@ -167,7 +167,12 @@ namespace lfs::io {
         SIMPLE_RADIAL_FISHEYE = 8,
         RADIAL_FISHEYE = 9,
         THIN_PRISM_FISHEYE = 10,
-        UNDEFINED = 11
+        UNDEFINED = 11,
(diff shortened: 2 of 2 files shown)


MrNeRF:
Thx for contributing. Any signs that this gets merged into colmap?

zalo:
> Thx for contributing. Any signs that this gets merged into colmap?

@ahojnnes seems to like it in this comment: https://github.com/colmap/colmap/pull/4440#issuecomment-4625946119

Also here is an example 30k iter 3DGS trained from the example sequence in the COLMAP PR description: 

Video Sequence:

https://github.com/user-attachments/assets/ae416c79-87e1-4b82-b55f-87abdb1ab164

Gaussian Splat: https://superspl.at/scene/e043c423

zalo:
> Thx for contributing. Any signs that this gets merged into colmap?

We got it merged into COLMAP! https://github.com/colmap/colmap/pull/4441 Let the era of direct from spherical begin!

With Caspar support in https://github.com/colmap/colmap/pull/4463/ and the existing ALIKED/LightGlue support it'll probably also be the fastest way to process large quantities of equirect footage.

MrNeRF:
Awesome! I will review it later and make sure that we get it into master today.

MrNeRF:
seems good!

PR #1294, opened 2026-06-11
lod_splat: budget-constrained LOD selection module (CPU reference + benchmarks)

maxandrews:
## Summary

This adds `src/lod_splat/` — a standalone, dependency-free C++20 module implementing a **budget-constrained LOD scheme** for Gaussian splats, plus a profiling/validation harness. It does not modify any existing code and is not wired into the main build.

> **⚠️ CPU-only reference implementation.** This was developed and profiled on Apple Silicon (M4 Pro) **without CUDA hardware** — the main app cannot build there, so the module compiles standalone (`cmake -S src/lod_splat -B build-lod`) and all rendering/benchmarks below use its CPU reference rasterizer. The selection/streaming logic is designed to map onto the existing GPU path (`ViewportRenderRequest::lod_indices/lod_weights`, `LodPageCache`), but no GPU integration is included. I'm opening this as a working reference + benchmark set for discussion; happy to split it up, rework it, or park parts of it.

## How it differs from the current LOD stack

The shipped stack (`bhatt_lod` + `lod_select_threshold.slang`) selects the cut with a **uniform pixel-scale threshold**: every resident node is tested independently against `pixel_scale_limit`, and `output_capacity` truncates overflow arbitrarily (whichever GPU threads append last). There is no notion of a splat budget — budget control has to be a reactive feedback loop on the threshold across frames.

This module replaces the threshold with a **hard budget + greedy allocation**:

- Per-frame selection maintains an explicit hierarchy cut and refines the **worst screen-space-error node first** (max-heap, error × screen coverage), stopping exactly at a splat budget. Concrete example of the difference: when a street-level camera climbs to a city overview and visible content jumps ~20×, the threshold approach must either blow past its output capacity (dropping arbitrary nodes, possibly foreground) or oscillate while a controller re-tunes the limit — here the cut force-coarsens cheapest-regions-first within the same frame and the working set never exceeds the cap (measured max = budget exactly, 1,500,000). A bounded working set also deterministically bounds the per-frame depth sort, which stayed flat at 12–15 ms through the overview reveal in the flythrough profile.
- The error metric is the **representatives' actual world extent** vs the pixel footprint (`extent · focal/d · √q`), not node-bounds radius over distance — so the quality knob `q` is a perceptual overdraw target (measured contributions/pixel track it monotonically), and sparse nodes with small splats aren't over-refined by their bounds.

Several pieces are deliberately **ported from / shared with the existing stack**, because they're good: Spark `lodOpacity` (>1 merged alpha; integrated-alpha conservation measured at 0.99 vs 0.34 with clamping), Bhattacharyya-similarity merge pairing, the continuous parent↔children dissolve from `lod_select_threshold.slang` (stateless, popping impossible by construction), and the parent-fallback / request-evicted-parent streaming behaviors (we independently hit the same zoom-out residency deadlock your shader comments describe, and fixed it the same way).

The build side differs structurally: instead of sequential greedy agglomeration into a binary tree, it's a Morton-order **parallel octree** with moment-matched representatives, where **leaf size = cut granularity**. The leaf-size sweep was the key result: at leaf=64 the cut granularity gets close enough to per-splat that it outperforms the binary tree under budget pressure, while the build stays embarrassingly parallel.

## Benchmarks

Synthetic 29.8M-splat city scene, 1280×720, q=3. For a like-for-like comparison the PR includes `bhatt_ref.{hpp,cpp}` — a faithful CPU port of `bhatt_lod` + the selection shader (verbatim formulas/constants; gaze cone disabled, `pixel_scale_limit` binary-searched to match the budget count).

**Quality at 1.5M splat budget (PSNR vs full 29.8M render):**

| view | this module (leaf=64) | bhatt + threshold (CPU port) |
|---|---|---|
| street, budget-saturated (5%) | **37.4 dB** | 32.3 dB |
| city overview | **32.0 dB** | 31.6 dB |

**Budget sweep (street / overview dB):** 1.5M → 37.4/32.0 · 3M → 43.5/33.6 · 5M → 60.2/35.9.

**Offline build, 29.8M splats (M4 Pro, 14 cores):** octree + merges **2.5 s** vs sequential Bhattacharyya clustering **85 s** (same machine, same scene). Hierarchy memory overhead ~2× input either way (density halves per level, ratio measured 0.500).

**Per-frame CPU cost (600-frame flythrough, streaming simulated at 800 MB/s / 2 ms):** selection 4.6 ms mean / 11.9 ms p95 (incremental, main-thread); depth sort of the 1.5M working set 12 ms (radix) vs 47 ms (`std::sort`); budget never exceeded; zero holes (blocked refines keep the parent visible); fetch wait 6.7 ms avg.

**Comparison frames** (in `src/lod_splat/docs/comparison/`, both methods at 1.5M splats — 5% of the scene):

| | street (saturated) | overview |
|---|---|---|
| ground truth |  |  |
| this module |  |  |
| bhatt + threshold |  |  |

## Reproducing

```sh
cmake -S src/lod_splat -B build-lod -DCMAKE_BUILD_TYPE=Release && cmake --build build-lod -j
./build-lod/lod_profile --splats 30000000 --frames 600 --validate   # profile + quality/temporal validation
./build-lod/lod_profile --splats 30000000 --compare --budget 1500000 --no-streaming  # 3-way comparison renders
./build-lod/lod_profile --ply scene.ply --validate                  # real 3DGS PLY scenes
```

A startup self-test asserts cut coverage (every leaf splat covered exactly once) and integrated-alpha conservation each run.

## Caveats

- Quality/behavior comparison only — **no GPU performance claims**; the shipped selection runs in microseconds on GPU, this selector is ~5 ms on CPU (but incremental and budget-bounded).
- SH degree 0 only at merged levels; synthetic-scene benchmarks (the harness loads real PLYs, but no real-scene numbers yet).
- Fine leaves mean small streaming chunks (~9× more requests on the flythrough); a production pager should aggregate chunks across nodes like the existing 65k-splat RAD pages do.
- The bhatt CPU port disables gaze-cone foveation (no eye tracking in the comparison) and matches budgets by searching the threshold — details in `bhatt_ref.hpp`.


MrNeRF:
I took some ideas from here so I acknowledged you as contributor. Closing.

PR #1314, opened 2026-06-17
New mask mode combining Ignore and Segment [Volinga]

Ediolot:
We have found ourselves in the need of an extra masking mode that combines both the segment (e.g. for the sky) and ignore (for specific components of the scene) in a single training.

An attempt to add this mode is proposed in this PR, called `SegmentAndIgnore`. One caveat is that this mode does not permit a user configurable threshold since the masks are expected to be single files and are split in 3 bands. Each band is made big enough so JPEG compression won't scramble them together. We went through the single file approach since it seemed to be a much faster implementation.

```
0 -> 128 elements to ignore
128 -> 250 elements to segment
250 -> 255 elements to keep
```

> Interface
>
> UI

Below there is an example of the same 10K training.

> Original image (no alpha channel) + mask
> 
> original

> Different masking modes (10K iter MCMC)
>
> changes


MrNeRF (approved):
We verified and tested. Code looks also solid. Thank you!

PR #1318, opened 2026-06-23
Global Tool-Control Shortcuts

kmeirlaen:
## Summary

This change makes tool/control activation shortcuts work regardless of the currently active left-toolbar tool, while keeping tool-local operational shortcuts scoped to their owning tool.

Before this change, shortcuts such as selection submode keys could be bound to a tool mode and therefore only resolved when that tool was already active. This made shortcuts like `Ctrl+1` / `Ctrl+2` for selection modes fail from other toolbar tools, even though their purpose is to activate a control mode rather than perform an operation inside the active tool.

## Problem

The input dispatcher resolved keyboard shortcuts only against the current `ToolMode`, then inherited eligible global bindings. If a shortcut was stored under another tool mode, it was invisible unless that tool was active.

This especially affected activation-style shortcuts:

- Main toolbar shortcuts, such as `1`, `2`, `3`, etc.
- Selection submode shortcuts, such as `Ctrl+1`, `Ctrl+2`, etc.
- Future tool-control activation shortcuts that should switch to their owning tool first.

The old behavior also diverged from toolbar clicks. Clicking a selection submode first activates the selection tool and then applies the submode. The keyboard path only applied the submode when the action was already resolved.

## Design

The fix introduces a distinction between:

- **Tool/control activation shortcuts**: shortcuts that select a tool or control mode and should be reachable from any active tool.
- **Tool-local operational shortcuts**: shortcuts that perform work inside a tool and should remain scoped, such as selection drags, crop apply, polygon confirmation, brush resize, node picking, and delete behavior.

Runtime routing now does this:

1. Resolve the key in the current tool mode as before.
2. If nothing is found, search the other tool modes only for whitelisted tool/control activation actions.
3. If such an action is found, activate its owning toolbar tool.
4. Apply any extra control state, such as selection submode.

This keeps the behavior intentionally narrow around activation actions, while avoiding a selection-only special case.

## Profile Migration

Input profile version was bumped to `16`.

When loading older profiles, tool/control activation bindings stored in a non-global mode are promoted to `ToolMode::GLOBAL` when safe:

- Exact duplicate global bindings are removed from the local mode.
- Alternate user triggers for the same action are preserved by promoting them.
- Bindings are left local if their trigger conflicts with an existing global trigger.

This migration makes existing user profiles behave consistently with the new runtime routing without broadly globalizing operational shortcuts.

### DIFF
--- a/src/visualizer/input/input_bindings.cpp
+++ b/src/visualizer/input/input_bindings.cpp
@@ -26,7 +26,7 @@ namespace lfs::vis::input {
 
     namespace {
 
-        constexpr int PROFILE_VERSION = 15; // Version 15 adds crop apply Enter bindings.
+        constexpr int PROFILE_VERSION = 16; // Version 16 promotes tool/control activation bindings to global.
         constexpr int REMOVED_TOOL_MODE_2 = 2;
         constexpr int REMOVED_ACTION_39 = 39;
         constexpr int REMOVED_ACTION_66 = 66;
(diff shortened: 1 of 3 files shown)


kmeirlaen:
duplicate/improved version: https://github.com/MrNeRF/LichtFeld-Studio/pull/1319

PR #1323, opened 2026-06-24
Feature: Camera Frustum Visibility Shortcut

kmeirlaen:
## Summary

This change adds a configurable shortcut for toggling dataset camera frustums. The default binding is **Alt+C**, and the action is available under the **View** section of Input Settings.

The camera-frustum master render setting is the single global source of truth. The shortcut, Rendering-panel checkbox, and top-level **Cameras** eye all operate on or reflect that same setting.

## Problem

Camera frustums are frequently toggled while training, but the scene tree was previously the only convenient place to control them. This required moving the pointer away from the viewport and expanding or locating the camera hierarchy.

An initial shortcut implementation exposed an important state mismatch: toggling the render setting hid the frustums without updating the scene-tree eye, while toggling ordinary scene-node visibility could not re-enable frustums when the master setting remained disabled.

## Behavior

- **Alt+C** toggles all camera-frustum rendering.
- The binding is configurable in **Input Settings > Global > View**.
- The top-level **Cameras** eye reflects the master camera-frustum setting:
  - Green when camera-frustum rendering is enabled.
  - Red when camera-frustum rendering is disabled.
- Clicking the top-level Cameras eye toggles the same master setting.
- The Rendering-panel **Camera Frustums** checkbox remains synchronized with the shortcut and Cameras eye.
- Training, Validation, and individual camera eyes continue to control their local scene-node visibility independently.

### Render-on-demand synchronization

The Scene panel uses render-on-demand caching. Its synchronization stamp now includes `render_settings_generation`, ensuring changes from the Rendering checkbox or `Alt+C` wake the cached panel and repaint the Cameras eye immediately.

### DIFF
--- a/src/python/lfs/py_keymap.cpp
+++ b/src/python/lfs/py_keymap.cpp
@@ -171,7 +171,8 @@ namespace lfs::python {
             .value("TOOL_ALIGN", Action::TOOL_ALIGN)
             .value("PIE_MENU", Action::PIE_MENU)
             .value("DEPTH_ADJUST_NEAR", Action::DEPTH_ADJUST_NEAR)
-            .value("HISTOGRAM_ZOOM_MARKED", Action::HISTOGRAM_ZOOM_MARKED);
+            .value("HISTOGRAM_ZOOM_MARKED", Action::HISTOGRAM_ZOOM_MARKED)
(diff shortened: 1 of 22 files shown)


MrNeRF:
Not much to comment here. Very clean.

PR #1326, opened 2026-06-25
Window Resize Behavior

kmeirlaen:
## Problem

LichtFeld Studio uses a borderless custom window with its own title bar and docked UI layout. The previous resize path depended on the platform/native resize interaction for the window frame. That made resize behavior feel disconnected from the application render loop.

The visible symptoms were:

- Dragging the side or bottom resize handles did not update the visible layout while the drag was in progress.
- Making the window larger showed no immediate visible change until the resize was released.
- Making the window smaller clipped the existing content during the drag, then redrew the panels only after release.
- The resize cursor could revert to the normal pointer while hovering over or actively dragging resize edges.
- During active resize drags, top toolbar and window-control buttons could still receive hover state, so buttons appeared to light up under the cursor even though the user was resizing the window.
- When the asset manager or right panels were wider than the available window width, the layout could enter a constrained state where the left toolbar disappeared or panel resize/recovery became difficult.

image

Together these issues made window resizing visually unstable.

## Analysis

The core issue was that native borderless resize handling did not provide the application with a continuously usable resize flow. The OS/windowing system resized the outer window, but the UI layout and viewport were not being recomputed and redrawn in a way that matched the drag motion.

There were several secondary issues layered on top of that:

- Cursor state was being influenced by both window-edge resize handling and normal GUI cursor requests.
- Mouse input continued to route through the regular ImGui/UI path during resize drags.
- Hover detection for toolbar and title-bar controls had no concept of "the current mouse drag belongs to window resizing".
- Panel layout constraints were local to individual panels, so the combined width of the left dock, right panel, icon toolbar, and viewport could exceed the available window width.
- The asset manager left dock and the right panel could each independently keep a large width, causing the central viewport and left icon toolbar to be squeezed out visually.

The solution therefore needed to address both resize mechanics and layout recovery, not only redraw timing.

### Dynamic Layout Updates During Resize

`GuiManager` now applies panel width constraints before layout/render decisions are made. This keeps layout state synchronized with the current window size during resize, instead of letting stale panel widths survive until after the drag ends.

This is important because live resizing only feels correct if the visual layout is recomputed on every resize step.

### Manual Borderless Resize Handling

Window resizing was moved into `WindowManager` so the application owns the resize drag lifecycle.

The manual resize path tracks:

- which edge or corner is being resized,
- whether a resize drag is active,
- the initial mouse position,
- the initial window position and size,
- the updated size and position during mouse motion.

This allows the application to update the SDL window size continuously during the drag instead of waiting for a native resize release event. Because the window dimensions now change through the normal application path, the viewport and docked panels can be recalculated every frame while resizing.

### Stable Resize Cursor Ownership

Resize cursor behavior was centralized around explicit resize-edge state. The cursor now remains a horizontal, vertical, or diagonal resize cursor when:

- hovering over a window resize zone,
- pressing on a resize zone,
- dragging after the pointer has moved away from the original edge.

This avoids cursor flicker caused by normal GUI widgets overwriting the resize cursor during an active resize operation.

### GUI Input Suppression During Window Resize

Active resize drags now suppress normal GUI routing for the resize-owned mouse events.

This prevents resize interactions from being interpreted as regular UI hover/click input. In practice this fixes the issue where top toolbar buttons, minimize/restore/maximize/close controls, or other widgets could enter hover state while the user was dragging the window edge.

The resize path still receives the mouse motion it needs, but the rest of the UI does not treat that same drag as a normal pointer interaction.

### Width Constraint Handling

Panel width constraints were added to `PanelLayoutManager` so the combined layout remains recoverable at narrow window sizes.

The constraint logic accounts for:

- the fixed left icon toolbar width,
- a minimum visible viewport width,
- a minimum visible right panel width,
- a minimum visible left dock width when the asset manager is open,
- the current left dock width,
- the current right panel width,
- the available work area after status/title bars.

The right panel and left dock are clamped against the shared available width budget. This prevents either side from consuming all space and hiding critical UI.

### Left Toolbar Preservation

The left icon toolbar is treated as reserved layout space. Even when the asset manager is open and the window is narrowed, the layout should not let the asset manager or right panels cover or push away the icon toolbar.

This keeps the primary tool access and panel recovery path visible.

### Recovery From Over-Constrained States

The resize constraints are enforced continuously, including after panels have previously been made large. This means the application can recover when:

- the asset manager is expanded and the window is then narrowed,
- the right panel is expanded and the window is then narrowed,
- both left and right docked areas compete for space,
- the user re-expands the window after entering a narrow state.

The goal is not only to avoid bad layout states, but also to make sure the user can resize back out of them.

### No Visible Live Resize Feedback

The old path allowed the outer window size to change without forcing the application layout to track each intermediate size. Manual resizing gives the application per-frame control, so layout and rendering update during the drag.

### Content Clipping While Shrinking

Panel sizes are now clamped against the current available work area during resize. This reduces stale widths that would otherwise draw outside the visible region until the resize ends.

### Resize Cursor Flicker

Cursor selection now respects active resize ownership. Once a resize drag starts, normal GUI cursor requests no longer immediately replace the resize cursor.

### Toolbar Hover During Resize

Resize-owned mouse events are withheld from regular GUI routing. This prevents accidental hover feedback on title-bar and toolbar controls while resizing.

### Left Toolbar Disappearing

The left toolbar is included as reserved space in panel width calculations. Left and right panels are constrained around it rather than being allowed to consume that area.

### Asset Manager / Right Panel Width Conflicts

The left dock and right panel now share a constrained width budget with a reserved minimum viewport width. When space is tight, panel widths are reduced instead of allowing one panel to hide the other or remove the resize recovery path.

## Files And Areas Involved

- `WindowManager`
  - Manual borderless resize state and resize-edge detection.
  - Dynamic window size/position updates during resize drags.
  - Resize cursor refresh behavior.

- `VisualizerImpl`
  - Event routing updates so resize-owned events are handled by the resize path.
  - Suppression of regular frame input for resize interactions.

- `GuiManager`
  - Suppression of GUI hover/click routing while a window resize is active.
  - Early enforcement of panel width constraints before layout rendering.

- `PanelLayoutManager`
  - Shared width budget calculations.
  - Left dock and right panel min/max width clamping.
  - Recovery behavior for constrained layouts.

### DIFF
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -5538,6 +5538,13 @@ namespace lfs::vis::gui {
             const auto* mvp = ImGui::GetMainViewport();
             const float status_bar_h = PanelLayoutManager::STATUS_BAR_HEIGHT * current_ui_scale_;
             const float panel_h = mvp->WorkSize.y - status_bar_h;
+            panel_layout_.enforceWidthConstraints(show_main_panel_, ui_hidden_,
+                                                  {
+                                                      .work_pos = {mvp->WorkPos.x, mvp->WorkPos.y},
+                                                      .work_size = {mvp->WorkSize.x, mvp->WorkSize.y},
+                                                      .any_item_active = ImGui::IsAnyItemActive() ||
+                                                                         rmlui_manager_.anyItemActive(),
+                                                  });
 
             ShellRegions shell_regions;
             shell_regions.screen = {mvp->Pos.x, mvp->Pos.y, mvp->Size.x, mvp->Size.y};
@@ -5651,6 +5658,7 @@ namespace lfs::vis::gui {
             screen.work_size = {mvp_input->WorkSize.x, mvp_input->WorkSize.y};
             screen.any_item_active = ImGui::IsAnyItemActive() || rmlui_manager_.anyItemActive();
         }
+        panel_layout_.enforceWidthConstraints(show_main_panel_, ui_hidden_, screen);
 
         constexpr uint8_t kUiLayoutSettleFrames = 3;
         const bool python_console_visible = window_states_["python_console"];
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -63,6 +63,7 @@
 #include "tools/selection_tool.hpp"
 #include "training/trainer.hpp"
 #include "training/training_manager.hpp"
+#include "window/window_manager.hpp"
 #include "visualizer/app_store.hpp"
 #include "visualizer/scene_coordinate_utils.hpp"
 #include "visualizer_impl.hpp"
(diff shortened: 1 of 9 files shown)


kmeirlaen:
copilot remarks handled in https://github.com/MrNeRF/LichtFeld-Studio/pull/1326/commits/f277128973fc8e31f9a9811b53bf34499c62233f

PR #1330, opened 2026-06-25
Resize Training Pause Bugfix

kmeirlaen:
## Symptoms

While training was active, resizing UI surfaces could make the viewport resize path unstable and slow. The issue was reproducible with both top-level window resizing and internal panel resizing.

Observed behavior included:

- Repeated warnings during resize:

  ```text
  Timed out waiting for temporary training pause
  Skipping Vulkan viewport resize because training did not pause in time
  ```

- Training became very slow during resize and sometimes did not recover until the user interacted with the 3D viewport again.
- Resizing an internal panel could pause training and leave it paused until another UI event woke the render loop.
- Double-click maximize/restore on the top application bar could fail to resume training correctly.
- Window resizing could repeatedly pause/resume training while dragging, especially when changing drag direction.
- During manual borderless resize, holding the mouse still and then reversing direction could stop live window updates.
- Plugin panels and docked panels used different resize paths, so fixes that only handled one path did not solve all resize scenarios.

## Analysis

The resize pipeline had two separate concerns that were coupled too tightly:

1. The UI resize interaction lifecycle.
2. The Vulkan viewport image resize that must not race training writes.

The Vulkan path used `pauseTrainingTemporaryAndWait(300ms)` immediately before resizing VkSplat output images. During resize, the viewport size changes frequently. Each change could re-enter this blocking pause path, wait up to 300 ms, fail to observe a paused trainer in time, mark the frame dirty, and retry on the next frame.

That produced a feedback loop:

- Resize changes viewport size.
- VkSplat output image needs a new size.
- Renderer asks training to pause and waits synchronously.
- Training may not reach a pause point within the 300 ms timeout.
- Renderer skips the resize, marks dirty, and tries again.
- The repeated retries slow the UI and training loop.

Earlier attempts that held or released the pause from narrow render paths were fragile because not every resize scenario entered the same code path. Some UI resizes were driven by RmlUi right-panel callbacks, others by ImGui dock layout code, and manual borderless window resize had its own SDL event handling. This made it possible to pause training in one path and only resume it after an unrelated later event.

## Root Cause

The root cause was that resize was treated as a sequence of independent render-frame resize attempts instead of a single resize interaction with a clear begin/end lifecycle.

More specifically:

- VkSplat output resize used a synchronous temporary pause wait inside the render frame.
- Active resizing could trigger many output-size changes before training had time to pause.
- The render path retried immediately after timeout, causing repeated warnings and stalls.
- Internal panel resize state was used for cursor/live-layout behavior, but not consistently forwarded into the rendering resize lifecycle.
- Manual borderless window resizing depended too much on incoming SDL motion events. If events stopped or were routed differently, the active resize could stop updating.
- Manual resize updates also woke the event loop on every update, which risked unnecessary self-wake churn.

### Explicit Resize Interaction Tracking

Added `ui::WindowResizeInteraction` with an `active` flag. Manual borderless window resize now emits:

- `active = true` when manual resize starts.
- `active = false` when manual resize finishes.

`RenderingManager` listens for this event and forwards it into `setViewportResizeActive()`.

### Dock Resize Lifecycle Forwarding

Added `PanelLayoutManager::isResizeInteractionActive()` to distinguish actual active drag resizing from simple hover state.

`GuiManager` now tracks dock resize transitions and calls:

```cpp
viewer_->getRenderingManager()->setViewportResizeActive(active);
```

when bottom dock, left dock, or python console resizing starts or ends.

The existing RmlUi right-panel resize callbacks already called `setViewportResizeActive()`, so the render lifecycle now covers:

- Main window manual resize.
- Right panel width resize.
- Right panel vertical splitter resize.
- Bottom dock resize.
- Left dock / asset manager resize.
- Python console resize.

### Held Resize Training Pause

Added resize pause ownership to `RenderingManager`:

- `requestResizeTrainingPause(TrainerManager*)`
- `releaseResizeTrainingPause()`

During resize deferral, the renderer requests one temporary training pause and holds that lease across the resize interaction. It does not repeatedly pause/resume while the mouse is still dragging.

Training is resumed only when the resize is no longer deferring and the final resize work has been handled or no longer needs to be handled.

### Non-blocking VkSplat Output Resize

Removed the synchronous `ScopedTemporaryTrainingPause` / `pauseTrainingTemporaryAndWait(300ms)` path from `rendering_manager_vulkan.cpp`.

When VkSplat output images need resizing during active training:

- If training is not paused yet, the renderer requests the resize pause, marks resize-related dirty state, returns the cached viewport image, and retries later.
- If training is paused, the renderer performs a CUDA synchronization and proceeds with the output resize.
- No 300 ms wait is performed inside the render frame.

This avoids the repeated timeout warning loop and prevents render frames from blocking while training reaches a pause point.

### Safe Pause Release

The Vulkan render path now releases the held resize pause on idle return paths after resize deferral is over. It also avoids releasing the pause too early when the final VkSplat resize still needs one more frame for the pause request to take effect.

This addresses cases where training previously remained paused until another user interaction woke the renderer.

### Manual Window Resize Updates

Manual borderless window resize now updates outside SDL motion events:

- `pollEvents()` calls `updateManualResize()` once per loop while resize is active.
- `waitEvents()` caps its wait to roughly 60 FPS while resize is active and then calls `updateManualResize()`.
- Mouse-up finishing is handled before checking whether the event targets the current window, so captured resize can finish reliably.
- Mouse motion updates active resize before normal event targeting.

`updateManualResize()` now checks whether position or size actually changed before calling SDL setters. It no longer wakes the event loop on every drag update, avoiding a self-sustaining wake/event loop.

## Result

Resize is now modeled as a single interaction instead of many independent pause attempts. Training pauses once for the resize lifecycle and resumes after the resize settles. Vulkan output resizing no longer blocks repeatedly on a 300 ms pause wait, and internal dock/panel resize paths participate in the same lifecycle as top-level window resizing.

The fix covers the reported scenarios:

- Main window resize while training.
- Right panel resize while training.
- Internal dock/panel resize while training.
- Left asset manager dock resize while training.
- Double-click maximize/restore.
- Holding the mouse during resize and then changing direction.
- Avoiding intermediate pause/resume churn during active drag.

### DIFF
--- a/src/core/include/core/events.hpp
+++ b/src/core/include/core/events.hpp
@@ -252,6 +252,7 @@ namespace lfs::core {
         namespace ui {
             EVENT(FileDropReceived, ); // Emitted when files are dropped onto the window
             EVENT(WindowResized, int width; int height;);
+            EVENT(WindowResizeInteraction, bool active;);
             EVENT(CameraMove, glm::mat3 rotation; glm::vec3 translation;);
             EVENT(SpeedChanged, float current_speed; float max_speed;);
             EVENT(ZoomSpeedChanged, float zoom_speed; float max_zoom_speed;);
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -5993,6 +5993,12 @@ namespace lfs::vis::gui {
         if (!hasMouseButtonDown(sdl_input))
             left_dock_pointer_live_capture_ = false;
 
+        const bool dock_resize_interaction_active = panel_layout_.isResizeInteractionActive();
+        if (dock_resize_interaction_active != dock_resize_interaction_active_) {
+            viewer_->getRenderingManager()->setViewportResizeActive(dock_resize_interaction_active);
+            dock_resize_interaction_active_ = dock_resize_interaction_active;
+        }
+
         if (has_side_panel_plugins || has_floating_panels || has_status_bar_panels ||
             right_panel_requires_live_layout || bottom_dock_requires_live_layout ||
             left_dock_requires_live_layout ||
--- a/src/visualizer/gui/gui_manager.hpp
+++ b/src/visualizer/gui/gui_manager.hpp
@@ -419,6 +419,7 @@ namespace lfs::vis {
                 RightPanelPointerRegion::None;
             bool bottom_dock_pointer_live_capture_ = false;
             bool left_dock_pointer_live_capture_ = false;
+            bool dock_resize_interaction_active_ = false;
             std::string last_ui_layout_active_tab_;
             std::uint64_t last_pre_scene_panel_sync_generation_ = 0;
 
--- a/src/visualizer/gui/panel_layout.hpp
+++ b/src/visualizer/gui/panel_layout.hpp
@@ -111,6 +111,10 @@ namespace lfs::vis::gui {
                    left_dock_resizing_ || left_dock_hovering_edge_;
         }
 
+        bool isResizeInteractionActive() const {
+            return python_console_resizing_ || bottom_dock_resizing_ || left_dock_resizing_;
+        }
+
         CursorRequest getCursorRequest() const { return cursor_request_; }
 
         void applyResizeDelta(float dx, const ScreenState& screen);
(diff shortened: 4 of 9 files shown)


kmeirlaen:
both copilot suggestions applied

PR #1332, opened 2026-06-26
#1327 fix(sequencer): disable toolbar button while training is active

### LINKED ISSUE #1327: Sequencer panel closes on checkpoint load and won't stay open afterward
**Environment**
LichtFeld Studio v0.5.3
headless Docker, selkies-gstreamer (WebRTC), Xorg + openbox
2560×1440, Vulkan

**Summary**
With no checkpoint loaded, the Sequencer opens from the sidebar and stays. Once a checkpoint is loaded (i.e. training is active), it won't stay open: if already open it closes on load; clicking the sidebar button afterward flashes it open for one frame, then it closes again.

**Steps to reproduce**
1. Start LichtFeld Studio (no checkpoint).
2. Open the Sequencer — confirm it appears and stays.
3. Load a checkpoint from an earlier run.
4. Sequencer closes on load.
5. Click the Sequencer button — it flashes open for one frame and closes.

**Cause**
`SequencerPanel::poll()` (`src/visualizer/gui/native_panels.cpp`) gates the panel on `isToolsDisabled()`:

```cpp
const bool is_enabled = !ctx.ui_hidden && ctx.ui && ctx.ui->editor &&
                        !ctx.ui->editor->isToolsDisabled() && layout_->isShowSequencer();
```

`isToolsDisabled()` is true for editor modes `TRAINING`/`PAUSED`/`FINISHED`, so a loaded checkpoint makes `poll()` return false every frame. The toggle sets `show_sequencer_ = true` for one frame, the panel renders once, the next `poll()` closes it. That gate is right for the editing gizmos (translate/rotate/scale/select), but the sequencer is a camera/animation timeline, not a scene-editing tool — it looks swept in unintentionally.

**Expected**
A loaded checkpoint / active training shouldn't force the Sequencer closed; it should stay open when toggled.

**Possible fix**
Drop the `isToolsDisabled()` term from the sequencer's gate so it depends only on `isShowSequencer()`. Happy to open a PR. (Open question: should the sequencer be usable during *running* training, or only paused/finished?)

---

*This issue was created with the help of a coding agent.*

tritolol:
fixes #1327

## Summary

The Sequencer sidebar button stays fully active during training even though the
panel itself is disabled then (the native `SequencerPanel::poll()` gates on
`EditorContext::isToolsDisabled()`, true for `TRAINING`/`PAUSED`/`FINISHED`).
The button therefore looks live but does nothing on press, and clicking it could
leave the panel's visibility flag set while the panel stays gated off — the
cause of the "opens for one frame then closes" behavior after a checkpoint load.

## Change

Make the sequencer button reflect training state, the same way the editing-tool
buttons already do. Tool buttons render with
`enabled=tool_def.can_activate(context)`; the sequencer button is hardcoded and
never had an equivalent. This passes `enabled=` on it, computed from the same
`_TOOLBAR_HIDDEN_STATES` trainer-state set the toolbar already uses elsewhere, so
the button greys out during training. The `toggle_sequencer` dispatch also
no-ops in those states as a guard.

Single file, no C++ changes; the `poll()` gate is untouched.

## Testing

- Built from source (Release).
- During training (and with a checkpoint loaded): sequencer button is greyed out;
  clicking it does nothing; the one-frame-flash behavior no longer occurs.
- Before training / not gated: button active, sequencer opens and works normally.
- Editing tools unaffected.

### DIFF
--- a/src/python/lfs_plugins/toolbar.py
+++ b/src/python/lfs_plugins/toolbar.py
@@ -836,6 +836,11 @@ class _UtilityToolbarController:
         utility_bottom_buttons = []
         if has_render_manager:
             seq_visible = lf.ui.is_sequencer_visible()
+            # The sequencer is disabled while training is active (the native
+            # SequencerPanel gates on EditorContext::isToolsDisabled). Reflect
+            # that in the button so it greys out instead of appearing live but
+            # doing nothing on press, matching the editing-tool buttons.
+            seq_enabled = RuntimeState.trainer_state.value not in _TOOLBAR_HIDDEN_STATES
             utility_extra_buttons.append(
                 _button_record(
                     "util-sequencer",
@@ -845,6 +850,7 @@ class _UtilityToolbarController:
                     tooltip_key="toolbar.sequencer",
                     tooltip_text="Sequencer",
                     selected=seq_visible,
+                    enabled=seq_enabled,
                 )
             )
 
@@ -881,6 +887,8 @@ class _UtilityToolbarController:
             lf.focus_selection()
             return
         if action == "toggle_sequencer":
+            if RuntimeState.trainer_state.value in _TOOLBAR_HIDDEN_STATES:
+                return
             lf.ui.set_sequencer_visible(not lf.ui.is_sequencer_visible())
             return
         if action == "toggle_panel":


MrNeRF:
looks good to me. Thx for the fix!

PR #1343, opened 2026-06-28
Fix logging panel action layout

kmeirlaen:
The logging panel shows some bad positioning and size:

This fix now reserve stable button dimensions and disable wrapping for the controls in that row.

Before:
image

After:

image


kmeirlaen:
I checked overflowing suggestions - at max scale and minimum side panel width, buttons still show properly

image

PR #1344, opened 2026-06-28
COLMAP Text Import Performance

kmeirlaen:
## Problem

Loading large COLMAP text datasets was very slow in LichtFeld Studio. A text-only dataset with 986 cameras and 3,530,724 sparse points took about 147 seconds to import, with most of the time spent in COLMAP metadata and point cloud parsing before the scene became usable.

The issue was most visible for datasets that only contain:

- `cameras.txt`
- `images.txt`
- `points3D.txt`

We profiled 2 datasets for this improvement Results are:

Dataset 1:
```text
146876 ms -> 21715 ms
Saved: 125161 ms
Speedup: about 6.8x
Reduction: about 85%
```

Dataset 2:
```68149 ms -> 45758 ms -> 30312 ms -> 17172 ms -> 10604 ms```

## Observations

Initial logs showed two major bottlenecks:

- `images.txt` was parsed twice during a normal full import.
- Both image observations and point tracks were parsed and stored even though normal dataset import only needs camera metadata and point cloud `xyz/rgb`.

For the larger validation dataset `colmap_lcc_images`, the old import profile was:

```text
Read images.txt took 52614 ms
Read images.txt took 53351 ms
Read COLMAP cameras and images (text) took 53804 ms
Read points3D.txt took 40019 ms
Read COLMAP point cloud (text) took 40376 ms
COLMAP dataset loaded successfully in 146876 ms
```

Dataset consistency from the old run:

```text
Cameras: 986
Points: 3,530,724
Format: COLMAP text
```

## Analysis

The loader was paying for data that was not needed by the normal import path:

- `images.txt` contains two logical lines per image: camera/image metadata and a large 2D observation line.
- Normal import needs image id, pose, camera id, and image name, but it does not need `points2D`.
- `points3D.txt` contains `xyz/rgb/error` plus track pairs.
- Normal point cloud import needs only `xyz/rgb`, but it was parsing and storing all track elements.

The first optimization removed redundant work. Later optimizations separated the full reconstruction-preserving parse path from the fast normal-import path.

## Root Cause

The COLMAP text loader used one general-purpose parser for multiple use cases:

- validating/importing a dataset for viewing/training
- preserving COLMAP observations/tracks for reconstruction export

That general parser eagerly loaded observation and track data for all imports. For large reconstructions, this meant millions of unnecessary `points2D` and track entries were parsed, allocated, and copied. In addition, a top-level validation pass caused `images.txt` to be read before the actual import, duplicating one of the most expensive steps.

## Changes

The import path was split into fast import-only readers while preserving the existing full readers for export/reconstruction workflows.

Implemented changes:

- Removed the duplicate full validation parse from normal COLMAP loads.
- Kept validation behavior for `validate_only`.
- Added targeted `[COLMAP_LOAD]` timing logs for discovery, camera metadata, validation, alpha probing, point cloud loading, and centralization.
- Added `read_images_text_camera_metadata_only(...)` for normal imports.
- Made normal text image import skip 2D observation lines instead of parsing them.
- Added fast numeric parsing helpers using `std::from_chars`.
- Added a direct `points3D.txt` point-cloud reader that parses `xyz/rgb` straight into tensor source arrays.
- Avoided building intermediate `Point3DData` records for normal text point cloud import.
- Kept full observation/track-preserving text parsers available for COLMAP reconstruction/export paths.

### Iteration 1: Identify duplicate image parsing

The original log showed `images.txt` parsed twice. The first parse came from validation, then the actual import parsed the same file again.

Result on first test dataset:

```text
Before: 68149 ms
After removing duplicate validation parse: 45758 ms
```

### Iteration 2: Skip observations and tracks during normal import

The next logs showed that `images.txt` and `points3D.txt` still spent most time parsing unused data:

```text
images.txt parse: 17570 ms, points2D=5,261,907
points3D.txt parse: 12733 ms, track_elements=5,261,907
```

Normal import was changed to use `parse_points2D=false` and `parse_tracks=false`.

Result on first test dataset:

```text
45758 ms -> 30312 ms
```

### Iteration 3: Directly skip image observation lines

Even with `parse_points2D=false`, the image parser still probed large observation lines as possible metadata. The import path was changed to skip the documented COLMAP observation line directly.

Result after rebuild:

```text
images.txt parse dropped to about 4 ms
```

### Iteration 4: Fast point cloud text import

The point cloud path still parsed `points3D.txt` into `Point3DData` records, then copied the useful data into tensors. A direct point-cloud reader was added for normal import.

Result on first test dataset:

```text
30312 ms -> 17172 ms
```

### Iteration 5: Streaming camera metadata reader

The `images.txt` path still read the full file into strings even though only 437 or 986 metadata lines were needed. A streaming metadata-only reader was added.

Result on first test dataset:

```text
17172 ms -> 10604 ms
```

### Iteration 6: Preserve one-line `images.txt` compatibility

Copilot review identified that the direct observation-line skip assumed every COLMAP text image record had a following `points2D` line. That is the normal COLMAP layout, but some minimal text datasets and tests use one metadata line per image with no observation line.

The fast path now uses a bounded lookahead:

- If the next physical line is another image metadata record, keep it for the next loop iteration.
- If the next physical line is an observation line, skip it without storing the full line.
- Keep the full parsers unchanged for reconstruction/export workflows that need observations.

Latest run after this safety fix:

```text
camera_metadata_fast: 2233 ms
Read COLMAP cameras and images (text): 2655 ms
point_cloud_fast: 18783 ms
Read COLMAP point cloud (text): 18835 ms
COLMAP dataset loaded successfully in 21505 ms
```

This added about 0.5 s to camera metadata parsing versus the previous fast run, while keeping the full import time effectively unchanged because `points3D.txt` parsing remains dominant.

### Validation Dataset: `colmap_lcc_images`

Old import:

```text
COLMAP dataset loaded successfully in 146876 ms
986 cameras
3,530,724 points
```

New import:

```text
COLMAP dataset loaded successfully in 21715 ms
986 cameras
3,530,724 points
```

After the final streaming metadata change, the relevant phase timings were:

```text
camera_metadata_fast: 1706 ms
Read COLMAP cameras and images (text): 2106 ms
point_cloud_fast: 19539 ms
Read COLMAP point cloud (text): 19593 ms
COLMAP dataset loaded successfully in 21715 ms
```

Overall improvement on this dataset:

```text
146876 ms -> 21715 ms
Saved: 125161 ms
Speedup: about 6.8x
Reduction: about 85%
```

### First Test Dataset

The smaller initial test dataset improved across the investigation as follows:

```text
68149 ms -> 45758 ms -> 30312 ms -> 17172 ms -> 10604 ms
```

The final profile showed:

```text
camera_metadata_fast: 803 ms
Read COLMAP cameras and images (text): 1073 ms
point_cloud_fast: 9489 ms
COLMAP dataset loaded successfully in 10604 ms
```

## Remaining Bottleneck

After these changes, the remaining cost is almost entirely `points3D.txt` text parsing. This is expected for very large text point clouds:

```text
3,530,724 points
points3D.txt size: 380 MB
point_cloud_fast: about 19.5 s
```

## Compatibility Notes

The normal import path now avoids observations and tracks by default, but the full parsers remain available. This keeps the optimization compatible with future track-based filtering if track loading is made explicit, for example:

```text
TrackLoadMode::None
TrackLoadMode::TrackLengthOnly
TrackLoadMode::FullTracks
```

The important invariant is that normal import stays on the fast path, while workflows that need tracks opt into the extra parsing cost.

## Validation Behavior

The duplicate top-level validation parse was removed, but normal imports still validate the dataset layout. Validation now happens inside the camera/image read path after the fast metadata-only image reader has loaded the COLMAP image records.

Validation still covers the important data used by normal import:

- required COLMAP camera/image metadata files
- image folder existence
- missing image files referenced by COLMAP metadata
- ambiguous image basename references
- ambiguous mask/depth references
- missing `camera_id` references from images
- invalid or unsupported camera models during camera assembly
- malformed required camera/image/point fields that are consumed by the fast readers

Additional checks still happen during camera assembly, including image lookup fallback and camera id/model validation.

What changed intentionally:

- Normal import no longer parses `images.txt` `points2D` observations.
- Normal import no longer parses `points3D.txt` track pairs.
- Therefore normal import no longer validates corruption that exists only in unused observation or track data.

This is acceptable for viewing/training import because that data is not consumed. Workflows that need observations or tracks, such as reconstruction-preserving export or future track-based filtering, should opt into the full parser or a dedicated track-loading mode. In those modes, the existing full observation/track-preserving parsers remain available and can perform the stricter validation.

### DIFF
--- a/src/io/formats/colmap.cpp
+++ b/src/io/formats/colmap.cpp
@@ -8,8 +8,11 @@
 #include "core/path_utils.hpp"
 #include "io/filesystem_utils.hpp"
 #include <algorithm>
+#include <charconv>
 #include <cctype>
+#include <chrono>
 #include <cmath>
+#include <cstdint>
 #include <cstring>
 #include <exception>
 #include <filesystem>
@@ -19,6 +22,7 @@
 #include <memory>
 #include <sstream>
 #include <stdexcept>
+#include <system_error>
 #include <unordered_map>
 #include <vector>
 
@@ -38,6 +42,34 @@ namespace lfs::io {
 
         [[nodiscard]] bool should_poll_cancel(const size_t index) {
             return (index % CANCEL_POLL_INTERVAL) == 0;
+        }
+
+        [[nodiscard]] double elapsed_ms(const std::chrono::high_resolution_clock::time_point start) {
+            return std::chrono::duration<double, std::milli>(
+                       std::chrono::high_resolution_clock::now() - start)
+                .count();
+        }
+
+        void skip_ascii_spaces(const char*& cur, const char* end) {
+            while (cur < end && std::isspace(static_cast<unsigned char>(*cur))) {
+                ++cur;
+            }
+        }
+
+        template <typename T>
+        bool parse_next_number(const char*& cur, const char* end, T& value) {
+            skip_ascii_spaces(cur, end);
+            if (cur >= end) {
+                return false;
+            }
+
+            const auto parsed = std::from_chars(cur, end, value);
+            if (parsed.ec != std::errc{}) {
+                return false;
+            }
+
+            cur = parsed.ptr;
+            return true;
         }
     } // namespace
 
@@ -388,7 +420,11 @@ namespace lfs::io {
         const std::vector<ImageData>& images,
         const LoadOptions& options = {}) {
 
+        LOG_TIMER_DEBUG("COLMAP validate dataset layout");
         const fs::path images_path = base / lfs::core::utf8_to_path(images_folder);
+        LOG_INFO("[COLMAP_LOAD] validate_layout images={} images_path='{}'",
+                 images.size(),
+                 lfs::core::path_to_utf8(images_path));
         if (!safe_is_directory(images_path)) {
             return make_error(ErrorCode::PATH_NOT_FOUND, "Images folder does not exist", images_path);
         }
@@ -775,6 +811,9 @@ namespace lfs::io {
     std::vector<std::string> read_text_file(const std::filesystem::path& file_path,
                                             const LoadOptions& options = {}) {
         LOG_TRACE("Reading text file: {}", lfs::core::path_to_utf8(file_path));
+        const auto start = std::chrono::high_resolution_clock::now();
+        std::error_code file_size_ec;
+        const auto byte_size = fs::file_size(file_path, file_size_ec);
         std::ifstream file;
         if (!lfs::core::open_file_for_read(file_path, file)) {
             LOG_ERROR("Failed to open text file: {}", lfs::core::path_to_utf8(file_path));
@@ -806,6 +845,11 @@ namespace lfs::io {
             lines.pop_back();
 
         LOG_TRACE("Read {} lines from text file", lines.size());
+        LOG_INFO("[COLMAP_LOAD] read_text_file file='{}' bytes={} data_lines={} elapsed_ms={:.2f}",
+                 lfs::core::path_to_utf8(file_path),
+                 file_size_ec ? std::string("unknown") : std::format("{}", byte_size),
+                 lines.size(),
+                 elapsed_ms(start));
         return lines;
     }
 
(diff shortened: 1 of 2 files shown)


kmeirlaen:
Fixed copilot comments, not sure about the last copilot comment, the same is present in other locations of the code, fixing it everywhere would bypass the purpose of this PR.  might be good to look into though.

MrNeRF:
> Fixed copilot comments, not sure about the last copilot comment, the same is present in other locations of the code, fixing it everywhere would bypass the purpose of this PR.  might be good to look into though.

As long as it not popping up somewhere I would just leave it as is.

PR #1349, opened 2026-06-29
COLMAP Minimum Track Length Filter

kmeirlaen:
This change adds an optional sparse-point reliability filter for COLMAP imports. A COLMAP point track is the list of image observations that support a reconstructed 3D point. Higher track lengths generally indicate points that were seen from more views and are therefore more reliable.

## User-Facing Behavior

The filter is disabled by default.

Set a minimum track length greater than `0` to keep only COLMAP sparse points observed in at least that many images. For example, a value of `5` keeps points with track length `5` or higher and removes points with track length `1` through `4`.

Command-line usage:

```powershell
build\LichtFeld-Studio.exe --min-colmap-track-length 5 -d D:\temp\gs_tests\tutorial\colmap -o D:\temp\gs_tests\tutorial\output
```

The command-line flag uses hyphens:

```text
--min-colmap-track-length
```

The persisted/settings/API property uses underscores:

```text
min_colmap_track_length
```

## Load Dataset Panel

The Load Dataset panel now exposes a `Min COLMAP Track Length` field.

image

- `0` disables filtering.
- The input accepts typed values.
- The `+` and `-` buttons step the value up or down.
- The UI clamps the value to two digits, `0..99`.

## Logging

When the filter is applied, the COLMAP loader logs the configured threshold, the number of sparse points before filtering, and the number of points kept after filtering:

```text
COLMAP min track length filter: min track length 5, total points 1699644, after filtering 35893
```

This makes it visible how aggressively the setting reduced the sparse initialization point cloud.

## Import Path Details

The optimized unfiltered text import path is preserved.

- When `min_colmap_track_length == 0`, text `points3D.txt` loading continues to use the fast text reader.
- When `min_colmap_track_length > 0`, the loader reads COLMAP point records with track metadata and filters them before building the point cloud.
- Binary `points3D.bin` imports also support filtering.
- Existing `points3D.ply` files do not contain COLMAP track metadata. If filtering is enabled and `points3D.bin` or `points3D.txt` is available, the loader skips `points3D.ply` and reads the COLMAP file with tracks instead.
- If filtering is enabled but only `points3D.ply` exists, the loader falls back to loading the PLY unfiltered and emits a warning.

### DIFF
--- a/src/core/argument_parser.cpp
+++ b/src/core/argument_parser.cpp
@@ -257,7 +257,7 @@ namespace {
                                                                 {"4", 4},
                                                                 {"8", 8}});
             ::args::ValueFlag<int> max_width(dataset_group, "max_width", "Max width of images in px; 0 disables the cap (default: 3840)", {"max-width"});
-            ::args::ValueFlag<int> min_colmap_track_length(dataset_group, "min_colmap_track_length", "Minimum COLMAP point track length; 0 disables sparse point filtering", {"min-colmap-track-length"});
+            ::args::ValueFlag<int> min_track_length(dataset_group, "min_track_length", "Minimum point track length for COLMAP sparse point import; 0 disables filtering", {"min-track-length"});
             ::args::Flag no_cpu_cache(dataset_group, "no_cpu_cache", "Disable CPU memory caching (default: enabled)", {"no-cpu-cache"});
             ::args::Flag no_fs_cache(dataset_group, "no_fs_cache", "Disable filesystem caching (default: enabled)", {"no-fs-cache"});
             ::args::Flag undistort(dataset_group, "undistort", "Undistort images on-the-fly before training", {"undistort"});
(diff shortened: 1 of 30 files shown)


MrNeRF:
Shouldnt the flag be named min-track-length simply or min-track-length-colmap?


I think it should also inform the user if it is unused, no?

kmeirlaen:
@MrNeRF when its non-colmap data, the "colmap track filter" is not shown in the GUI.
the only edge case is that if there is a "points3D.ply" but no .bin or .txt - but i dont think anybody uses points3D.ply at this moment (could be wrong, just a hunch, as it seems an "undocumented" feature anywya)

and there is a "WARN" message as well when its not used.

kmeirlaen:
@MrNeRF no preference to rename the CLI flag - please suggest which one you prefer and i'll rename. 

right now it only applies to colmap data, but if other data formats surface with the same feature (track length), then a common name would make more sense. but now it is pretty clear it only applies to colmap data (.bin and .txt).

MrNeRF:
min-track-length and add to description that it only applies to the colmap format

kmeirlaen:
renamed - also logged warning when used in combination with --init-ply

MrNeRF (approved):

PR #1351, opened 2026-06-29
Various fixes: frustum picking, point cloud selection, file menu, toolbar, and crop tools (box + ellipsoid)

francescofugazzi:
# Summary

This pull request collects a series of fixes aimed at improving the overall stability and usability of the **Scene View**, **File Menu**, **Toolbar**, and **Crop Tools**.

## Scene View

- Fixed frustum picking so cameras are only selected when the click is within a reasonable hit radius, preventing accidental camera selection from arbitrary distances.
- Added support for selecting point cloud nodes using both ray picking and rectangle selection, making pre-training point clouds directly selectable in the viewport.

## File Menu

- Fixed an issue where the **New Project** confirmation dialog would not appear after importing data if the scene temporarily transitioned through an empty state.

## Toolbar

- Fixed disabled toolbar buttons not displaying reduced opacity. Since RmlUI's native `:disabled` selector does not apply to `<div>` elements, a custom `data-class-disabled` attribute is now used.
- When the scene becomes empty (for example after creating a **New Project**), the toolbar now automatically clears the active tool, preventing orphaned tool states.

## Crop Tool (Box & Ellipsoid)

This area received the largest set of fixes. Several issues prevented the ellipsoid crop from functioning correctly, while the box crop suffered from incorrect transformation logic and rendering inconsistencies.

### Selection

- Fixed the ellipsoid crop handler failing to locate splat nodes because it did not recognize all supported node types.
- Both crop handlers now correctly recognize the following node types:
  - `SPLAT`
  - `POINTCLOUD`
  - `CROPBOX`
  - `ELLIPSOID`

### Crop Direction

- Fixed the box crop transformation being computed in the wrong direction, which caused gaussians inside the crop box to be deleted instead of those outside.
- The world-to-local transform is now correctly inverted.

### GPU Stability

- Fixed a GPU hang during undo operations.
- The undo snapshot is now created only after verifying that there are valid splat nodes to process, preventing dangling GPU resources and application freezes.

### Renderer Refresh

- Fixed the renderer not updating after applying an ellipsoid crop.
- Deleted gaussians were correctly marked in memory but remained visible because the Vulkan deletion mask was never uploaded.
- The render dirty flag is now properly set before emitting the crop event, matching the box crop implementation.

### Overlay Rendering

- Fixed overlay ghosting when switching between box and ellipsoid crop modes.
- Overlay clear and update routines now correctly reset all overlay visibility flags.

### Scene Tree Synchronization

- The ellipsoid crop now emits the same `PLYAdded` event as the box crop, keeping the scene tree synchronized.

### Undo Support

- Both box and ellipsoid crop operations now register their changes in the undo history, allowing crop operations to be reverted correctly.


MrNeRF:
Thank you for all the fixes first and foremost! 
I am currently afk so cant test. I just can look at the code if the changes are plausible.

Let's test and review as is but would be great to have the issues described and how to trigger each and fixed in separate prs even if there are 10 pr or more.

src/visualizer/gui_capabilities.cpp
```diff
@@ -567,6 +568,27 @@ namespace lfs::vis::cap {
         if (filtered->editable_names.empty() || (policy == TransformTargetPolicy::RequireAllEditable && has_any_invalid))
             return std::unexpected(format_transform_target_error(*filtered, requested_node, policy));
 
+        // When a SPLAT or POINTCLOUD under a DATASET is transformed,
+        // also include the parent DATASET so cameras (siblings) move too.
```
MrNeRF: I think this should be reverted. It breaks how a scene graph works. Transforming the sibbling should not transform the parent. I admit it can be confusing but the relation is directional.

MrNeRF:
@francescofugazzi please see my comment. Even of that's a bit counterintuitive I would not break the scene graph.

francescofugazzi:
@MrNeRF done. I reverted that modification

MrNeRF:
@francescofugazzi 
Seems like this patch broke
 ```
Overlay Rendering
--> here i saw a regression - when switching tools, the crop box overlay dissapears, even if the crop is still active
```

Can you please check and fix it in a follow up commit.

francescofugazzi:
> @francescofugazzi Seems like this patch broke
> 
> ```
> Overlay Rendering
> --> here i saw a regression - when switching tools, the crop box overlay dissapears, even if the crop is still active
> ```
> 
> Can you please check and fix it in a follow up commit.

Maybe I did not get it properly. The cropboxes should always be visible also when choosing the other tools in the vertical toolbar? Just asking for confirmation.

MrNeRF:
I defer to @kmeirlaen as I can't test and compare. 

Maybe it is good to not show it at all when switching the tool. Instead the crop tool should be closed/deactivated.

kmeirlaen:
I raised it as a regression as before the implementation, this was handled differently.  I still think the whole crop process can be improved so dont mind if we handle this in later PRs.
Nice job overall though!  I just wish I could have tested more - most of the issues were a bit unclear for me on what the issue was + what was solved.

francescofugazzi:
> I raised it as a regression as before the implementation, this was handled differently. I still think the whole crop process can be improved so dont mind if we handle this in later PRs. Nice job overall though! I just wish I could have tested more - most of the issues were a bit unclear for me on what the issue was + what was solved.

@kmeirlaen thank you :)
about what I fixed was:
- new project wasn't triggering the confirmation dialog and overwriting the open project
- the toolbar with the instruments was displayed as active when it wasn't outside of edit mode now the buttons can be seen deactivated
- with a loaded dataset whenever I clicked in the viewport it was always selecting the same exact camera. Now it doesn't happen and you can properly click on cameras only at the rendered frustum, pcd etc
- i unlocked the use of translation/rotation tool even with dataset for correcting the position rotation before training (it's very  useful sometimes with some datasets orientated badly at soruce)
- the crop tools werent' working to me so i trieed to make them work

this is a quick list :)

PR #1355, opened 2026-06-29
Fix - PPISP Correction Has No Effect In VkSplat Viewport

### LINKED ISSUE #1353: PPISP Correction appears to have no effect in v0.5.3 (works correctly in v0.5.2)
Hi!

First of all, thank you for LichtFeld Studio. It has already become an extremely useful tool for 3DGS reconstruction.

I think I found a regression in v0.5.3 related to PPISP Correction.

Environment:
- Windows 10 x64
- RTX 3060 12 GB
- NVIDIA driver 572.47
- CUDA Toolkit 12.8
- Built from source, v0.5.3 tag
- Portable build

Issue:
In v0.5.3, PPISP Correction appears to have no visible effect on the rendered viewport image.

Changing the following settings does not visibly affect the viewport/render:
- PPISP Correction on/off
- Auto / Manual mode
- Exposure
- Gamma
- Color Correction

The same dataset in v0.5.2 works correctly:
- PPISP Correction visibly affects the image
- Exposure/color correction works during training and in Edit Mode
- The result is much more natural and less dark

Expected behavior:
PPISP Correction should affect the rendered viewport image as it does in v0.5.2.

Actual behavior:
The UI controls are present and editable, but the rendered viewport image does not seem to change.

Additional note:
I compared the source code between v0.5.2 and v0.5.3 and noticed substantial changes around the viewport rendering pipeline. I also noticed changes involving viewport_appearance_correction.hpp and applyViewportAppearanceCorrection(...), so I wonder if the regression may be related to that refactoring.

So my suspicion is that the PPISP settings are still exposed in the UI, but the correction is no longer applied to the main viewport image path.

I can reliably reproduce this every time by opening the same dataset in both v0.5.2 and v0.5.3.
The issue is reproducible with both a newly trained model and previously trained checkpoints.

v0.5.3 (PPISP appears to have no effect):
LichtFeld 0.5.3 - PPISP corrections no effect

v0.5.2 (same dataset, PPISP working correctly):
LichtFeld 0.5.2 - PPISP corrections working

Thanks!

kmeirlaen:
Reporting issue: #1353 

## Cause

The PPISP controls were still updating render settings and marking the viewport dirty, but the main VkSplat viewport path did not pass through the tensor image path where PPISP correction is applied.

In the working v0.5.2 path, PPISP-enabled rendering produced a tensor image, called `applyViewportAppearanceCorrection(...)`, and then uploaded/materialized the corrected image for display.

In v0.5.3 and the current Vulkan viewport path, VkSplat renders directly into an external Vulkan image and returns early through `publish_vksplat_result(...)`. That bypassed the later block:

```cpp
if (rendered_image && !rendered_image_contains_ground_truth) {
    rendered_image = applyViewportAppearanceCorrection(...);
}
```

Because `rendered_image` was null for the external-image VkSplat path, PPISP was never applied to the displayed main viewport image. The UI appeared functional, but exposure/gamma/color changes had no visual effect.

## Regression Point

Manual testing with a PPISP-enabled checkpoint narrowed the regression to:

```text
a6ec9950 vulkan point cloud renderer          GOOD
b4f4f094 Remove cuda renderer (#1234)         BAD
```

Confirmed working checkpoints before the regression:

```text
f1475d0f Vulkan - wip (#1170)                 GOOD
0908455e Fix VkSplat viewport alpha compositing GOOD
94079f5c 3dgut vulkan                         GOOD
5d64936c Selection tools, fixes, anc co to vulkan (#1194) GOOD
8a00e0a2 Depth composition mesh splat (#1197) GOOD
a6ec9950 vulkan point cloud renderer          GOOD
```

Confirmed bad checkpoint:

```text
b4f4f094 Remove cuda renderer (#1234)         BAD
```

This matches the code-level cause: `b4f4f094` removed the old CUDA/tensor renderer route. PPISP correction still existed, but the main displayed VkSplat viewport image now came from the external Vulkan-image path, which bypassed `applyViewportAppearanceCorrection(...)`.

## Change

`src/visualizer/rendering/rendering_manager_vulkan.cpp` now handles PPISP inside the VkSplat publish path and the GT comparison rendered-panel path.

When `settings_.apply_appearance_correction` is enabled:

1. Read the just-rendered VkSplat output image back into a tensor.
2. Apply the existing `applyViewportAppearanceCorrection(...)` helper.
3. Publish the corrected tensor image through the normal tensor viewport path.
4. Clear the external-image fields so the GUI does not treat the corrected tensor as a Vulkan-external image.

When PPISP is disabled, the existing fast external Vulkan image path is unchanged.

The shared mesh/depth frame publishing logic was factored into a small local helper so both the corrected tensor path and the original external-image path keep overlays, mesh depth, and environment background behavior consistent.

For GT comparison, the rendered comparison panel can also arrive as a VkSplat external image in `OutputSlot::SplitRight`. When PPISP is enabled, that panel is now read back, corrected with `applyViewportAppearanceCorrection(...)`, and converted to a tensor-backed panel before split-view composition. This keeps the ground-truth photo unchanged while applying PPISP to the rendered side.

Generic split view and PLY comparison are intentionally left out of scope because split view was not a working PPISP baseline in v0.5.2.

## Validation

Manual test with a checkpoint/sidecar that contains PPISP data:

- v0.5.2 visibly responds to PPISP controls, confirming the asset is valid.
- Unpatched v0.5.3/current path does not visibly respond.
- Patched current build visibly responds in the main single viewport:
  - PPISP Correction on/off changes the image.
  - Manual Exposure changes the image.
  - Gamma/color controls affect the image.
- GT comparison rendered side is handled through the same PPISP correction path when the rendered panel is backed by a VkSplat external image.

An initial test showed:

```text
Vulkan viewport scene image upload requires an external CUDA/Vulkan image; none supplied
```

That happened because the corrected tensor return still forwarded the VkSplat completion semaphore, causing the GUI uploader to expect an external Vulkan image. The corrected tensor return no longer forwards that semaphore/value.

Regression identification test:

- `a6ec9950` responds to PPISP controls.
- `b4f4f094` does not respond to PPISP controls.
- Therefore the first bad commit is `b4f4f094 Remove cuda renderer (#1234)`.

## Notes

This is a correctness-first fix for the main viewport and GT comparison rendered panel. While PPISP is enabled, affected VkSplat outputs are read back to tensors before applying PPISP, so they may be slower than the pure external-image path. The non-PPISP path remains unchanged.

Independent split view and PLY comparison are not fixed here because they were also not working with PPISP in v0.5.2.

Closes #1353

### DIFF
--- a/src/visualizer/rendering/rendering_manager_vulkan.cpp
+++ b/src/visualizer/rendering/rendering_manager_vulkan.cpp
@@ -1464,7 +1464,32 @@ namespace lfs::vis {
                             }
 
                             if (compare_panel.image || compare_panel.external_image_view != VK_NULL_HANDLE) {
-                                if (compare_panel.image) {
+                                bool compare_panel_ppisp_applied = false;
+                                if (!compare_panel.image &&
+                                    compare_panel.external_image_view != VK_NULL_HANDLE &&
+                                    settings_.apply_appearance_correction &&
+                                    vksplat_viewport_renderer_ &&
+                                    context.vulkan_context) {
+                                    auto image = vksplat_viewport_renderer_->readOutputImage(
+                                        *context.vulkan_context,
+                                        VksplatViewportRenderer::OutputSlot::SplitRight);
+                                    if (image && *image) {
+                                        compare_panel.image = applyViewportAppearanceCorrection(
+                                            std::move(*image),
+                                            scene_manager,
+                                            settings_,
+                                            camera->uid());
+                                        if (compare_panel.image && compare_panel.image->is_valid()) {
+                                            compare_panel.external_image_view = VK_NULL_HANDLE;
+                                            compare_panel.external_image_generation = 0;
+                                            compare_panel_ppisp_applied = true;
+                                        }
+                                    } else {
+                                        LOG_WARN("VkSplat GT comparison PPISP readback failed: {}",
+                                                 image ? "missing image payload" : image.error());
+                                    }
+                                }
+                                if (compare_panel.image && !compare_panel_ppisp_applied) {
                                     compare_panel.image = applyViewportAppearanceCorrection(
                                         std::move(compare_panel.image),
                                         scene_manager,
@@ -2396,8 +2421,14 @@ namespace lfs::vis {
             if (result.image_generation == 0) {
                 result.image_generation = ++split_view_image_generation_;
             }
-            result.completion_semaphore = latest_vksplat_completion_semaphore;
-            result.completion_value = latest_vksplat_completion_value;
+            const bool split_uses_external_image =
+                pending_split_view.enabled &&
+                (pending_split_view.left.external_image_view != VK_NULL_HANDLE ||
+                 pending_split_view.right.external_image_view != VK_NULL_HANDLE);
+            if (split_uses_external_image) {
+                result.completion_semaphore = latest_vksplat_completion_semaphore;
+                result.completion_value = latest_vksplat_completion_value;
+            }
             result.size = vulkan_viewport_image_size_;
             result.flip_y = vulkan_viewport_image_flip_y_;
 
--- a/src/visualizer/rendering/rendering_manager_vulkan.cpp
+++ b/src/visualizer/rendering/rendering_manager_vulkan.cpp
@@ -1092,6 +1092,23 @@ namespace lfs::vis {
             bool flip_y = false;
         };
 
+        const auto ensure_cuda_viewport_image =
+            [](std::shared_ptr<lfs::core::Tensor> image,
+               const std::string_view label) -> std::shared_ptr<lfs::core::Tensor> {
+            if (!image || !image->is_valid()) {
+                return {};
+            }
+            if (image->device() == lfs::core::Device::CUDA) {
+                return image;
+            }
+            auto cuda_image = image->cuda();
+            if (!cuda_image.is_valid() || cuda_image.device() != lfs::core::Device::CUDA) {
+                LOG_WARN("{} produced a non-CUDA tensor; falling back to the external image path", label);
+                return {};
+            }
+            return std::make_shared<lfs::core::Tensor>(std::move(cuda_image));
+        };
+
         VkSemaphore latest_vksplat_completion_semaphore = VK_NULL_HANDLE;
         std::uint64_t latest_vksplat_completion_value = 0;
         bool vksplat_inputs_forced_this_frame = false;
@@ -1479,6 +1496,9 @@ namespace lfs::vis {
                                             scene_manager,
                                             settings_,
                                             camera->uid());
+                                        compare_panel.image = ensure_cuda_viewport_image(
+                                            std::move(compare_panel.image),
+                                            "VkSplat GT comparison PPISP correction");
                                         if (compare_panel.image && compare_panel.image->is_valid()) {
                                             compare_panel.external_image_view = VK_NULL_HANDLE;
                                             compare_panel.external_image_generation = 0;
@@ -2123,6 +2143,9 @@ namespace lfs::vis {
                                     scene_manager,
                                     settings_,
                                     frame_ctx.current_camera_id);
+                                corrected_image = ensure_cuda_viewport_image(
+                                    std::move(corrected_image),
+                                    "VkSplat viewport PPISP correction");
                                 if (corrected_image && corrected_image->is_valid()) {
                                     vulkan_viewport_image_ = corrected_image;
                                     ++vulkan_viewport_image_generation_;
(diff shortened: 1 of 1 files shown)


kmeirlaen:
The Vulkan viewport and split-view upload paths require CUDA-backed tensors for CUDA/Vulkan interop. VkSplat readback produces CPU tensors, so the corrected readback image is explicitly converted to CUDA before being published. If a corrected image cannot be made CUDA-backed, the code leaves the original external-image path in place instead of publishing an unusable CPU tensor.

Copilot review also flagged that VkSplat readback returns CPU tensors while Vulkan interop requires CUDA tensors. The patch now converts corrected PPISP tensors to CUDA before publishing them, and falls back to the original external image path if that contract cannot be met.

kmeirlaen:
I tested this path manually and the suggested inversion makes the VkSplat PPISP viewport render upside down.

The important detail is that this tensor does not come from the normal tensor renderer path. It comes from `VksplatViewportRenderer::readOutputImage()`, which reads back the already-presented VkSplat external image. So after PPISP correction, the tensor still has the same orientation convention as the VkSplat external image, not the raster/auxiliary tensor renderer convention.

Because of that, the corrected VkSplat path must keep `render_result.flip_y` / `metadata.flip_y` instead of switching to `!metadata.flip_y`.

I kept the CUDA-backed tensor guard from the earlier review because that part is valid: Vulkan interop still requires a CUDA tensor, otherwise we fall back to the original external image path.

Same reasoning applies to the GT comparison panel: this panel is converted from the VkSplat external split-right output via `readOutputImage()`, so its orientation follows the VkSplat external-image convention. Using `!metadata.flip_y` flips the panel incorrectly; the corrected tensor should keep `compare_panel.metadata.flip_y`.

src/visualizer/rendering/rendering_manager_vulkan.cpp
```diff
@@ -1464,7 +1481,36 @@ namespace lfs::vis {
...
+                                bool compare_panel_ppisp_applied = false;
+                                if (!compare_panel.image &&
+                                    compare_panel.external_image_view != VK_NULL_HANDLE &&
+                                    settings_.apply_appearance_correction &&
+                                    vksplat_viewport_renderer_ &&
+                                    context.vulkan_context) {
+                                    auto image = vksplat_viewport_renderer_->readOutputImage(
+                                        *context.vulkan_context,
+                                        VksplatViewportRenderer::OutputSlot::SplitRight);
+                                    if (image && *image) {
```
MrNeRF, line 1493: Image is a pointer here, right? Don't need to dereference in that second part of the if statement. But I can't check the signature easily here on mobile.

MrNeRF:
Just one comment @kmeirlaen

src/visualizer/rendering/rendering_manager_vulkan.cpp
kmeirlaen: image is not the tensor pointer directly. It is an std::expected that contains a std::shared_ptr<Tensor>

`std::expected<std::shared_ptr<lfs::core::Tensor>, std::string>`

So there are two layers:
```
image      // did readOutputImage succeed?
*image     // the shared_ptr<Tensor> inside the expected
```
So `image && *image` checks both that the expected has a value and that the contained shared_ptr is non-null.

if we would only use `if (image)`, that would only check that the function succeeded. It would not check whether the contained shared_ptr is non-null.

That said, seems that in practice, readOutputImage() currently always returns a non-null make_shared on success, so *image is defensive. But it is not wrong...

MrNeRF: Thanks for clarification.

luxoid1:
Hi!

I rebuilt the latest master after PR #1355 was merged and tested it on the same dataset that originally exposed the regression.

I can confirm that the issue is fixed.

PPISP Correction now works again as expected:

PPISP on/off visibly affects the rendered image.
Exposure, Gamma and Color Correction all work correctly.
The rendering now matches the behavior of v0.5.2.

Thank you very much for the incredibly fast fix!

PPISP enabled:
PPISP enabled

PPISP disabled:
PPISP disabled


While testing the latest build, I noticed another small rendering issue.

When deleting selected splats in Ring Selection mode using the Delete key, the viewport briefly flashes black for a single frame before rendering normally again.

This happens consistently on my system and does not affect functionality, but it looks like the framebuffer or viewport is being recreated for one frame.

MrNeRF:
@luxoid1 Awesome. Can you please create another issue to track it?

luxoid1:
Yes, of course! I'll create a new issue shortly.
Thanks a lot for the quick fix and for all your hard work!

PR #1356, opened 2026-06-30
fix(l10n): sync missing keys, reorder, add --lang filter to validate_localization

LeoYiMon:
## Summary

The `--fix` flag in `scripts/validate_localization.py` `line 178` was broken:

```python
if fix:
  print(f"  {Colors.YELLOW}→ Adding missing keys with TODO markers...{Colors.END}")
  # Add missing keys
  for key in missing:
      locale_data['flat'][key] = f"TODO: {locales['en']['flat'][key]}"
````

It modified a flattened in-memory dict and never wrote the result back to disk.  
This PR replaces the flat comparison with recursive `OrderedDict` sync, adds key reordering to match en.json, and introduces a `--lang` filter for scoped validation.

### Bug fixes
- `--fix` now actually writes synced data back to the target JSON file
- "unused keys" count now combines `string_keys.hpp` constants  
  with raw `LOC("...")` calls instead of counting only the latter

### New features
- `sync_dict()` recursively aligns a target locale file with en.json:
  fills missing keys from the English baseline and reorders all keys
  to the canonical en.json order
- `--lang zh,ja,fr` flag to scope validation/fix to one or more
  languages; unknown codes produce a hard error

## Usage

```bash
python scripts/validate_localization.py                         # Check all
python scripts/validate_localization.py --fix                   # Fix all
python scripts/validate_localization.py --lang zh --fix         # Fix only zh
```

image

### DIFF
--- a/scripts/validate_localization.py
+++ b/scripts/validate_localization.py
@@ -9,19 +9,25 @@ Features:
 - Finds missing translations
 - Finds unused translations
 - Validates JSON syntax
-- Compares with string_keys.hpp for consistency
+- Compares with string_keys.hpp and C++ LOC() calls
+- --fix: auto-adds missing keys and reorders to match en.json key order
+- --lang: scope to specific language(s)
 
 Usage:
-    python scripts/validate_localization.py
-    python scripts/validate_localization.py --fix  # Auto-add missing keys with TODO markers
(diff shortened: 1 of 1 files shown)


MrNeRF:
Great! Thank you!

PR #1361, opened 2026-07-02
VkSplat Training Viewport Stability

kmeirlaen:
## Summary

This change fixes an intermittent CUDA illegal-address failure seen on a 2080 Ti when training starts while the VkSplat viewport is active, especially when dragging or moving the viewport around training startup.

The fix keeps hardware decoding and the Vulkan viewport enabled. It does not exclude specific GPUs, disable nvImageCodec, or increase the VkSplat shared-scratch allocation estimate.

## Root Cause

The failure was initially reported as nvImageCodec decode failures, but the detailed logs showed those errors were often downstream of an already-poisoned CUDA context. The first CUDA illegal address usually came from FastGS rasterization or related training work, while nvImageCodec later observed the same CUDA error during decode or sync.

The useful signal came from the Vulkan/VkSplat path:

- VkSplat creates/imports a shared rasterizer scratch arena for viewport rendering.
- Training also uses the CUDA rasterizer arena.
- At training start, the render model pointer changes from the pre-training model to the training model.
- The old render path treated that model change as a full scene switch and reset `VksplatViewportRenderer`.
- That reset dropped the renderer state and shared-scratch import right as the training thread was starting.
- The first interactive viewport frames then tried to recreate or bind VkSplat shared scratch while FastGS training work was active.
- On the affected setup this could corrupt CUDA state and surface as `cudaErrorIllegalAddress` in FastGS, tensor ops, nvImageCodec, or Vulkan interop.

So the core issue was not "hardware decode is bad" and not "the 2080 Ti must be disabled". The crash came from unsafe lifecycle/ownership around the VkSplat shared scratch during training startup and interactive viewport work.

### Preserve VkSplat Renderer During Training Model Swap

When `renderVulkanFrame()` detects a model pointer change while training with the VkSplat backend, it now preserves the existing `VksplatViewportRenderer` instead of resetting it.

This keeps the already-created handshake and shared-scratch state alive across the normal training-start model transition.

For non-training model changes, the previous reset behavior is kept.

### Prime VkSplat Shared Scratch Before Training Starts

Training startup now asks the rendering manager to prepare VkSplat training shared scratch before the training thread begins.

This happens after trainer initialization and memory-pool trimming, but before `training_complete_` is cleared and before the training thread starts. The goal is to have the renderer's shared scratch installed before FastGS starts using the CUDA arena.

### Fail Viewport Contention Safely

During training renders, VkSplat now checks that shared scratch is already installed and tries to acquire the rasterizer arena guard before doing input upload or other CUDA/Vulkan render preparation.

If the arena is busy, the viewport returns the existing retryable "shared scratch unavailable" path. The rendering manager can keep or return the cached viewport image and retry later, instead of performing late scratch setup while training owns the arena.

### Prevent Camera Pause Lease Stacking

A second issue appeared after the crash was fixed: training sometimes did not resume after viewport/camera movement.

Logs showed repeated temporary camera pause requests at the same iteration while the underlying trainer was already paused. Those requests stacked temporary pause leases, so a later camera timeout could log "resuming" without actually reaching a zero pause depth.

Camera movement now only requests a temporary training pause when the trainer manager is running and the underlying trainer is not already paused.

## Files Changed

- `src/visualizer/rendering/rendering_manager.hpp`
- `src/visualizer/rendering/rendering_manager_vulkan.cpp`
- `src/visualizer/rendering/vksplat_viewport_renderer.hpp`
- `src/visualizer/rendering/vksplat_viewport_renderer.cpp`
- `src/visualizer/training/training_manager.hpp`
- `src/visualizer/training/training_manager.cpp`
- `src/visualizer/input/input_controller.cpp`

## Validation

Tested manually on the affected Windows/2080 Ti setup with:

```bat
build\LichtFeld-Studio.exe -o D:\temp\gs_tests\dual_camera\colmap\output -d D:\temp\gs_tests\dual_camera\colmap --log-file output.txt
```

Observed results after the patch:

- No CUDA illegal-address crashes while dragging/moving the viewport during training startup.
- nvImageCodec remains enabled.
- Vulkan viewport remains enabled.
- Training resumes correctly after camera/viewport movement.
- No GPU-specific exclusion was added.
- No larger shared-scratch capacity multiplier was kept.

Build validation:

```bat
cmake --build build -j 8
```

The build completed successfully.

## Notes For Reviewers

The key behavioral change is that training-start model pointer changes are not treated the same as ordinary scene switches for VkSplat renderer lifetime. In training, preserving the renderer is intentional because the renderer owns CUDA/Vulkan interop state that must remain stable across the handoff into live training.

The retry path for busy shared scratch is also intentional. A missed viewport frame is acceptable; racing the trainer for shared rasterizer arena ownership is not.

### DIFF
--- a/src/visualizer/rendering/vksplat_viewport_renderer.cpp
+++ b/src/visualizer/rendering/vksplat_viewport_renderer.cpp
@@ -6607,19 +6607,6 @@ namespace lfs::vis {
             splat_data.lod_tree->rad_source.valid();
         static const bool kDisableSharedScratch = (std::getenv("LFS_NO_SHARED_SCRATCH") != nullptr);
         std::optional<RasterizerArenaRenderGuard> shared_arena_guard;
-        if (synchronize_input_upload && !kDisableSharedScratch) {
-            if (!shared_scratch_.installed_in_training_arena || !shared_scratch_.block) {
-                return std::unexpected(
-                    "VkSplat shared scratch is not installed for training render (prime shared scratch before training starts)");
-            }
-            try {
-                shared_arena_guard.emplace();
-            } catch (const std::exception& e) {
-                return std::unexpected(std::format(
-                    "VkSplat shared scratch training rasterizer arena is busy; reason={}",
-                    e.what()));
-            }
-        }
         std::vector<LodPageCache::PendingUpload> lod_page_uploads;
         std::vector<std::uint32_t> protected_lod_chunks;
         bool lod_page_inputs_active = false;
--- a/src/visualizer/rendering/vksplat_viewport_renderer.cpp
+++ b/src/visualizer/rendering/vksplat_viewport_renderer.cpp
@@ -3170,10 +3170,16 @@ namespace lfs::vis {
 
         const std::size_t width = static_cast<std::size_t>(viewport_size.x);
         const std::size_t height = static_cast<std::size_t>(viewport_size.y);
+        if (height != 0 && width > (std::numeric_limits<std::size_t>::max() / height)) {
+            return std::unexpected("VkSplat training shared-scratch prime viewport size overflows size_t");
+        }
         const std::size_t num_pixels = width * height;
-        const std::size_t num_tiles =
-            ((width + TILE_WIDTH - 1) / TILE_WIDTH) *
-            ((height + TILE_HEIGHT - 1) / TILE_HEIGHT);
+        const std::size_t tiles_x = (width + TILE_WIDTH - 1) / TILE_WIDTH;
+        const std::size_t tiles_y = (height + TILE_HEIGHT - 1) / TILE_HEIGHT;
+        if (tiles_y != 0 && tiles_x > (std::numeric_limits<std::size_t>::max() / tiles_y)) {
+            return std::unexpected("VkSplat training shared-scratch prime tile count overflows size_t");
+        }
+        const std::size_t num_tiles = tiles_x * tiles_y;
         const std::size_t sort_capacity =
             num_splats > (std::numeric_limits<std::size_t>::max() / 4u)
                 ? num_splats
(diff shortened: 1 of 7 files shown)


src/visualizer/rendering/vksplat_viewport_renderer.cpp
```diff
@@ -7035,7 +7087,9 @@ namespace lfs::vis {
             shared_scratch_attempt_id = ++shared_scratch_attempt_serial_;
             if (auto ok = ensureSharedScratchArena(context, required_shared_scratch); ok) {
```
MrNeRF, line 7075: this will fail because line 6609 creates the arena guard which prevents the arena from growing.

MrNeRF: Currently this will only break on massive resize of the viewport or in case we had just a huge densification.  

I think this order must be ensured in the code:

```
ensureSharedScratchArena(...);  // may grow, requires no active arena frame
shared_arena_guard.emplace()
```

src/visualizer/training/training_manager.hpp
```diff
@@ -90,6 +90,7 @@ namespace lfs::vis {
         [[nodiscard]] bool isPaused() const { return state_machine_.isInState(TrainingState::Paused); }
         [[nodiscard]] bool isFinished() const { return state_machine_.isInState(TrainingState::Finished); }
         [[nodiscard]] bool isTrainingActive() const { return state_machine_.isActive(); }
+        [[nodiscard]] bool isTrainerPaused() const;
```
MrNeRF: I think this is confusing here and does not fit properly into the list of methods:

How about something like:
```
  bool TrainerManager::pauseTrainingTemporaryIfActive() {
      if (!isRunning() || !trainer_ || trainer_->is_paused()) {
          return false;
      }

      pauseTrainingTemporary();
      return true;
  }
```

MrNeRF:
thanks! Looks good.

PR #1364, opened 2026-07-02
Fix Copy/Paste And Selection Of Soft-Deleted Splats

kmeirlaen:
## Problem

Soft-deleted splats could still be included in Gaussian-level selection workflows. After deleting splats with the edit tools, the rows remained in `SplatData` behind a `_deleted` mask, but some selection and clipboard paths continued to operate on raw row indices without excluding those deleted rows.

This made invisible/deleted splats reappear or move through later user actions:

- `Ctrl-A`, then copy/paste, could paste splats that had already been soft-deleted.
- Brush/rectangle/lasso selections over regions containing deleted splats could include invisible rows.
- Cut inherited the same issue because it copies before deleting.
- Transform/mirror operations could affect stale selections that still overlapped deleted rows.

## Replication steps:
- load a splat
- select part of the splat using the (brush) selection tool and delete the selected splats (splats are deleted - ok)
- perform a ctrl-A to select all remaining splats (remaining splats are selected - ok)
- press ctrl-C (copy selected splats) and ctrl-V (paste selected splats)
- result: all splats (also the previously deleted splats) are pasted into a new splat (nok - only the remaining splats should be copied)

this also happens in the following scenario:
- load a splat
- select part of the splat using the (brush) selection tool and delete the selected splats (splats are deleted - ok)
- select a second part with brush - make sure to stroke splats that were previously present but now deleted
- copy / paste: result: not only the last selected splats are pasted, but also plats that were deleted

## Root Cause

Soft deletion is represented as data, not immediate row removal. The deleted rows remain in the model until compaction/export/apply-deleted style paths physically remove them.

Most rendering paths correctly treat `_deleted` as hidden, but selection masks and copy masks are still masks over model rows. If those masks are not intersected with `!deleted`, deleted rows can continue to flow through:

- `copySelectedGaussians()` extracts selected rows using `extract_by_mask()`.
- `SelectionService::commitSelection()` funnels most selection tools through `applyFilters()`.
- Some `SceneManager` fallback selection commands build full-scene masks directly.

The bug was not in `extract_by_mask()` itself. That helper intentionally extracts raw rows and is used by other systems that need raw-row semantics. The missing behavior was at the user-facing selection/copy boundaries.

## Chosen Solution

The fix treats deleted splats as unselectable and uncopyable at the command boundaries, while leaving raw extraction semantics unchanged.

### 1. Filter Deleted Splats During Copy

`SceneManager::copySelectedGaussians()` now intersects each selected chunk mask with the source model's deleted mask:

```cpp
mask = mask.logical_and(src.deleted().logical_not().to(mask.device()));
```

This applies to both consolidated scenes and per-node scenes because both branches pass through the same `add_selected_chunk` lambda. `cutSelectedGaussians()` also inherits the fix because it calls `copySelectedGaussians()` before deleting.

If everything selected was already deleted, the chunk becomes empty and the copy becomes a no-op instead of repopulating the clipboard with invisible rows.

### 2. Make Deleted Splats Unselectable

The central selection filtering path should remove deleted rows unconditionally in `SelectionService::applyFilters()`. That covers brush, rectangle, lasso, sphere, `Ctrl-A` through filtered selection, invert filtered selection, and direct mask application because they all flow through `commitSelection()` and `applyFilters()`.

The deleted-mask filter is guarded by a size check so scene-wide masks are not incorrectly combined with render-model masks in mismatched multi-node cases.

### 3. Cover SceneManager Fallback Selection Paths

Some fallback paths in `SceneManager` build selection masks directly instead of going through `SelectionService`:

- `invertSelection()`
- `selectAllGaussians()` fallback behavior

Those masks should also be intersected with `!deleted`, using consolidated-model deleted state when available and per-node deleted masks for non-consolidated scenes.

## Why This Approach

This keeps the behavior local to user-facing selection and clipboard operations:

- It does not change `extract_by_mask()`, preserving existing raw-row callers such as trainer/export paths.
- It handles copy, cut, selection, and transforms from stale masks without requiring immediate physical compaction.
- It works for both consolidated multi-splat scenes and single/per-node splat scenes.
- It keeps undo semantics intact: undoing a delete restores the deleted flags and the selection snapshot together, so restored splats can be selected and copied again.

## Verification Plan

- Load a scene, brush-delete part of it, then `Ctrl-A`, copy, and paste. The pasted node should contain only visible splats.
- Delete part of a scene, brush over an area overlapping deleted splats, then copy/paste. Only visible brushed splats should paste.
- After deletions, `Ctrl-A` should select the visible splat count, not the total row count.
- `Ctrl-X` followed by paste should behave the same as copy/paste.
- Undo the delete, then `Ctrl-A` and copy again. Restored splats should copy normally.
- Transform or mirror a selection near deleted regions, then undo the delete. Restored splats should remain in their original positions.
- Verify both consolidated multi-splat scenes and single/per-node splat scenes.

### DIFF
--- a/src/core/include/core/scene.hpp
+++ b/src/core/include/core/scene.hpp
@@ -469,6 +469,13 @@ namespace lfs::core {
         void updateWorldTransform(const SceneNode& node) const;
         void removeNodeInternal(const std::string& name, bool keep_children, bool force);
         [[nodiscard]] size_t currentSelectionCapacity() const;
+        [[nodiscard]] lfs::core::Tensor liveSelectionMask(size_t expected_size,
+                                                          Device device,
+                                                          DataType dtype) const;
+        [[nodiscard]] std::shared_ptr<lfs::core::Tensor> normalizeSelectionMask(
+            std::shared_ptr<lfs::core::Tensor> mask,
+            size_t expected_size,
+            size_t* selected_count = nullptr) const;
         void resizeSelectionIfSizeMismatch(size_t expected_size);
 
         SelectionGroup* findGroup(uint8_t id);
--- a/src/core/scene.cpp
+++ b/src/core/scene.cpp
@@ -952,6 +952,88 @@ namespace lfs::core {
         return getSelectionGaussianCount();
     }
 
+    lfs::core::Tensor Scene::liveSelectionMask(const size_t expected_size,
+                                               const Device device,
+                                               const DataType dtype) const {
+        Tensor live = Tensor::ones({expected_size}, device, dtype);
+        if (expected_size == 0) {
+            return live;
+        }
+
+        if (consolidated_) {
+            const auto* combined = getCombinedModel();
+            if (combined &&
+                combined->has_deleted_mask() &&
+                combined->deleted().numel() == expected_size) {
+                return combined->deleted().logical_not().to(device).to(dtype);
+            }
+        }
+
+        size_t offset = 0;
+        for (const auto& node : nodes_) {
+            if (node->type != NodeType::SPLAT) {
+                continue;
+            }
+
+            const size_t node_size = node->model
+                                         ? static_cast<size_t>(node->model->size())
+                                         : node->gaussian_count.load(std::memory_order_acquire);
+            const size_t node_end = offset + node_size;
+            if (node_end > expected_size) {
+                break;
+            }
+
+            if (node->model &&
+                node->model->has_deleted_mask() &&
+                node->model->deleted().numel() == node_size) {
+                live.slice(0, offset, node_end) = node->model->deleted().logical_not().to(device).to(dtype);
+            }
+
+            offset = node_end;
+        }
+
+        return live;
+    }
+
+    std::shared_ptr<lfs::core::Tensor> Scene::normalizeSelectionMask(
+        std::shared_ptr<lfs::core::Tensor> mask,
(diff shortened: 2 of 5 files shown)


kmeirlaen:
Copilot suggestions implemented

MrNeRF:
have you looked into the selection groups if they are affected btw the pr?

kmeirlaen:
not directly, but I did a review now.  Although I dont see how these selection groups are supposed to be used.

image

For selection groups it used be be:
- select splats gropu A (red) - group showed counter (X)
- select other splats group B (green) - group showed counter (Y)
- select other splats group C (blue) - group showed counter (Z - overlapping splats are removed from X and Y)
- delete splats - all splats from all are deleted (even though only group C is active), all groups showed counter (0)
- another stroke where no splats are supposed to be : splats count is updated for the active group (though nothing is visible)

Now:
- select splats gropu A (red) - group showed counter (X)
- select other splats group B (green) - group showed counter (Y)
- select other splats group C (blue) - group showed counter (Z - overlapping splats are removed from X and Y)
- delete splats - all splats from all are deleted (even though only group C is active), all groups showed counter (0)
- NEW: another stroke where no splats are supposed to be : counter still (0) - no splats selected

MrNeRF:
Ok, that seems alright. Thank you!

PR #1366, opened 2026-07-03
Optimize COLMAP `points3D.txt` Loading

kmeirlaen:
## Summary

This change significantly speeds up COLMAP `points3D.txt` loading for both the normal point-cloud path and the `--min-track-length` filtering path.

The implementation was split into three independently measurable phases:

1. Bulk-read `points3D.txt` and scan lines in-place instead of using `std::getline`.
2. Add track-count-only parsing so filtered loads do not materialize full track pairs.
3. Parallelize large `points3D.txt` parsing with newline-aligned chunks using TBB.

## Main Changes

- Replaced line-by-line stream parsing with a whole-file buffer and `std::string_view` line scanning.
- Optimized ASCII whitespace handling by checking only spaces and tabs in hot parse loops.
- Parse only the fields required by the fast point-cloud load path.
- Parse point positions directly into `float` for the normal load path.
- Added `track_count` to point records so filtering can avoid full track materialization.
- Added internal track parse modes:
  - `None`
  - `CountOnly`
  - `Full`
- Reworked text record parsing to use `from_chars` instead of `split_string` / `std::stod`.
- Updated min-track filtering to use `track_count`.
- Added parallel chunk parsing for large files:
  - chunks are aligned to newline boundaries,
  - each worker writes to private vectors,
  - results are concatenated in file order for deterministic output,
  - small files keep the serial path.
- Added focused `points3D.txt` parser tests, including comments, blanks, CRLF, no-final-newline files, long track tails, malformed lines, empty/comment-only files, track count modes, filtering behavior, round-trip behavior, and deterministic large-file parsing.

## Timing Results

Dataset:

```text
D:\temp\gs_tests\POPERINGE\outdoor\colmap
```

Point count:

```text
1,699,644
```

### Normal Load

| Metric | Baseline | Step 1 | Step 3 | Improvement vs Baseline |
|---|---:|---:|---:|---:|
| `points3D.txt` parse | `9199.83 ms` | `2541.97 ms` | `643.11 ms` | `14.31x faster`, `93.0% less time` |
| COLMAP Loading | `10425.68 ms` | `3840.90 ms` | `1962.09 ms` | `5.31x faster`, `81.2% less time` |

### Min-Track Load

Command mode:

```text
--min-track-length 5
```

Filter result stayed unchanged:

```text
1,699,644 -> 219,184
```

| Metric | Baseline | Step 2 | Step 3 | Improvement vs Baseline |
|---|---:|---:|---:|---:|
| record parse | `11699.04 ms` | `2512.79 ms` | `516.94 ms` | `22.63x faster`, `95.6% less time` |
| full `Read points3D.txt` | `18793.76 ms` | `2630.20 ms` | `633.73 ms` | `29.66x faster`, `96.6% less time` |
| COLMAP Loading | `20158.49 ms` | `4004.66 ms` | `2017.84 ms` | `9.99x faster`, `90.0% less time` |

### DIFF
--- a/src/io/formats/colmap.cpp
+++ b/src/io/formats/colmap.cpp
@@ -1411,6 +1411,7 @@ namespace lfs::io {
 
             const auto chunks = split_line_aligned_chunks(std::span<const char>(*buffer));
             std::vector<RecordChunkResult> results(chunks.size());
+            // oneTBB propagates parse/cancellation exceptions to the caller and cancels sibling work.
             tbb::parallel_for(size_t{0}, chunks.size(), [&](const size_t chunk_index) {
                 const auto& chunk = chunks[chunk_index];
                 auto& result = results[chunk_index];
@@ -1439,6 +1440,11 @@ namespace lfs::io {
             }
         }
         buffer.reset();
+
+        if (points.empty()) {
+            LOG_ERROR("No valid points found in {}", lfs::core::path_to_utf8(file_path));
+            throw std::runtime_error("No valid points in points3D.txt");
+        }
 
         LOG_DEBUG("Reading {} 3D points from text file", points.size());
         const char* mode_name = track_mode == TrackParseMode::Full ? "full" : (track_mode == TrackParseMode::CountOnly ? "count_only" : "none");
@@ -1493,14 +1499,15 @@ namespace lfs::io {
 
         detail::ColmapPoint3DTextPointCloudData data;
         data.byte_size = file_size_ec ? 0 : byte_size;
-        if (!file_size_ec) {
-            const auto estimated_points = static_cast<size_t>(std::max<uintmax_t>(byte_size / 96, 1));
-            data.positions.reserve(estimated_points * 3);
-            data.colors.reserve(estimated_points * 3);
-        }
 
         auto buffer = read_binary(file_path);
         if (buffer->size() < POINTS3D_PARALLEL_MIN_BYTES) {
+            if (!file_size_ec) {
+                const auto estimated_points = static_cast<size_t>(std::max<uintmax_t>(byte_size / 96, 1));
+                data.positions.reserve(estimated_points * 3);
+                data.colors.reserve(estimated_points * 3);
+            }
+
             data.file_lines = for_each_data_line(
                 std::span<const char>(*buffer),
                 options,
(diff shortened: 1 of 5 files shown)


kmeirlaen:
@MrNeRF let me know if the colmap comments are a concern. They are also valid for the .bin files  Would suggest to accept this and tackle when it becomes problematic?

src/io/formats/colmap.cpp
```diff
@@ -1177,173 +1382,206 @@ namespace lfs::io {
...
+            std::vector<RecordChunkResult> results(chunks.size());
+            // oneTBB propagates parse/cancellation exceptions to the caller and cancels sibling work.
+            tbb::parallel_for(size_t{0}, chunks.size(), [&](const size_t chunk_index) {
+                const auto& chunk = chunks[chunk_index];
+                auto& result = results[chunk_index];
+                result.points.reserve(std::max<size_t>(static_cast<size_t>(chunk.end - chunk.begin) / 96, 1));
+                result.file_lines = for_each_data_line(
+                    std::span<const char>(chunk.begin, static_cast<size_t>(chunk.end - chunk.begin)),
+                    options,
+                    "COLMAP point cloud parse cancelled",
```
MrNeRF, line 1418: localisation?

src/io/formats/colmap.cpp
```diff
@@ -1177,173 +1382,206 @@ namespace lfs::io {
...
+                 mode_name,
+                 file_lines,
                  elapsed_ms(parse_start));
         return points;
     }
 
-    PointCloud read_point3D_text(const std::filesystem::path& file_path,
-                                 const LoadOptions& options = {}) {
-        LOG_TIMER_TRACE("Read points3D.txt point cloud");
+    std::vector<detail::ColmapPoint3DTextRecordData> detail::parse_points3D_text_records_for_test(
```
MrNeRF: stuff for testing should not go here. Tests must test the deployed implementations and not artificially generated data.

MrNeRF:
> @MrNeRF let me know if the colmap comments are a concern. They are also valid for the .bin files Would suggest to accept this and tackle when it becomes problematic?

I agree!

src/io/formats/colmap.cpp
kmeirlaen: all of the current loader messages in this area are hard-coded English strings, so full localization is not established here - introducing it is out of scope of this PR

PR #1368, opened 2026-07-03
Video Frame Extractor enhancements

francescofugazzi:
# Summary

This pull request significantly enhances the **Video Frame Extractor**, introducing smarter frame selection, improved metadata generation, automatic rotation handling, GPU quality fixes, and several usability improvements.

Highlights include:

- Sharpness-based filtering to discard blurry frames
- Configurable candidate sampling for Window Mode extraction
- Automatic metadata JSON generation alongside extracted frames
- Automatic video rotation detection with rotation applied consistently across every extraction path
- GPU JPEG quality fix so the quality slider is respected in all pipelines
- Overwrite confirmation dialog when the output folder already contains extracted frames
- Improved progress reporting and localized UI
- Preview stability improvements after repeated rotations

---

# New Features

## Sharpness Filtering

A new **Sharpness Filter** section has been added to the extraction settings panel, allowing blurry frames to be discarded before they are saved.

### Available algorithms

- **Tenengrad** - Sobel-based edge energy for fast directional blur detection
- **Laplacian** - Variance of the Laplacian for detecting out-of-focus blur
- **Combined** *(default)* - Combines Tenengrad and Laplacian for the best overall accuracy.

### Filtering modes

Two filtering methods are available.

#### Threshold Mode

Frames whose sharpness score falls below the selected threshold are discarded.

The threshold ranges from:

- **0** → Disabled (no filtering)
- **100** → Very strict filtering

#### Window Mode

Instead of evaluating every frame, the extractor evaluates only a configurable number of candidate frames within each interval and saves the sharpest one.

Available presets:

- Auto
- 3 (Fast)
- 5 (Balanced)
- 10 (Quality)
- 20 (High Quality)
- 50 (Very High)
- All Frames

Candidate selection happens before frame processing, allowing skipped frames to completely bypass expensive operations such as:

- Pixel conversion
- Sharpness computation
- GPU/CPU copies

An effective readout (for example **10 / ~25 frames**) shows the actual number of evaluated candidates for each interval.

The sampling algorithm has also been rewritten to use zero-based bucket sampling, eliminating the previous off-by-one oversampling issue.

---

## Metadata JSON Output

When enabled, the extractor generates an `extracted_metadata.json` file inside the output directory.

The metadata includes:

- Source video information
- Extraction parameters
- Sharpness filtering configuration
- Rotation angle
- Output image size
- Per-frame information:
  - Filename
  - Timestamp
  - Source frame number
  - Sharpness score

When the detected rotation is **90°** or **270°**, the reported output dimensions are automatically swapped so the metadata matches the actual saved images.

---

## Rotation Controls

Rotation support has been expanded throughout the extractor.

Features include:

- CW / CCW buttons in the transport bar
- Real-time preview rotation
- Automatic rotation detection from FFmpeg metadata
- Consistent rotation during extraction in:
  - Software decoding
  - Hybrid decoding (NVDEC → CPU transfer)
  - Full GPU pipeline (NV12 → RGB → JPEG)
  - Window Mode extraction

The GPU extraction pipeline now also correctly applies the selected JPEG quality instead of always using the nvJPEG default quality.

---

## Overwrite Confirmation

Before extraction starts, the output directory is checked for existing `.jpg` and `.png` files.

If extracted frames are found, an in-place confirmation dialog lets the user either:

- **Cancel**
- **Proceed**, which recursively clears the output directory before extraction begins

---

## Progress Reporting

Progress reporting has been improved.

The extraction progress now displays:

- Current extracted frames
- Estimated total frames
- Number of discarded frames (when filtering is enabled)

The completion dialog also reports discarded frames when applicable, for example:

> **Extraction completed successfully! 60 frames extracted (12 discarded)**

---

## Preview Stability

Fixed a Vulkan resource lifetime issue affecting the preview after repeated rotations.

Preview textures are now destroyed through deferred cleanup, preventing corrupted UI fragments caused by RmlUI still referencing images scheduled for deletion.

---

# Localization

All newly introduced UI strings have been translated into **all 10 supported languages**:

- German
- English
- Spanish
- French
- Italian
- Japanese
- Korean
- Dutch
- Polish
- Chinese

### Added localization keys

- `sharpness`
- `sharpness_filter`
- `sharpness_algorithm`
- `sharpness_algo_combined`
- `sharpness_algo_tenengrad`
- `sharpness_algo_laplacian`
- `sharpness_mode`
- `sharpness_mode_threshold`
- `sharpness_mode_window`
- `sharpness_mode_desc_threshold`
- `sharpness_mode_desc_window`
- `sharpness_threshold_label`
- `sharpness_coverage_label`
- `candidates_auto`
- `candidates_fast`
- `candidates_balanced`
- `candidates_quality`
- `candidates_high_quality`
- `candidates_very_high`
- `candidates_all`
- `candidates_readout_fmt`
- `generate_metadata`
- `overwrite_prompt`
- `overwrite_clear`
- `overwrite_cancel`
- `rotate_cw`
- `rotate_ccw`
- `discarded_format`
- `all`

---

# Files Changed

## Video Processing

- `src/io/video/color_convert.cu`
  - Added CUDA rotation kernel
  - Fixed 90° / 270° rotation formulas

- `src/io/video/color_convert.cuh`
  - CUDA rotation kernel declarations

- `src/io/nvcodec_image_loader.cpp`
  - Fixed GPU JPEG quality handling

- `src/io/video_frame_extractor.cpp`
  - Sharpness filtering
  - Window Mode candidate sampling
  - Rotation support across all extraction paths
  - Metadata generation
  - Progress reporting

- `src/io/video_frame_extractor.hpp`
  - Added `SharpnessParams`
  - Extended extraction parameters

- `src/io/video_player.cpp`
  - Automatic rotation metadata detection

- `src/io/video_player.hpp`
  - Added `rotation()` accessor

---

## User Interface

- `src/visualizer/gui/windows/video_extractor_dialog.cpp`
  - Sharpness filtering UI
  - Candidate selection presets
  - Progress reporting
  - Overwrite confirmation dialog
  - Rotation controls

- `src/visualizer/gui/windows/video_extractor_dialog.hpp`
  - Added new UI controls and state

- `src/visualizer/gui/rmlui/resources/video_extractor.rml`
  - Sharpness filtering UI
  - Rotation controls
  - Overwrite confirmation overlay

- `src/visualizer/gui/rmlui/resources/video_extractor.rcss`
  - Layout improvements
  - Styling for new controls
  - Overlay styling

- `src/visualizer/gui/vulkan_ui_texture.cpp`
  - Deferred Vulkan image destruction to prevent preview corruption after repeated rotations

---

## Localization

Updated all supported locale files:

- `src/visualizer/gui/resources/locales/de.json`
- `src/visualizer/gui/resources/locales/en.json`
- `src/visualizer/gui/resources/locales/es.json`
- `src/visualizer/gui/resources/locales/fr.json`
- `src/visualizer/gui/resources/locales/it.json`
- `src/visualizer/gui/resources/locales/ja.json`
- `src/visualizer/gui/resources/locales/ko.json`
- `src/visualizer/gui/resources/locales/nl.json`
- `src/visualizer/gui/resources/locales/pl.json`
- `src/visualizer/gui/resources/locales/zh.json`

Also updated:

- `src/visualizer/gui/string_keys.hpp`

---

## Assets

Added new toolbar icons:

- `src/visualizer/gui/assets/icon/sequencer/rotate-cw.png`
- `src/visualizer/gui/assets/icon/sequencer/rotate-ccw.png`


kmeirlaen:
Nice one, cant wait to test it out!

Actually, already testing - think the rotation is not applied?

image

kmeirlaen:
During extractions, it shows the discarted frames:
image
Can you also show it when completed?  Now it just shows "60 frames extracted" - would be good to know how many were not selected
image

francescofugazzi:
> Nice one, cant wait to test it out!
> 
> Actually, already testing - think the rotation is not applied?
> 
> image

can you try if it works with png output?

francescofugazzi:
@kmeirlaen

Thanks for testing and for the feedback!

### Rotation not applied

The issues were:

- In the full GPU pipeline (**NV12 → RGB → JPEG**), the rotation step was completely missing. I added the `rotateRgbCuda` call before the batch JPEG encoder.
- The CUDA rotation kernel itself had the **90°** and **270°** formulas swapped. They were doing the opposite of what the comments described. Both branches have now been corrected.

Rotation should now work correctly in all three extraction paths:

- Software
- Hybrid
- Full GPU

Give it another try when you have a chance.

### Discarded frame count

Good suggestion. I've added it.

The completion message now looks like:

> **Extraction completed successfully! 60 frames extracted (12 discarded)**

instead of simply:

> **60 frames extracted**

The update has already been pushed to the same branch.

kmeirlaen:
I'm doing some further testing so expect some more comments (functional wise)
This one is weird:
- jpg output, others are default settings, file sizes of frames are about 600-800kb
- same settings, butg "discard below threshold" - file sizes are about 600-800kb
- same settings but "pick sharpest per interval" - file sizes are 1.5Mb and larger -> can you explain this?

kmeirlaen:
Suggestion (no blocker)
for "for pick sharpest per interval", the routine loops through the video from strat to end frame and extracts all frames, calculates the score for all those frames and then when the inteval is reached, writes out the ones with the sharpest values.

So in the default case:
- 1 fps
- loop through 30 image (30 fps video)
- calculate score for 30 images
- write the sharpest

Though this works, it can be slow. Could we do something like this:

- set the extractor to 5 fps
- set "pick sharpest XX per interval" -> set to 1
- extract the 5 frames that would normally be extracted
- calculate the score for those 5 frames
- write the sharpest 1

Basically, it now has this 
- set the extractor to 30 fps
- set "pick sharpest XX per interval" -> set to 1
- extract the 30 frames 
- calculate the score for those 30 frames
- write the sharpest 1
 
So basically, the "pick sharpest XX per interval" selects a subset of the original fps. hope it makes sense.

kmeirlaen:
image

This "Close Window" should be "cancel". Maybe you are reusing some labels?

Also, the message "clear all contents" is real: it deletes ALL data in the folder! (also existing other data, subfolders, etc..)

kmeirlaen:
Localisation is incomplete - only .en and .it are covered

image


[08:30:54.252] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_filter
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_algorithm
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_algo_combined
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_algo_tenengrad
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_algo_laplacian
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_mode
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_mode_threshold
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_mode_window
[08:30:54.253] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.sharpness_threshold_label
[08:30:54.254] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.generate_metadata
[08:30:54.254] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.overwrite_prompt
[08:30:54.254] [warn] localization_manager.cpp:87  Missing localization key: video_extractor.overwrite_clear

kmeirlaen:
Some LLM feedback - no blockers I think:

- Default threshold inconsistency: sharpness_threshold defaults to 40.0 in the struct (dialog.hpp) and the readIntValue fallback in beginExtractionFromUi is 40 (:1433), but syncControls uses fallback 10 (:1013) and the RML slider defaults to 10. Pick one.
- Doc/impl mismatch: SharpnessParams::threshold comment says "0 = auto (percentile-based, picks best 80%)" (hpp:42), but the code only filters when threshold > 0.0 — there is no percentile auto-mode. Either implement it or fix the comment (and note that setting the slider to 0 silently disables filtering even with sharpness "enabled").
- Per-frame heap churn: rotation allocates a fresh std::vector every frame (:780, :889). Hoist a reusable buffer.
- Sharpness in full-GPU path defeats the pipeline: it does a synchronous cudaMemcpy DtoH per frame (:674), negating the "stay on GPU" benefit. Acceptable for now, but worth a comment or a GPU-side sharpness kernel later.
- Window-mode memory: window_candidates holds full RGB frames for every frame in an interval; large frame_interval at high resolution can hold hundreds of MB. Consider keeping only the current-best frame + score instead of all candidates.
- Dead field: CandidateFrame::filename is computed (:682) but never used — flush_window recomputes via written_count + 1.
- Metadata naming: root["source_frames"] is actually the trim-scaled count, not the source total (:1065) — mildly misleading.

kmeirlaen:
Gonna be great to have this in, if you need help completing, do let us know!

francescofugazzi:
@kmeirlaen

Thanks for the detailed review! Here's what I changed since your last feedback:

### GPU JPG quality

This turned out to be a pre-existing issue in the nvJPEG encoder rather than something introduced by this PR.

The full GPU pipeline was using `NVIMGCODEC_QUALITY_TYPE_DEFAULT`, which completely ignored the quality slider and always fell back to nvJPEG's internal default quality (**75**).

This has now been fixed by switching to `NVIMGCODEC_QUALITY_TYPE_QUALITY`, so the user-selected quality value is correctly applied in the full GPU path as well.

---

### Preview corruption after rotation

I found and fixed a separate bug in `VulkanUiTexture::ensureImage()`.

When the preview image was resized, the previous Vulkan image was destroyed immediately while RmlUI could still have in-flight descriptor sets referencing it. This occasionally produced corrupted UI fragments after repeated rotations.

The fix introduces a deferred destruction queue so old image handles remain valid until the next cleanup cycle.

---

### Window mode candidates

The old percentage-based coverage slider has been replaced with a dropdown:

- Auto
- 3 (Fast)
- 5 (Balanced)
- 10 (Quality)
- 20 (High)
- 50 (Very High)
- All Frames

Candidate selection now happens **before** `process_frame_hw()` / `process_frame_sw()`, meaning frames that won't be evaluated completely skip all expensive operations, including:

- Pixel conversion
- Sharpness computation
- Memory copies

An effective readout is also displayed, for example:

> **10 / ~25 frames**

so the actual evaluation budget is always visible.

---

### Bucket sampling

Replaced the previous step/modulo implementation (which had an off-by-one issue causing slight oversampling) with a zero-based bucket sampling approach.

The number of evaluated candidates now never exceeds the selected target.

---

### Frame Interval fix

Skipped frames were not advancing the window counter, which caused incorrect window boundaries when using **Frame Interval** mode.

This has now been fixed.

---

### Localization

All newly added strings have been translated into all 10 supported languages:

- German
- English
- Spanish
- French
- Italian
- Japanese
- Korean
- Dutch
- Polish
- Chinese

That said, I can only personally verify the Italian translations. The others should ideally be reviewed by native speakers if needed.

---

Let me know how it feels now, especially the **GPU JPG quality** fix and the new **Window Mode candidates** dropdown.

kmeirlaen:
would you mind another sync with master? at the time of your sync, some new packages were introduced which result in very long configure/buld times.  this was resolved in latest master...

francescofugazzi:
> would you mind another sync with master? at the time of your sync, some new packages were introduced which result in very long configure/buld times. this was resolved in latest master...

done

P.S. Yeah I noticed that it was so much longer when I recompiled everything this afternoon ;)

kmeirlaen:
I have hard time understanding the frames that are extracted. 

### example 1
image

I would expect:
- for each second, we extract 10 frames (10fps in main slider)
- we calculate the score for those 10 frames and select the best 3 (so it should show 3/10)

### example 2
image

- for each second, we extract 30 frames (30fps in main slider)
- we calculate the score for those 30 frames and select the best 30 (so yeah, all should be saved - maybe skip calculation even)

### exxample 3
image

- we extract 5 frames
- we select the best 3


I would want an option to select 2 out of 10 and 1 out of 10 as well 
(now its 3/5/20/50) -> it assumes 60fps for the video - but my video is only 30fps

I get this inspiration from [sharpframes](https://sharp-frames.reflct.app/) where the app first goes through the video with an fps count you select and then you can filter out of those images. 

They also allow selection within a higher timeframe (for example, select 5 best images out of 2 second timeframe)
They also show the total number of frames that will be downloaded, which is helpful as i tend to target ~600 images if possible

This is not a blocker though, its just not clear for me personally.

Will go through the other changes as well, looks promising!

francescofugazzi:
> I have hard time understanding the frames that are extracted.
> 
> ### example 1
> image
> I would expect:
> 
> * for each second, we extract 10 frames (10fps in main slider)
> * we calculate the score for those 10 frames and select the best 3 (so it should show 3/10)
> 
> ### example 2
> image
> * for each second, we extract 30 frames (30fps in main slider)
> * we calculate the score for those 30 frames and select the best 30 (so yeah, all should be saved - maybe skip calculation even)
> 
> ### exxample 3
> image
> * we extract 5 frames
> * we select the best 3
> 
> I would want an option to select 2 out of 10 and 1 out of 10 as well (now its 3/5/20/50) -> it assumes 60fps for the video - but my video is only 30fps
> 
> I get this inspiration from [sharpframes](https://sharp-frames.reflct.app/) where the app first goes through the video with an fps count you select and then you can filter out of those images.
> 
> They also allow selection within a higher timeframe (for example, select 5 best images out of 2 second timeframe) They also show the total number of frames that will be downloaded, which is helpful as i tend to target ~600 images if possible
> 
> This is not a blocker though, its just not clear for me personally.
> 
> Will go through the other changes as well, looks promising!

Actually, I think there's a misunderstanding about how the candidates control works. The FPS slider creates intervals on the timeline (shown as segments in the preview bar). For each interval, the system evaluates theselected number of candidates and then saves only the single best frame - not multiple frames. So "10 candidates"means "evaluate 10 frames and pick the best 1", not "save 10 frames".

kmeirlaen:
I understand it, but i'm pointing out I (personally) like it differently. 
Also, in your case can i pick the best 2 out of 10 ?
That is different than pick 1 out of 5

2 out of 10 could be the first 2 frames it evaluates
1 out of 5 will select only the first of those 2 and then another frame from the last 5.

And also, the slider for FPS becomes irrelevant, correct?

Anyway, good job!  We can always iterate on this, its already a very good addition!

kmeirlaen:
90° rotation with JPG as output is still broken (does not output any frames)

kmeirlaen:
I would also not hard DELETE the contents of a directory. This sound scary to me.  As said, all contents is deleted (including .txt files, .bin files, subfolders, etc...). 

Maybe only delete the .jpg/png in the folder?

francescofugazzi:
> 90° rotation with JPG as output is still broken (does not output any frames)

gosh... I had a regression

francescofugazzi:
> I would also not hard DELETE the contents of a directory. This sound scary to me.  As said, all contents is deleted (including .txt files, .bin files, subfolders, etc...). 
> 
> 
> 
> Maybe only delete the .jpg/png in the folder?
> 
> 
> 
> 

ok and also the json metadata e no subfolders. I will change it

kmeirlaen:
For the UI/UX, lets wait to change until we have some feedback from other users.  Once the other 2 remarks are fixed and there was a code review done - merge can be performed and we will have more feedback.

francescofugazzi:
@kmeirlaen I fixed the rotation problem on GPU pipeline and changed the delete content behaviour

MrNeRF:
What's the state here? Anything missing?

MrNeRF (approved):

kmeirlaen:
merged, great enhancement, thanks!

PR #1369, opened 2026-07-04
First-class spherical undistort support

lyehe:
## Summary

Makes the existing `--undistort` path work for spherical/equirectangular COLMAP
datasets by expanding each panorama into internal pinhole views at load time.

The generated views stay virtual: they reuse the source panorama and mask paths,
flow through the projection-aware image loader/cache, and never write split
images to disk. The CLI and UI are unchanged.

Each panorama expands into **6 cube faces at 96 degrees FOV**, resampled with an
**anisotropic (EWA) filter**. Both of those choices are load-bearing and are
explained below, because the obvious alternatives are measurably worse.

## Why 6 faces at 96 degrees

Every view expanded from one panorama shares that panorama's optical centre.
`pano_to_view` is orthonormal, so

```
C_face = -(M R)^T (M T) = -R^T M^T M T = -R^T T = C_source
```

Overlapping views from a common centre therefore re-sample identical rays and
add no parallax. Redundant angular coverage is pure cost, and it is *uneven*
cost: it weights the photometric loss by however many views happen to see a
direction.

Measured over 2M Monte Carlo directions on the sphere:

| layout | uncovered | coverage spread | mean |
|---|---|---|---|
| 12 icosahedral @ 90 deg | 0% | 25% 1x / 50% 2x / **25% 3x** | 2.000 |
| **6 cube faces @ 96 deg** | 0% | **88.7% 1x** / 10.9% 2x / 0.4% 3x | 1.117 |

Six faces tile the sphere. The extra 3 degrees per edge is a deliberate seam
margin so SSIM windows and densification see context across face boundaries, at
~12% angular overcoverage instead of 100%.

On a 3840x1920 panorama this is the **same total pixel budget** as the previous
layout (11.06 Mpx, 1.001x), so training cost is unchanged — the saving from
dropping redundant views pays for the higher per-face resolution below.

## Why the face size changed

`resolveSphericalViewSize` used `width * fov / 360`. That matches the total pixel
*count* across the field of view, but not the sampling *rate*: a gnomonic face
spaces its columns non-uniformly (dense at the edges, sparse at the centre) while
equirectangular spacing is uniform in angle.

The result was **1.27x undersampling at the centre of every face** — aliasing
precisely where the panorama is sharpest — while oversampling the corners ~3x.
Sizing now matches the source angular rate at the face centre:

```
size = 2 * max(W/2pi, H/pi) * tan(fov/2)
```

For a 4K panorama at 96 degrees that is 1358 px rather than 960 px.

## Why EWA resampling

The previous resampler was a bare bilinear fetch with no prefilter, so the
minification above aliased outright. The map minifies at the centre of every face
and magnifies towards the corners, so no fixed kernel is correct across a single
face.

The projection now derives the analytic Jacobian `J = d(source px)/d(face px)` and
builds a per-pixel elliptical footprint with covariance

```
V = sigma_dst^2 * J J^T + sigma_src^2 * I
```

Both sigmas are variance-matched to a one-pixel box (`1/sqrt(12)`), which makes
the filter reduce to exactly bilinear sharpness at unity magnification (a tent has
variance 1/6, and 1/12 + 1/12 = 1/6) and widen only where the map genuinely
minifies. Footprints are strided near the poles so the tap count stays bounded.

Resampling error against a 16x16 supersampled reference, sweeping the signal
across the face Nyquist limit (RMSE in 8-bit levels, lower is better):

| signal | regime | bilinear | EWA | improvement |
|---|---|---|---|---|
| 20 cyc/rad | below Nyquist | 8.145 | **1.089** | 7.5x |
| 45 cyc/rad | below Nyquist | 17.175 | **2.563** | 6.7x |
| 100 cyc/rad | aliasing | 29.579 | **5.548** | 5.3x |
| 140 cyc/rad | aliasing | 31.673 | **8.357** | 3.8x |

Worst single face 2.1x, best 8.3x.

## Behaviour

- `--undistort` on equirectangular COLMAP imports (model 17 `EQUIRECTANGULAR`,
  legacy name `SPHERICAL`) expands each panorama into 6 internal pinhole cameras.
- Image loading, eval loading, masks, alpha masks and immediate strategy loads all
  go through the same projection-aware loader path.
- Faces inherit `split()` from the source panorama, so all 6 faces of a panorama
  land on the same side of a train/eval split and no shared rays leak across it.
- Without `--undistort` or `--gut`, non-pinhole datasets still fail early with a
  direct message.
- Depth priors are projected onto the faces when provided, with nearest sampling
  and a radial-to-z conversion (see below). Both 8-bit and 16-bit sidecars work.
- `--gut` remains the native non-pinhole training path and is untouched.

## Depth priors

Equirectangular depth stores distance along the ray from the camera centre, but
LFS supervises camera-space z: fastgs takes the third row of world-to-camera for
its depth, and `depth_anchor_collect_kernel` pairs the prior against that same z
when fitting the per-camera affine anchor. Projection therefore applies

```
z = r / sqrt(1 + u^2 + v^2)
```

for a face pixel at plane coordinates `(u, v)`. Because `pano_to_face` is
orthonormal this is exactly the normalisation term the projection already
computes, so it is free. Without it the corner of a 96 degree face is wrong by
~1.9x, and since the error varies across the image the loss's
scale-and-shift-invariant fit cannot absorb it -- that fit is affine in the
prior, not in pixel position.

Depth also uses nearest sampling rather than the colour path's EWA filter:
averaging across a depth discontinuity produces a value in neither the
foreground nor the background, and the loss would then pull geometry towards
that invented surface.

This assumes the panorama stores radial distance, the usual convention. A
disparity prior would need the reciprocal correction; that is documented at the
function.

## Validation

Correctness of the projection, via a harness linked against the real
`project_cube_face_hwc`:

- Analytic Jacobian vs central differences, 6 faces x 49 probes: worst relative
  error **6.656e-07**.
- Face pixels vs the source panorama sampled along the same ray, on a real RaR
  panorama: **1.582 levels RMS** over 72,000 samples, confirming the face basis is
  not transposed or mirrored.

Depth projection, against an analytic scene (a cube of half-extent h centred on
the camera, encoded as 16-bit equirectangular radial distance). Camera-space z on
a flat wall must be constant and equal to h wherever it is sampled on the face,
which holds only if the radial-to-z conversion is applied:

- faces +X/-X/+Z/-Z: mean error 0.00027, max 0.00110 over 46,656 px
- faces +Y/-Y: mean error 0.00028, max 0.00093 over 46,656 px
- off-axis sample: z = 0.49949 against 0.50000 expected
- the raw radial value at that same point differs by > 0.05, so the check fails
  if the conversion is removed

End to end on the RaR `I_alley` scene (145 equirectangular panoramas, COLMAP
model 17, 3840x1920):

- Expands to 870 faces at 1358 px, 96 degrees, scene scale 4.0679965.
- Eval split 756 train + 114 val, i.e. 19 whole panoramas held out, so no face
  of a panorama straddles the split.

## How this compares to `--gut`

`--gut` already trains natively on equirectangular images, so the obvious
question is whether this path earns its place. Both were trained from scratch on
`I_alley` with identical settings (`-r 1 -i 30000 --eval --headless`; headless
honours `-i` exactly, so the step schedule is the same for both).

Their own eval metrics are **not** comparable -- `--undistort` scores 1358 px
faces, `--gut` scores 3840x1920 panoramas, and equirectangular PSNR is dominated
by the heavily oversampled polar regions. To get one yardstick, both runs' saved
`[GT | render]` eval images were projected onto the same six 96 degree cube faces
with the production resampler and scored there (114 samples each). The identical
resampling is applied to GT and render alike, so it largely cancels.

| | PSNR | SSIM | gaussians | wall clock |
|---|---|---|---|---|
| `--undistort` (this PR) | 26.836 | 0.8347 | 3.90M | **755 s** |
| `--gut` | **27.045** | **0.8369** | 4.42M | 3100 s |

At 7000 iterations the ordering is the same (24.191 / 0.7803 against
24.781 / 0.7947).

So `--gut` is marginally ahead on quality -- about 0.2 dB at 30k -- while this
path trains **4.1x faster** and converges to 12% fewer gaussians. The quality gap
is small enough that the extra resampling on the `--gut` side of the measurement
could account for some of it: low-pass filtering both GT and render tends to
raise PSNR, and only `--gut` goes through that step. The honest reading is that
quality is close to on par and the real difference is throughput.

That is the tradeoff this PR offers, and it is worth stating plainly rather than
claiming a quality win: `--undistort` is the fast path for panoramic data,
`--gut` remains the accurate one.

Everything above is reproducible from `tests/test_undistort.cpp` rather than from
throwaway scripts. Each test builds its panorama analytically, so none needs a
dataset and the expected value is a closed form rather than a golden image:

| test | what would break it |
|---|---|
| `FacePixelsCarryTheContentOfTheirOwnRay` | a transposed or mirrored face basis |
| `ConstantPanoramaIsPreservedOnEveryFace` | filter weights not summing to one, including at the strided poles |
| `SignalAboveFaceNyquistIsAttenuatedNotAliased` | dropping the prefilter (a plain bilinear fetch fails this) |
| `FlatWallHasConstantCameraSpaceZ` | dropping the radial-to-z conversion |
| `RadialToZConversionIsApplied` | negative control for the above |
| `MatchesPanoramaSamplingRateAtFaceCentre` | changing the sizing rule (pins 1358 / 1222 / 724 px) |
| `FocalRoundTripsToTheRequestedFov` | focal and fov disagreeing |
| `DegenerateFovStaysFinite` | removing the field-of-view clamp |

Plus the pre-existing expansion tests (view count, face size, focal,
orthonormality of every face basis, uid allocation, passthrough of non-spherical
cameras). All 8 new tests pass; the 7 `UndistortCameraTest` cases are
dataset-gated and skip without one, as they did before.

## Error handling

`expandEquirectangularCamerasForUndistort` returns `lfs::Result<T>` and reports
through `lfs::make_error`, so failures carry a code, a domain and a detection
site. An earlier revision used `std::expected<T, std::string>` and tripped the
no-new-error-debt gate by adding three sites on top of the baseline's 15 for
`training_setup`. `tools/error_debt_census.py --baseline` now reports no new
violations.

## Notes for reviewers

- The eval numbers above are measured **in face space**, not on equirectangular
  images, so they are not directly comparable to published panoramic PSNR figures
  (which include the pole regions where equirect heavily oversamples). They are
  useful for comparing configurations of this path against each other.
- `autoScaleSteps` receives the post-expansion camera count (870 rather than 145),
  which stretches the step schedule ~2.9x and contributed to the 4.62M gaussian
  count above. That behaviour is unchanged by this PR, but it is worth a separate
  decision about whether the heuristic should count capture positions rather than
  views, since the six faces of a panorama share one optical centre.
- Depth support is verified analytically. No panoramic depth dataset was
  available to exercise it end to end, and the projection assumes the prior
  encodes radial distance.

### DIFF
--- a/src/core/camera.cpp
+++ b/src/core/camera.cpp
@@ -180,10 +180,13 @@ namespace lfs::core {
           _T(std::move(other._T)),
           _radial_distortion(std::move(other._radial_distortion)),
           _tangential_distortion(std::move(other._tangential_distortion)),
+          _camera_model_type(other._camera_model_type),
           _image_path(std::move(other._image_path)),
           _image_name(std::move(other._image_name)),
           _mask_path(std::move(other._mask_path)),
           _depth_path(std::move(other._depth_path)),
+          _cube_face_projection(std::move(other._cube_face_projection)),
+          _has_alpha(other._has_alpha),
           _normal_path(std::move(other._normal_path)),
           _split(other._split),
           _camera_width(other._camera_width),
@@ -235,10 +238,13 @@ namespace lfs::core {
             _T = std::move(other._T);
             _radial_distortion = std::move(other._radial_distortion);
             _tangential_distortion = std::move(other._tangential_distortion);
+            _camera_model_type = other._camera_model_type;
             _image_path = std::move(other._image_path);
             _image_name = std::move(other._image_name);
             _mask_path = std::move(other._mask_path);
             _depth_path = std::move(other._depth_path);
+            _cube_face_projection = std::move(other._cube_face_projection);
+            _has_alpha = other._has_alpha;
             _normal_path = std::move(other._normal_path);
             _split = other._split;
             _camera_width = other._camera_width;
@@ -288,6 +294,8 @@ namespace lfs::core {
           _image_path(other._image_path),
           _mask_path(other._mask_path),
           _depth_path(other._depth_path),
+          _cube_face_projection(other._cube_face_projection),
+          _has_alpha(other._has_alpha),
           _normal_path(other._normal_path),
           _split(other._split),
           _camera_width(other._camera_width),
@@ -346,6 +354,10 @@ namespace lfs::core {
         return {_focal_x * x_scale, _focal_y * y_scale, _center_x * x_scale, _center_y * y_scale};
     }
 
+    void Camera::set_cube_face_projection(CubeFaceProjection projection) {
+        _cube_face_projection = std::move(projection);
+    }
+
     Tensor Camera::load_and_get_image(int resize_factor, int max_width, const bool output_uint8,
                                       const bool update_dimensions) {
         const ImageLoadParams params{
@@ -376,7 +388,10 @@ namespace lfs::core {
 
     void Camera::load_image_size(int resize_factor, int max_width) {
         int w, h;
-        if (_undistort_prepared) {
+        if (_cube_face_projection) {
+            w = _cube_face_projection->face_size;
+            h = _cube_face_projection->face_size;
+        } else if (_undistort_prepared) {
             w = _camera_width;
             h = _camera_height;
         } else {
@@ -423,11 +438,11 @@ namespace lfs::core {
     }
 
     size_t Camera::get_num_bytes_from_file(int resize_factor, int max_width) const {
-        auto result = get_image_info(_image_path);
-
-        int w = std::get<0>(result);
-        int h = std::get<1>(result);
-        int c = std::get<2>(result);
+        auto [w, h, c] = get_image_info(_image_path);
+        if (_cube_face_projection) {
+            w = _cube_face_projection->face_size;
+            h = _cube_face_projection->face_size;
+        }
 
         if (resize_factor > 0) {
             w = w / resize_factor;
@@ -450,6 +465,10 @@ namespace lfs::core {
 
     size_t Camera::get_num_bytes_from_file() const {
         auto [w, h, c] = get_image_info(_image_path);
+        if (_cube_face_projection) {
+            w = _cube_face_projection->face_size;
+            h = _cube_face_projection->face_size;
+        }
         size_t num_bytes = w * h * c * sizeof(uint8_t);
         return num_bytes;
     }
--- a/src/core/include/core/camera.hpp
+++ b/src/core/include/core/camera.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "core/camera_types.h"
+#include "core/cube_face_projection.hpp"
 #include "core/cuda/undistort/undistort.hpp"
 #include "core/export.hpp"
 #include "core/tensor.hpp"
@@ -14,6 +15,7 @@
 #include <cuda_runtime.h>
 #include <filesystem>
 #include <future>
+#include <optional>
 #include <string>
 
 namespace lfs::core {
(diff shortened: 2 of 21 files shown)


ichsan2895:
Just suggestion. Please compare the benchmark (PSNR, SSIM, time/step) for spherical undistort versus spherical with GUT

PR #1371, opened 2026-07-04
fix duplicate of models with deleted splats

kmeirlaen:
## Summary

Scene graph Duplicate now creates a compact splat model that contains only currently visible splats. Soft-deleted rows are no longer resurrected in duplicated nodes.

## Observed Symptoms

1. Open a splat model.
2. Use brush selection to select and remove some splats.
3. In the scene graph, select the model and choose Duplicate.
4. The duplicated model appears with the removed splats visible again.

## What This Patch Fixes

- The splat-node duplicate path now reuses the existing visible-row compaction logic in `Scene::mergeSplatsWithTransforms()`.
- Duplicates of soft-deleted splat models physically contain only non-deleted rows.
- The duplicated model does not carry a deleted mask for rows that were removed from the duplicate.
- Fully soft-deleted splat nodes are treated as no-op duplicates, and their children are not duplicated or promoted to the scene root.
- GUI, Python API, and MCP duplicate paths inherit the corrected behavior because they route through `SceneManager::duplicateNodeTree()` and `Scene::duplicateNode()`.

## Root Cause

Soft deletion keeps rows in `SplatData` and hides them with the `_deleted` mask. Rendering and several copy/export paths already respect this mask, but scene graph Duplicate manually rebuilt a fresh `SplatData` from the raw tensors and did not copy or apply `_deleted`.

Because the duplicate had all raw rows and no deleted mask, previously removed splats became live and visible again.

A related edge case appears when every row in a splat is soft-deleted. The compaction clone intentionally returns no model for that node. Without a guard, duplicate recursion could still duplicate child nodes with no duplicated parent, promoting children such as crop boxes to the scene root.

## Replication Steps

1. Load a splat scene.
2. Brush-select a visible region and remove the selected splats.
3. Select the splat node in the scene graph.
4. Run Duplicate from the scene graph context menu.
5. Before this patch, the duplicate shows deleted splats again.
6. After this patch, the duplicate contains only the still-visible splats.

## Validation

- Added a regression test for `SceneManager::duplicateNodeTree()` that soft-deletes rows, duplicates the node, verifies the duplicate has only visible rows, checks non-zero SH-rest rows survive compaction, and checks undo/redo restores the compacted duplicate.
- Added a regression test for a fully soft-deleted splat with a crop box child to verify the duplicate is a no-op and the child is not orphaned.

### DIFF
--- a/src/core/scene.cpp
+++ b/src/core/scene.cpp
@@ -2422,15 +2422,10 @@ namespace lfs::core {
                 const auto* src_for_model = getNodeById(src_id);
                 if (src_for_model && src_for_model->model) {
                     const auto& model = *src_for_model->model;
-                    auto cloned = std::make_unique<lfs::core::SplatData>(
-                        model.get_max_sh_degree(),
-                        model.means_raw().clone(), model.sh0_raw().clone(),
-                        model.shN_raw().is_valid() ? model.shN_raw().clone() : lfs::core::Tensor{},
-                        model.scaling_raw().clone(), model.rotation_raw().clone(), model.opacity_raw().clone(),
-                        model.get_scene_scale(),
-                        lfs::core::SplatData::ShNLayout::Swizzled);
-                    cloned->set_active_sh_degree(model.get_active_sh_degree());
-                    new_id = addSplat(new_name, std::move(cloned), parent_id);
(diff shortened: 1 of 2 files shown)


MrNeRF:
minimal fix. Thank you!

PR #1374, opened 2026-07-04
transform tools on multiple models

### LINKED ISSUE #780: Apply transformations to multiple models
In Scene panel be able to select more than one model, so that any further editing, say rotation, applies to all selected models.

kmeirlaen:
## Summary

Fixes #780 

This patch adds multi-selection transform modes for scene graph model transforms. Multi-selected models default to Group mode, preserving the existing shared-pivot behavior, and can now be switched to Individual mode so the same rotate or scale operation is applied from each model's own transform basis.

## What this patch does

- Replaces Local/World toolbar buttons with Group/Individual buttons when multiple scene graph nodes are selected and a transform tool is active.
- Persists the multi-transform mode through the app store and Python UI runtime bridge.
- Adds localized Group Transform and Individual Transform keys for every existing locale file.
- Keeps Group mode as the default and preserves shared-pivot translate, rotate, and scale behavior.
- Adds a combined visualizer-world bounds cage for multi-selection Group scale.

## Demo

https://github.com/user-attachments/assets/1b58d5bb-bd95-4407-9486-e0984e97e32e

### DIFF
--- a/src/python/lfs/py_store.cpp
+++ b/src/python/lfs/py_store.cpp
@@ -278,6 +278,8 @@ namespace lfs::python {
                 store.transform_space.set(nb::cast<int>(value));
             else if (field == "pivot_mode")
                 store.pivot_mode.set(nb::cast<int>(value));
+            else if (field == "multi_transform_mode")
+                store.multi_transform_mode.set(nb::cast<int>(value));
             else if (field == "import_overlay_state")
                 store.import_overlay_state.set(import_overlay_state_from_object(value));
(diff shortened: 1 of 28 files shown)


kmeirlaen:
both copilot issues handled

MrNeRF:
I think the most confusing thing here is the naming because we already can create groups in the scene graph and drag and drop nodes into this group.

Should this be called differently like multi node transform? Should be reflected in code to understand why this is different.

PR #1378, opened 2026-07-05
add camera selector plugin to official repo

kmeirlaen:
Probably this shall become main funcionality. until then, a plugin to manage camara selections more easily.


MrNeRF:
This must go here so that everybody gets it: https://github.com/mrnerf/lichtfeld-studio-web
See for instance here: https://github.com/MrNeRF/lichtfeld-studio-web/commit/cd9978e72f3a7197941017fa4d807580a79adaf8

This here is legacy and should be probably cleaned up as the system seems to work reliably.

PR #1388, opened 2026-07-06
Drag/drop video files for frame extraction

kmeirlaen:
## Summary

This patch extends main-window drag/drop detection so a single supported video file opens the existing Extract Frames from Video panel. The video path is passed into the existing video extractor widget, so the current output folder default remains `<video parent>/<video stem>_frames`.

Supported drag/drop video extensions now match the video file picker: `.mp4`, `.avi`, `.mov`, `.mkv`, `.webm`, `.flv`, and `.wmv`.

Review follow-up updates keep the extractor panel visible and raised on every video drop, deduplicate the video extension list through the video I/O surface, and reset the default output folder when a different video is opened so extracted frames do not accidentally target a previous video's folder.

## Observed symptoms and replication

Before this change:

1. Open LichtFeld Studio.
2. Drag a common video file such as `clip.mp4` onto the main application window.
3. The file drop was treated as unrecognized input instead of opening the newly added Extract Frames from Video flow.

Dataset folders still opened the dataset import window, and splat formats such as `.ply`, `.sog`, `.spz`, and `.rad` still followed the loader path.

## What this patch fixes

- Adds a `ShowVideoExtractor` command carrying the dropped video path.
- Adds `IVideoExtractorWidget::openVideoPath(...) -> bool`, implemented by `VideoExtractorDialog` through the existing `openVideo(...)` method.
- Routes exactly-one-file video drops to the video extractor command.
- Leaves mixed drops unchanged unless the drop consists of exactly one recognized video file.
- Keeps the video extractor panel visible even if the video fails to open, so the existing status/error UI can show the failure.
- Raises the video extractor panel when a video is dropped, including when the panel was already enabled.
- Shares supported video extension detection between drag/drop and the native video file picker.
- Resets the default output directory for a newly opened video while preserving a manually selected output directory when reopening the same video path.

## Root cause

The video extraction panel existed, and the file picker knew which video extensions were supported, but the central drag/drop classifier only recognized splats, datasets, configs, COLMAP camera files, meshes, and environment maps. Common video extensions therefore fell through to the unrecognized-file path.

The issue showed up after adding the video extraction feature because no event/API bridge existed from generic file-drop handling into the video extractor panel.

The review follow-up root causes were two small state gaps: panel enabling did not guarantee an already-open floating extractor came to the front, and `VideoExtractorDialog::openVideo(...)` preserved an existing output directory even when switching to a different video. The file picker and drag/drop path also each carried their own video extension list, which could drift over time.

## Impact notes

- GUI: single supported video drops open and prefill the Extract Frames from Video panel.
- CLI: no behavior change.
- Python API: no behavior change; `open_video_file_dialog` is unchanged.
- MCP: no behavior change expected beyond the additional core command being present in the shared event surface if inspected.

## Tests

Build:

```text
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe" --build build-tests-release --target lichtfeld_tests -j 8'
```

Result: passed.

Focused test run:

```text
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && cd /d build-tests-release\tests && set PATH=..;.;%PATH% && lichtfeld_tests.exe --gtest_filter=InputControllerDatasetLoadTest.*:PanelRegistryAnimationDemandTest.*'
```

Result: passed, 11 tests from `InputControllerDatasetLoadTest` and `PanelRegistryAnimationDemandTest`.

Note: the test process emitted existing CUDA shutdown log errors from `PinnedMemoryAllocator` after Google Test reported success.


src/visualizer/input/input_controller.cpp
```diff
@@ -2321,10 +2332,11 @@ namespace lfs::vis {
         }
 
         if (!unrecognized_files.empty() && splat_files.empty() && !dataset_path && !environment_map_path) {
-            static constexpr auto SUPPORTED_FORMATS =
```
MrNeRF, line 2324: Why has this changed @kmeirlaen  
The naming was upper case to make clear we have constants. static pushes the vars into the bss part of the executable mapped into memory.

MrNeRF:
Thx! All is looking good and tested. But I think the marked code changes should be reverted.

PR #1389, opened 2026-07-07
add update submodule

xiaobiaodu:
I just added a ``git submodule update --init --recursive external/libvterm`` inside the code.

It is for better compiling the code. Some people don't know there's a submodule in the repository. They may miss this important submodule, so they may fail to compile.


MrNeRF:
The question is if they read and understand the mag. But I agree, it is clearer. Thank you!

PR #1416, opened 2026-07-15
fix: correct sign inversion in video rotation detection

francescofugazzi:
`av_display_rotation_get()` returns the angle in counterclockwise degrees, but the rest of LichtFeld's pipeline (rotation kernel and rotate metadata tag) uses clockwise convention. The fix negates the returned angle to align with the expected convention.
Also adds a fallback path that reads the display matrix from each decoded frame's side data when the stream-level codecpar metadata is not populated - some MOV/MP4 files only expose this data after the first frame decode.

**Changed files:**
- `src/io/video_player.cpp` : negate rotation angle + frame-level side data fallback


MrNeRF:
looks good! Thx!

PR #1419, opened 2026-07-16
Suppress CUDA failure-report flood when the runtime is unloading at process exit

MotivaCG:
After a successful headless training run, exiting the process floods the log with `LFS FAILURE REPORT` blocks, all with `cudaErrorCudartUnloading (4): driver shutting down`, coming from the allocator destructors (`SizeBucketedPool::trim_cache`, `GPUSlabAllocator::cleanup`, `PinnedMemoryAllocator::Block::~Block`).

Cause: the CUDA runtime shuts itself down before those static singletons are destroyed, so their frees can only fail with that error. 

Fix: in the `LogOnly` branch of `ensure_cuda_success`, treat `cudaErrorCudartUnloading` as benign — one debug line, clear the sticky error, no failure report. `Throw` behavior is unchanged, and `LogOnly` is only used on teardown/recovery paths, so real mid-run failures are unaffected.

Tested on Windows 11 (CUDA 12.8, driver 13.3), now training completes and saves as before, and exit is now clean.

### DIFF
--- a/src/core/cuda_error.cpp
+++ b/src/core/cuda_error.cpp
@@ -378,6 +378,22 @@ namespace lfs::core {
             finish_cuda_check(result, state, expression_copy.c_str(), message, location);
             return;
         }
+        if (result == cudaErrorCudartUnloading) {
+            // Process exit: the CUDA runtime has already shut itself down, so
+            // failed frees are expected and the OS reclaims the memory anyway.
+            // Log once instead of flooding the shutdown log.
+            static std::once_flag logged_once;
+            try {
+                std::call_once(logged_once, [] {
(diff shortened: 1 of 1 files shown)


MotivaCG:
It seems that it failed due to an error on the builder node not the code itself.

MrNeRF:
No, that's a real issue. I am working on a principled fix.

MotivaCG:
> No, that's a real issue. I am working on a principled fix.

I compiled it originally on Windows so I cannot understand why it doesn't compile there

MrNeRF:
this is fixed in a different way.

PR #1426, opened 2026-07-20
Allow distorted cameras in standard 3DGS training

DuVogel87:
## Summary

This PR allows standard 3DGS training to accept distorted/non-pinhole cameras without requiring `--gut` or `--undistort`.

When a training camera still needs its original camera model at render time, the trainer now routes that step through the existing camera-model-aware `gsplat` rasterizer instead of the FastGS pinhole path.

## What changed

- Keep rejecting unsupported ORTHO cameras without `--gut`/`--undistort`
- Stop rejecting distorted/non-pinhole cameras in regular 3DGS mode
- Detect cameras that need camera-model-aware rasterization:
  - active distortion coefficients, or
  - non-PINHOLE model while not undistorted
- Use `gsplat_rasterize_forward` for those cameras even when `--gut` is not enabled
- Keep the FastGS path for pinhole/undistorted cameras
- Avoid precomputing COLMAP undistortion on load for this path

## Motivation

Some COLMAP datasets use camera models such as `THIN_PRISM_FISHEYE`. LFS already contains camera-model-aware rasterization support in the `gsplat` path, but regular 3DGS training currently rejects distorted/non-pinhole cameras unless users enable `--gut` or undistort the dataset.

This change lets regular 3DGS training handle those distorted camera datasets by falling back to the existing camera-model-aware rasterizer only when needed.

## Notes

This PR intentionally does **not** add direct FastGS Thin-Prism-Fisheye support. That should be a separate follow-up because it touches FastGS projection/backward code and needs more focused validation.

## Validation

- `git diff --check origin/master...HEAD`
- Conflict-resolved and reviewed against current `origin/master`

Full local build was not run: existing build directories triggered a CMake glob reconfigure and vcpkg install/remove pass, so I stopped that to avoid changing the local dependency state during PR preparation.

### DIFF
--- a/src/io/formats/colmap.cpp
+++ b/src/io/formats/colmap.cpp
@@ -2357,8 +2357,6 @@ namespace lfs::io {
                 depth_path,
                 normal_path);
 
-            camera->precompute_undistortion();
-
             cameras.push_back(std::move(camera));
         }
 
--- a/src/training/trainer.cpp
+++ b/src/training/trainer.cpp
@@ -3667,12 +3667,8 @@ namespace lfs::training {
                     return std::unexpected("Training on cameras with ortho model is not supported yet.");
                 }
             } else if (!params_.optimization.undistort || !cam->is_undistort_prepared()) {
-                if (cam->radial_distortion().numel() != 0 ||
-                    cam->tangential_distortion().numel() != 0) {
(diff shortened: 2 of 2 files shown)


MrNeRF:
as explained, this is not the viable path forward. Closing.

PR #1438, opened 2026-07-25
Stabilize right panel resizing and empty viewport overlay

### LINKED ISSUE #1435: (3c878a92) UI: Missing icon on hover for horizontal resize of side panel
The blue bar that's used to resize the panel doesn't show the horizontal double arrow resize icon on hover, but only when actually resizing. Users may not be aware that we can resize it.

Image

For comparison, this vertical resize section shows the icon on hover properly.

Image

francescofugazzi:
## Right panel

Fixes #1435.

- Restore reliable cursor handling on the resize edge.
- Prevent resize-edge clicks from also starting viewport selection.
- Increase the default panel width to avoid clipping controls at startup.

## Empty viewport overlay

- Keep the dashed placeholder animation running while the scene is empty.
- Center the empty-state text correctly when text measurement is unavailable.

## Python

- Refresh committed generated Python stubs.

### DIFF
--- a/src/python/lfs/py_ui.cpp
+++ b/src/python/lfs/py_ui.cpp
@@ -2423,7 +2423,8 @@ namespace lfs::python {
         if (auto* const renderer = activeOverlayRenderer()) {
             const glm::vec2 measured =
                 renderer->measureText(text, kOverlayTextSize * scale);
-            return {measured.x, measured.y};
+            if (measured.x > 0.0f && measured.y > 0.0f)
+                return {measured.x, measured.y};
         }
         return {static_cast<float>(text.size()) * 7.0f * scale, 16.0f * scale};
     }
--- a/src/python/lfs/py_ui.cpp
+++ b/src/python/lfs/py_ui.cpp
@@ -2423,8 +2423,7 @@ namespace lfs::python {
         if (auto* const renderer = activeOverlayRenderer()) {
             const glm::vec2 measured =
                 renderer->measureText(text, kOverlayTextSize * scale);
-            if (measured.x > 0.0f && measured.y > 0.0f)
-                return {measured.x, measured.y};
+            return {measured.x, measured.y};
         }
         return {static_cast<float>(text.size()) * 7.0f * scale, 16.0f * scale};
     }
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -5774,7 +5774,6 @@ namespace lfs::vis::gui {
         bool right_panel_active_tab_live = false;
         bool right_panel_pointer_over_scene_header = false;
         bool right_panel_pointer_over_active_tab = false;
-        bool right_panel_resize_hovering = false;
         if (show_main_panel_ && !ui_hidden_) {
             LOG_TIMER_THRESHOLD("gui_render.panel_setup.rml_right_panel", 0.25);
             const float rpw = panel_layout_.getRightPanelWidth();
@@ -5792,8 +5791,7 @@ namespace lfs::vis::gui {
             rp_layout.splitter_h = splitter_h;
             right_panel_was_dirty = rml_right_panel_.needsAnimationFrame();
             const float right_panel_edge_grab_w =
-                std::max(PanelLayoutManager::SPLITTER_H * current_ui_scale_,
-                         8.0f * current_ui_scale_);
+                PanelLayoutManager::RIGHT_PANEL_RESIZE_EDGE_HALF_WIDTH * current_ui_scale_;
             const bool pointer_over_right_panel =
                 pointInRect(panel_input.mouse_x, panel_input.mouse_y,
                             rp_layout.pos, rp_layout.size);
@@ -5804,13 +5802,8 @@ namespace lfs::vis::gui {
                 panel_input.mouse_y < rp_layout.pos.y + rp_layout.size.y;
             const bool float_blocks_rp = has_floating_panels &&
                                          reg.isPositionOverFloatingPanel(panel_input.mouse_x, panel_input.mouse_y);
-            const float resize_strip_half_w = 4.0f * current_ui_scale_;
-            right_panel_resize_hovering =
-                !float_blocks_rp &&
-                panel_input.mouse_x >= rp_layout.pos.x - resize_strip_half_w &&
-                panel_input.mouse_x <= rp_layout.pos.x + resize_strip_half_w &&
-                panel_input.mouse_y >= rp_layout.pos.y &&
-                panel_input.mouse_y < rp_layout.pos.y + rp_layout.size.y;
+            right_panel_resize_edge_was_hovered_ = !float_blocks_rp &&
+                                                   pointer_over_right_panel_edge;
             constexpr float RIGHT_PANEL_PAD = 8.0f;
             const float content_x = rp_layout.pos.x + RIGHT_PANEL_PAD;
             const float content_top = screen.work_pos.y + RIGHT_PANEL_PAD;
@@ -5922,7 +5915,7 @@ namespace lfs::vis::gui {
         } else {
             right_panel_pointer_live_capture_ = false;
             right_panel_pointer_capture_region_ = RightPanelPointerRegion::None;
-            right_panel_resize_hovering = false;
+            right_panel_resize_edge_was_hovered_ = false;
         }
         if (!hasMouseButtonDown(sdl_input)) {
             right_panel_pointer_live_capture_ = false;
@@ -6408,18 +6401,6 @@ namespace lfs::vis::gui {
                 apply_cursor(panel_layout_.getCursorRequest());
                 if (auto* input_controller = viewer_->getInputController())
                     input_controller->applySplitterCursorOverride();
-                if (right_panel_resize_hovering) {
-                    static SDL_Cursor* const resize_cursor =
-                        SDL_CreateSystemCursor(SDL_SYSTEM_CURSOR_EW_RESIZE);
-                    if (resize_cursor)
-                        SDL_SetCursor(resize_cursor);
-                    right_panel_resize_cursor_active_ = true;
-                } else if (right_panel_resize_cursor_active_) {
-                    // This transition can only be from the outer edge into
-                    // the viewport. It must override any stale RmlUi cursor.
-                    SDL_SetCursor(SDL_GetDefaultCursor());
-                    right_panel_resize_cursor_active_ = false;
-                }
             }
         } else if (auto* const wm = viewer_->getWindowManager()) {
             wm->refreshResizeCursor();
@@ -7011,6 +6992,21 @@ namespace lfs::vis::gui {
         return PanelRegistry::instance().isPositionOverFloatingPanel(x, y);
     }
 
+    bool GuiManager::isPositionOverRightPanelResizeEdge(const double x, const double y) const {
+        if (!show_main_panel_ || ui_hidden_ ||
+            last_ui_layout_work_size_.x <= 0.0f || last_ui_layout_work_size_.y <= 0.0f) {
+            return false;
+        }
+
+        const float panel_x = last_ui_layout_work_pos_.x + last_ui_layout_work_size_.x -
+                              panel_layout_.getRightPanelWidth();
+        const float strip_half_w =
+            PanelLayoutManager::RIGHT_PANEL_RESIZE_EDGE_HALF_WIDTH * current_ui_scale_;
+        return x >= panel_x - strip_half_w && x <= panel_x + strip_half_w &&
+               y >= last_ui_layout_work_pos_.y &&
+               y < last_ui_layout_work_pos_.y + last_ui_layout_work_size_.y;
+    }
+
     GuiHitTestResult GuiManager::hitTestPointer(const double x, const double y) const {
         if (isCapturingInput() || isModalWindowOpen() || startup_overlay_.blocksUnderlayInput() ||
             (global_context_menu_ && global_context_menu_->isOpen())) {
@@ -7034,19 +7030,10 @@ namespace lfs::vis::gui {
             return {.blocks_pointer = true, .takes_keyboard_focus = true};
         }
 
-        // The right-panel edge is a native resize target rather than an RmlUi
-        // element. Treat the same strip used for its cursor as GUI before the
-        // press is routed, otherwise a resize started a few pixels inside the
-        // viewport can also start a viewport selection.
-        if (show_main_panel_ && !ui_hidden_) {
-            const float edge_x = viewport_layout_.pos.x + viewport_layout_.size.x;
-            const float strip_half_w = 4.0f * current_ui_scale_;
-            if (x >= edge_x - strip_half_w && x <= edge_x + strip_half_w &&
-                y >= viewport_layout_.pos.y &&
-                y < viewport_layout_.pos.y + viewport_layout_.size.y) {
-                return {.blocks_pointer = true, .takes_keyboard_focus = true};
-            }
-        }
+        // Match the resize edge geometry used by the panel before routing a
+        // press, without treating hover or wheel input as panel interaction.
+        if (isPositionOverRightPanelResizeEdge(x, y))
+            return {.blocks_mouse_button = true};
 
         if (sequencer_ui_.blocksPointer(x, y) || rml_viewport_overlay_.blocksPointer(x, y)) {
             return {.blocks_pointer = true, .takes_keyboard_focus = true};
@@ -7298,20 +7285,11 @@ namespace lfs::vis::gui {
             return true;
         }
 
-        // The external right-panel resize edge is native (not an RmlUi
-        // element). Wake a frame both while entering its strip and once more
-        // after leaving it, so its cursor cannot remain stale while the GUI is
-        // otherwise idle.
-        if (right_panel_resize_cursor_active_)
+        // Wake a frame while the pointer is over the resize edge so the panel
+        // can update its cursor request even while the GUI is otherwise idle.
+        if (right_panel_resize_edge_was_hovered_ ||
+            isPositionOverRightPanelResizeEdge(mouse_x, mouse_y)) {
             return true;
-        if (show_main_panel_ && !ui_hidden_) {
-            const float edge_x = viewport_layout_.pos.x + viewport_layout_.size.x;
-            const float strip_half_w = 4.0f * current_ui_scale_;
-            if (mouse_x >= edge_x - strip_half_w && mouse_x <= edge_x + strip_half_w &&
-                mouse_y >= viewport_layout_.pos.y &&
-                mouse_y < viewport_layout_.pos.y + viewport_layout_.size.y) {
-                return true;
-            }
         }
 
         if (!guiFocusState().want_capture_mouse && isPositionInViewport(mouse_x, mouse_y)) {
@@ -7535,13 +7513,6 @@ namespace lfs::vis::gui {
             return true;
         if (startup_overlay_.needsAnimationFrame())
             return true;
-        if (!ui_hidden_ && !drag_drop_hovering_ && !startup_overlay_.isVisible() &&
-            !async_tasks_.isImporting() && !async_tasks_.isImportCompletionShowing()) {
-            if (auto* const scene_manager = viewer_ ? viewer_->getSceneManager() : nullptr;
-                scene_manager && scene_manager->isEmpty()) {
-                return true;
-            }
-        }
         if (rml_modal_overlay_ && rml_modal_overlay_->needsAnimationFrame())
             return true;
         if (rml_toast_overlay_ && rml_toast_overlay_->needsAnimationFrame())
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -44,6 +44,7 @@
 #include "gui/icon_cache.hpp"
 #include "input/frame_input_buffer.hpp"
 #include "input/input_controller.hpp"
+#include "input/input_router.hpp"
 #include "input/sdl_key_mapping.hpp"
 #include "internal/resource_paths.hpp"
 #include "tools/align_tool.hpp"
@@ -5802,7 +5803,12 @@ namespace lfs::vis::gui {
                 panel_input.mouse_y < rp_layout.pos.y + rp_layout.size.y;
             const bool float_blocks_rp = has_floating_panels &&
                                          reg.isPositionOverFloatingPanel(panel_input.mouse_x, panel_input.mouse_y);
-            right_panel_resize_edge_was_hovered_ = !float_blocks_rp &&
+            // Keep a viewport drag captured by the viewport when it crosses
+            // the resize edge, rather than starting panel hover or resize UI.
+            const bool viewport_pointer_captured =
+                hasMouseButtonDown(sdl_input) &&
+                viewer_->getWindowManager()->inputRouter().state().pointer_capture == input::InputTarget::Viewport;
+            right_panel_resize_edge_was_hovered_ = !float_blocks_rp && !viewport_pointer_captured &&
                                                    pointer_over_right_panel_edge;
             constexpr float RIGHT_PANEL_PAD = 8.0f;
             const float content_x = rp_layout.pos.x + RIGHT_PANEL_PAD;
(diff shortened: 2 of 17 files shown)


MrNeRF (commented):
Thanks for taking this. The root cause is found correctly: no frame was rendered on plain hover, and for the pixels left of the panel edge the input controller sets the cursor back to default right after the GUI set it (`input_controller.cpp:1395`). The `ProcessMouseLeave()` call is the right fix for the stale RmlUi hover.

But the diff has some problems, please fix them before this goes in.

1. The blue hover highlight on the edge is deleted. That affordance is what #1435 is about.
2. Two of the three new strips use the wrong edge. The viewport right edge is not the panel left edge when the docked python console is open.
3. The new native SDL cursor block is not needed, `apply_cursor()` two lines above already does it.
4. `hitTestPointer` must not take the keyboard focus.
5. The empty overlay animation gate sits in the wrong layer and uses a different "is empty" than the overlay itself, so the frozen dashes are still reachable.
6. `clang-format` was not run.

Details in the line comments. Build is fine on Linux, and the stub refresh is correct (the cropbox params came in with #1365).

src/visualizer/gui/rmlui/resources/right_panel.rml
```diff
@@ -3,7 +3,6 @@
     <link type="text/rcss" href="right_panel.rcss"/>
 </head>
 <body id="rp-body" data-model="right_panel_tabs">
-    <div id="resize-handle"></div>
```
MrNeRF, line 6: Please keep the handle. This div was the only thing that ever painted this edge: `#resize-handle:hover` gave the blue strip (info @ 0.3) and `.dragging` @ 0.5. #1435 is about the user not noticing that the panel can be resized, so removing the highlight makes exactly that part worse. The left dock still has the same indicator with the same colors (`#left-dock-resize-indicator`, `viewport_overlay.theme.rcss:22-28`), so now the two edges of the same viewport behave differently.

You also don't need to delete it. The stale hover you describe is already fixed by your `ProcessMouseLeave()`. Keep the element for the visuals and decide the cursor from the geometry, like you do now.

src/visualizer/gui/rmlui/resources/right_panel.theme.rcss
```diff
@@ -67,11 +67,3 @@
 #tab-separator {
     background-color: @{right_panel.separator};
 }
-
-#resize-handle:hover {
```
MrNeRF, line 71: If the handle really has to go, then also remove `right_panel.resize_hover` and `right_panel.resize_active` in `rml_theme.cpp:546-547`. They are dead now.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7013,6 +7034,20 @@ namespace lfs::vis::gui {
             return {.blocks_pointer = true, .takes_keyboard_focus = true};
         }
 
+        // The right-panel edge is a native resize target rather than an RmlUi
+        // element. Treat the same strip used for its cursor as GUI before the
+        // press is routed, otherwise a resize started a few pixels inside the
+        // viewport can also start a viewport selection.
+        if (show_main_panel_ && !ui_hidden_) {
+            const float edge_x = viewport_layout_.pos.x + viewport_layout_.size.x;
```
MrNeRF: Wrong edge. `viewport_layout_.pos.x + size.x` is the right edge of the viewport, not the left edge of the right panel. `computeViewportWidth()` also subtracts the docked python console plus `PANEL_GAP` (`panel_layout.cpp:815-834`). So with the console docked this strip sits on the console edge and the panel edge is not covered at all.

Your cursor code in the same file uses `rp_layout.pos.x` (line 5810), so the two disagree. The y range differs too: the panel is full work height, the viewport stops above the bottom dock.

Please compute the panel edge and the strip half width in one place and use it everywhere. Right now `4.0f` is copied in `rml_right_panel.cpp:423`, `gui_manager.cpp:5807`, `7043` and `7309`, and there is already `right_panel_edge_grab_w` with 8dp at line 5794.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7013,6 +7034,20 @@ namespace lfs::vis::gui {
...
+            const float strip_half_w = 4.0f * current_ui_scale_;
+            if (x >= edge_x - strip_half_w && x <= edge_x + strip_half_w &&
+                y >= viewport_layout_.pos.y &&
+                y < viewport_layout_.pos.y + viewport_layout_.size.y) {
+                return {.blocks_pointer = true, .takes_keyboard_focus = true};
```
MrNeRF: `takes_keyboard_focus = true` is wrong here. Grabbing a resize edge should not blur the viewport, that kills the camera keys. Return `.blocks_pointer = true` only.

Also note `blocks_pointer` is used for hover as well (`input_router.cpp:89-99`), so this eats the mouse wheel zoom in a 4px band at the viewport edge.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7263,6 +7298,22 @@ namespace lfs::vis::gui {
...
         }
 
+        // The external right-panel resize edge is native (not an RmlUi
+        // element). Wake a frame both while entering its strip and once more
+        // after leaving it, so its cursor cannot remain stale while the GUI is
+        // otherwise idle.
+        if (right_panel_resize_cursor_active_)
+            return true;
+        if (show_main_panel_ && !ui_hidden_) {
+            const float edge_x = viewport_layout_.pos.x + viewport_layout_.size.x;
```
MrNeRF: Same wrong edge as in `hitTestPointer`. With the python console docked no frame is woken at the real panel edge, so #1435 stays broken in that layout.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -6399,6 +6408,18 @@ namespace lfs::vis::gui {
                 apply_cursor(panel_layout_.getCursorRequest());
                 if (auto* input_controller = viewer_->getInputController())
                     input_controller->applySplitterCursorOverride();
+                if (right_panel_resize_hovering) {
+                    static SDL_Cursor* const resize_cursor =
```
MrNeRF: This block is not needed. Five lines above, `apply_cursor(rml_right_panel_.getCursorRequest())` already sets the EW cursor, and `RmlRightPanel` requests `ResizeEW` for the same +-4dp strip (`rml_right_panel.cpp:500`). This is now the third EW cursor static in this file (`2928`, `6094`, `6412`), all leaked. What really fixes the bug is the frame wake and the hover routing, not this.

If you keep it anyway, two things are wrong:
- The comment. Leaving the strip to the right goes into the panel body, not into the viewport. Then this forces the default cursor for one frame over a text field or the splitter.
- If a floating panel takes the cursor, the whole block is skipped (line 6404) and `right_panel_resize_cursor_active_` stays true. After that every passive mouse move costs a full frame until we get here again.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7484,6 +7535,13 @@ namespace lfs::vis::gui {
             return true;
         if (startup_overlay_.needsAnimationFrame())
             return true;
+        if (!ui_hidden_ && !drag_drop_hovering_ && !startup_overlay_.isVisible() &&
```
MrNeRF: The animation is really needed (the dashes use `lf.ui.get_time()`, `overlays/__init__.py:353`), but this is the wrong layer and the condition does not match the overlay:

- The overlay draws when `lf.ui.is_scene_empty()` is true, that is `EditorContext::isEmpty()` (mode == EMPTY). Here you use `SceneManager::isEmpty()`, which also requires `content_type_ == Empty`. Delete all nodes without `clearScene()` and `content_type_` stays `SplatFiles`: the overlay draws, gets no frames, dashes freeze. That is the bug you want to fix.
- Python also returns early below a 200x200 viewport (`overlays/__init__.py:329`) and does not check `ui_hidden`. Here it is the other way round.
- No throttle. Everything else in this function has a deadline (`isVramHudPublishDue` 250ms, the tooltip wake). Unconditional `true` means we never idle while no scene is loaded.

Cleaner: call `lf.ui.request_redraw()` at the end of `_draw_empty_state_overlay` (binding in `py_ui.cpp:2796`). Then the visibility logic stays in one place and cannot drift.

src/python/lfs/py_ui.cpp
```diff
@@ -2423,7 +2423,8 @@ namespace lfs::python {
         if (auto* const renderer = activeOverlayRenderer()) {
             const glm::vec2 measured =
                 renderer->measureText(text, kOverlayTextSize * scale);
-            return {measured.x, measured.y};
+            if (measured.x > 0.0f && measured.y > 0.0f)
```
MrNeRF: Please fix this one layer lower. `measureText` returns (0,0) only when no measure function is installed (`screen_overlay_renderer.cpp:217-222`, installed in `gui_manager.cpp:3194`, cleared at `3827`). Put the fallback into `ScreenOverlayRenderer::measureText`, then the C++ caller in `gui_manager.cpp:6891` gets it too.

Small detail: for an empty string `measured.x` is 0, so `calc_text_size("")` now returns height `16 * scale` instead of 0.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -429,23 +425,36 @@ namespace lfs::vis::gui {
...
             mx >= -resize_handle_half_w &&
             mx <= resize_handle_half_w &&
             my >= 0.0f &&
             my <= layout.size.y;
-        const bool over_resize_handle =
-            over_resize_handle_geom || (hover && isOrHasAncestor(hover, "resize-handle"));
+        // The handle straddles the left edge of this context. Do not rely on
+        // RmlUi hover state here: it can remain stale after the pointer leaves
+        // the context into the viewport.
+        const bool over_resize_handle = over_resize_handle_geom;
```
MrNeRF: The names and the comment talk about a handle that does not exist anymore (`resize_handle_half_w`, `over_resize_handle_geom`, "the handle straddles the left edge"). Either keep the element or rename this.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -429,23 +425,36 @@ namespace lfs::vis::gui {
...
         const bool over_splitter = hover && isOrHasAncestor(hover, "splitter");
         const bool over_interactive = hover && hover->GetTagName() != "body" &&
-                                      hover->GetId() != "rp-body" &&
-                                      hover->GetId() != "left-border" &&
-                                      hover->GetId() != "tab-separator";
-        const bool over_resize_control = over_resize_handle || over_splitter;
+                                       hover->GetId() != "rp-body" &&
```
MrNeRF: `clang-format` was not run, these continuation lines got one space too much. Please run it over the changed files.

src/visualizer/gui/panel_layout.hpp
```diff
@@ -161,7 +161,7 @@ namespace lfs::vis::gui {
         [[nodiscard]] float maxRightPanelWidth(bool show_main_panel, bool ui_hidden,
                                                const ScreenState& screen) const;
 
-        float right_panel_width_ = 340.0f;
+        float right_panel_width_ = 360.0f;
```
MrNeRF, line 166: Fine for me. Just so you know: the width is never persisted (`panel_layout.cpp:34-46`, always starts at the default), and no test covers it.

francescofugazzi:
> Thanks for taking this. The root cause is found correctly: no frame was rendered on plain hover, and for the pixels left of the panel edge the input controller sets the cursor back to default right after the GUI set it (`input_controller.cpp:1395`). The `ProcessMouseLeave()` call is the right fix for the stale RmlUi hover.
> 
> But the diff has some problems, please fix them before this goes in.
> 
> 1. The blue hover highlight on the edge is deleted. That affordance is what #1435 is about.
> 2. Two of the three new strips use the wrong edge. The viewport right edge is not the panel left edge when the docked python console is open.
> 3. The new native SDL cursor block is not needed, `apply_cursor()` two lines above already does it.
> 4. `hitTestPointer` must not take the keyboard focus.
> 5. The empty overlay animation gate sits in the wrong layer and uses a different "is empty" than the overlay itself, so the frozen dashes are still reachable.
> 6. `clang-format` was not run.
> 
> Details in the line comments. Build is fine on Linux, and the stub refresh is correct (the cropbox params came in with #1365).

- Restored the visible resize handle and its hover/drag affordance.
- Centralized the right-panel resize-edge geometry and now use the actual panel left edge across the full work area.
- Removed the redundant native SDL cursor override.
- Kept resize presses from changing keyboard focus or starting viewport selection while preserving viewport hover and wheel input.
- Moved empty-overlay animation ownership back to the overlay itself and fixed the redraw handoff so the request survives until the next frame.
- Moved the text-measure fallback into `ScreenOverlayRenderer`, including the empty-string case.
- Ran clang-format on all changed C/C++ files.

I also manually verified the empty overlay animation, cursor enter/leave behavior, sidebar resizing, and resize-edge input routing.

src/visualizer/gui/rmlui/resources/right_panel.rml
francescofugazzi: Restored the `#resize-handle` element and its visual role.

Cursor selection still uses the resize-edge geometry, while the element is retained for the hover and dragging affordance. `ProcessMouseLeave()` remains in place to clear stale RmlUi hover state when the pointer leaves the panel context.

src/visualizer/gui/rmlui/resources/right_panel.theme.rcss
francescofugazzi: Restored the hover and dragging rules, so `right_panel.resize_hover` and `right_panel.resize_active` are live again.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -424,28 +425,44 @@ namespace lfs::vis::gui {
...
             mx >= -resize_handle_half_w &&
             mx <= resize_handle_half_w &&
             my >= 0.0f &&
             my <= layout.size.y;
+        // Use geometry for the resize edge so stale RmlUi hover state cannot
+        // keep the cursor active after the pointer leaves the panel.
         const bool over_resize_handle =
             over_resize_handle_geom || (hover && isOrHasAncestor(hover, "resize-handle"));
+        if (resize_handle_el_ && !resize_dragging_)
+            resize_handle_el_->SetAttribute("class", over_resize_handle ? "hover" : "");
```
MrNeRF, line 462: This brought the stale hover back, only from the other side now.

The class is written from geometry, but nothing sets `input_dirty_`, so `needs_render` in `render()` stays false and the cached FBO is not refreshed. When the pointer leaves the edge through its left half (`mx` in `[-4dp, 0)`) the class is cleared in memory while the blue strip stays painted.

Measured on the strip (24x120 crop at the panel edge, mean RGB):

```
pointer far in the viewport    23.6 / 25.0 / 29.4   clean
pointer at edge + 2px          25.1 / 30.6 / 39.4   lit (correct)
pointer at edge - 3px          25.1 / 30.6 / 39.4   lit (correct)
pointer 70px into viewport     25.1 / 30.6 / 39.4   still lit
pointer 590px into viewport    25.1 / 30.6 / 39.4   still lit
```

Frames were flowing the whole time (the empty overlay dashes kept moving), so this is the panel cache, not a missing wake. It only clears once the pointer goes back into the panel body.

Please write only on change and mark the panel dirty:

```cpp
if (resize_handle_el_ && !resize_dragging_ && over_resize_handle != last_over_resize_handle_) {
    resize_handle_el_->SetAttribute("class", over_resize_handle ? "hover" : "");
    last_over_resize_handle_ = over_resize_handle;
    input_dirty_ = true;
}
```

`RmlViewportOverlay::setLeftDockResizeIndicator` does exactly this for the left dock (early out when nothing changed, then `markRenderNeeded`). It also gets rid of the `SetAttribute` on every mouse move, which dirties the style definition each time.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -424,28 +425,44 @@ namespace lfs::vis::gui {
...
+            rml_pointer_inside_ = false;
+            last_over_interactive_ = false;
+            input_dirty_ = true;
+        }
+        auto* hover = pointer_in_context ? rml_context_->GetHoverElement() : nullptr;
         const bool over_resize_handle_geom =
             mx >= -resize_handle_half_w &&
             mx <= resize_handle_half_w &&
             my >= 0.0f &&
             my <= layout.size.y;
```
MrNeRF: `my <= layout.size.y` is inclusive here, `isPositionOverRightPanelResizeEdge` uses `y < ...`. One pixel at the bottom, no drama, but you may as well make them match while you are in there.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -5791,8 +5791,7 @@ namespace lfs::vis::gui {
             rp_layout.splitter_h = splitter_h;
             right_panel_was_dirty = rml_right_panel_.needsAnimationFrame();
             const float right_panel_edge_grab_w =
-                std::max(PanelLayoutManager::SPLITTER_H * current_ui_scale_,
-                         8.0f * current_ui_scale_);
+                PanelLayoutManager::RIGHT_PANEL_RESIZE_EDGE_HALF_WIDTH * current_ui_scale_;
```
MrNeRF, line 5795: Only so it is a conscious decision: this band was `max(SPLITTER_H, 8dp)` = 8dp and is 4dp now, so half as wide. It matches the visible handle, which is what you want for the cursor, but it also feeds `pointer_targets_right_panel` and the live capture further down (line 5860). Fine by me, I just did not ask for it, so I want to be sure it was intended.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7013,6 +7030,11 @@ namespace lfs::vis::gui {
             return {.blocks_pointer = true, .takes_keyboard_focus = true};
         }
 
+        // Match the resize edge geometry used by the panel before routing a
+        // press, without treating hover or wheel input as panel interaction.
+        if (isPositionOverRightPanelResizeEdge(x, y))
+            return {.blocks_mouse_button = true};
```
MrNeRF, line 7047: Small ordering detail: this returns before the `sequencer_ui_` and `rml_viewport_overlay_` checks. Where the strip overlaps the right border of the docked python console (`PANEL_GAP` is 2dp and the strip reaches 4dp to the left), a press there no longer takes GUI keyboard focus. That is a 2dp sliver, not worth restructuring the function for, just be aware of it.

src/visualizer/gui/panel_layout.hpp
```diff
@@ -142,6 +142,7 @@ namespace lfs::vis::gui {
         static constexpr float TAB_BAR_H = 28.0f;
         static constexpr float STATUS_BAR_HEIGHT = 22.0f;
         static constexpr float PANEL_GAP = 2.0f;
+        static constexpr float RIGHT_PANEL_RESIZE_EDGE_HALF_WIDTH = 4.0f;
```
MrNeRF, line 146: Thanks for the single constant. `right_panel.rcss` carries the same number twice (`left: -4dp`, `width: 8dp`) and cannot include this header, so a one-line comment here pointing at the rcss rule would keep the two from drifting apart later.

MrNeRF (commented):
Thanks, this is a lot better. Everything from the last round is addressed, and in the right place:

- the handle and its hover/drag colors are back, `right_panel.resize_hover` and `right_panel.resize_active` are live again
- one constant for the strip, and `isPositionOverRightPanelResizeEdge` anchors on `work_pos.x + work_size.x - getRightPanelWidth()`, which is the same value as `rp_layout.pos.x`, so the docked python console no longer breaks it
- `blocks_mouse_button` is a good solution: the press is captured, the keyboard focus stays with the viewport, and hover/wheel still route to the viewport
- the redundant SDL cursor block is gone
- the overlay owns its own animation now, behind its own early returns, and the non-consuming `collectFrameDemand` at the end of the loop keeps the request from being swallowed
- text measure fallback sits in `ScreenOverlayRenderer` including the empty string
- clang-format is clean

I built it and drove the edge with the mouse: dragging from the viewport side works (100px drag moved the edge exactly 100px), the dashes animate, no validation errors.

One thing came back though, in the new class write - the stale highlight, just from the other side now. Details in the line comment. The rest are nits, take them or leave them.

francescofugazzi:
- The resize-handle class is now updated only when its state changes, and each transition invalidates the right-panel cache. The blue strip now appears immediately from either side and clears when leaving the edge.
- Aligned the Rml-side bottom boundary with the shared GUI hit test (`<`).
- The resize strip is intentionally `±4dp` / `8dp` total, matching the visible RCSS handle (`left: -4dp; width: 8dp`) instead of the previous invisible `16dp` logical band.
- Moved the sequencer/viewport-overlay checks before the button-only resize hit result, so the small overlap with the docked Python Console keeps its normal GUI focus behavior.
- Added a comment next to the shared C++ constant pointing to the matching RCSS geometry.

While testing, I also fixed a related visual issue: a viewport rectangle-selection drag crossing the panel edge could show the resize cursor and blue handle. An active viewport pointer capture now keeps ownership until mouse release, so the selection drag continues without showing resize UI.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -5801,6 +5801,15 @@ namespace lfs::vis::gui {
                 panel_input.mouse_x <= rp_layout.pos.x + right_panel_edge_grab_w &&
                 panel_input.mouse_y >= rp_layout.pos.y &&
                 panel_input.mouse_y < rp_layout.pos.y + rp_layout.size.y;
+            const bool float_blocks_rp = has_floating_panels &&
+                                         reg.isPositionOverFloatingPanel(panel_input.mouse_x, panel_input.mouse_y);
+            // Keep a viewport drag captured by the viewport when it crosses
+            // the resize edge, rather than starting panel hover or resize UI.
+            const bool viewport_pointer_captured =
+                hasMouseButtonDown(sdl_input) &&
+                viewer_->getWindowManager()->inputRouter().state().pointer_capture == input::InputTarget::Viewport;
```
MrNeRF: Please do not dereference `getWindowManager()` unchecked here. `render()` already fetches it safely in its first line (`auto* window_manager = viewer_ ? viewer_->getWindowManager() : nullptr;`), and 250 lines above this one the same call is guarded again (`if (auto* const wm = viewer_->getWindowManager())`), so the codebase clearly treats it as nullable - it is null once `vulkan_gui_` is false, e.g. after teardown.

Just use the local that is already in scope:

```cpp
const bool viewport_pointer_captured =
    hasMouseButtonDown(sdl_input) && window_manager &&
    window_manager->inputRouter().state().pointer_capture == input::InputTarget::Viewport;
```

src/visualizer/gui/rml_right_panel.hpp
```diff
@@ -104,6 +104,7 @@ namespace lfs::vis::gui {
         bool splitter_dragging_ = false;
 
         bool resize_dragging_ = false;
+        bool last_over_resize_handle_ = false;
```
MrNeRF, line 107: `reloadResources()` resets `last_over_interactive_` and `rml_pointer_inside_` but not this one, and it reloads the document, so `resize_handle_el_` comes back without any class while the latch still says `true`. Reload the theme while the pointer sits on the edge and the highlight stays missing until the pointer leaves and comes back. One line next to `last_over_interactive_ = false;` fixes it.

MrNeRF (commented):
Nice, this is it. I built `dbf06a622` and went through the edge again with the mouse:

- the strip lights up from both sides now, including a direct approach from the viewport, and it clears as soon as the pointer leaves the edge - the stuck highlight from the last round is gone (measured on the 6px handle column: `36.5/38.2/45.1` idle vs `42.4/60.7/84.9` lit, and it goes back to idle at every exit path I tried)
- resize drag started from the viewport side still works after the reorder: 100px drag moved the edge exactly 100px, and the `.dragging` color shows during the drag
- the viewport capture guard does what you describe: pressing in the viewport and dragging across the edge into the panel keeps the strip dark the whole way, the camera keeps rotating, and the panel width does not change
- panel clicks (tab switch), the empty overlay animation and the startup are all unaffected, no errors in the log, clang-format clean, build clean

Good catch on the selection drag, that one was mine to miss.

Two small things left, both one-liners, then this is good to merge from my side.

francescofugazzi:
Updated both points:

- `viewport_pointer_captured` now uses the existing nullable `window_manager` local, so it stays safe after Vulkan GUI teardown.
- `reloadResources()` now resets `last_over_resize_handle_` together with the other input latches, keeping it in sync with the recreated resize-handle element.

MrNeRF:
Thank you!

PR #1441, opened 2026-07-26
Stabilize headless execution and plugin validation

### LINKED ISSUE #1439: --python-script always fails in headless mode: plugin preload starts async on main thread and run_scripts aborts with 'Plugins are still loading'
## Summary

In headless mode, `--python-script` can never work: trainer initialization always fails with `Plugins are still loading`. The failure is deterministic because the plugin preload is started asynchronously on the main thread (which headless mode treats as the graphics thread) and `run_scripts()` gives up immediately instead of waiting.

## Environment

- Version: v0.5.3-144-gdee66c79 (built from source at commit `dee66c79`)
- OS: Windows 11 (the logic is platform-independent, so this should reproduce on Linux as well)

## Reproduction

```
LichtFeld-Studio.exe --headless --train -d <dataset> -o <output> --python-script foo.py
```

Any script reproduces it, even a trivial one-liner. Training aborts during trainer initialization with:

```
Failed to initialize trainer: Failed to run Python scripts: ... [Unavailable/Python] Plugins are still loading
```

## Root cause (traced at dee66c79)

1. `register_ui()` in `src/python/lfs/py_ui.cpp:2778` calls `set_graphics_thread_id(std::this_thread::get_id())` when the `lichtfeld` module is registered. In headless mode this happens on the **main thread**, so from then on the main thread is considered the graphics thread.
2. Trainer initialization reaches the configured scripts: `src/training/trainer.cpp:3005` calls `lfs::python::run_scripts()`, which calls `ensure_plugins_loaded()` (`src/python/runner.cpp:1313`).
3. In `ensure_plugins_loaded()`, the preload state is `NotStarted` and `on_graphics_thread()` is true, so it takes the async branch (`runner.cpp:1326-1328`): it calls `start_plugin_preload_worker()` and immediately returns `are_plugins_loaded()` — which is `false` because the worker has only just been started.
4. `run_scripts()` then returns the `Unavailable/Python` error `"Plugins are still loading"` (`runner.cpp:1684-1691`), and trainer initialization aborts (`trainer.cpp:3006-3010`).

The "start async and return without blocking" behavior makes sense for the GUI render thread, but in headless mode the main thread is not pumping frames and there is no later retry — so the error is guaranteed, and there is no way to use `--python-script` in headless mode at all.

Notably, `ensure_plugins_loaded()` already contains a synchronous wait path (`runner.cpp:1351-1358`, cv.wait until the preload reaches a terminal state), but it is unreachable here because the main thread is flagged as the graphics thread.

## Expected behavior

In headless mode, `ensure_plugins_loaded()` should wait synchronously for the plugin preload to finish before running the CLI-provided scripts. Possible fixes:

- Don't mark the main thread as the graphics thread in headless mode (there is no render loop to protect), so the existing synchronous wait path is taken; or
- Add an explicit headless/CLI flag that forces the synchronous path in `ensure_plugins_loaded()`.

## Related observation (GUI mode)

`--python-script` also does not reach the trainer in GUI mode: `set_python_scripts()` is only called in the headless code paths of `src/app/application.cpp` (lines 149, 283, 328). When a trainer is created through the GUI dataset-load / `--train` flow, the CLI-provided scripts are never applied to it. Happy to file that separately if preferred.

### LINKED ISSUE #1440: 'plugin check' crashes with UnicodeDecodeError on non-UTF-8 Windows locales (validator scans .venv with locale-encoded read_text)
## Summary

`LichtFeld-Studio.exe plugin check <name>` crashes with `UnicodeDecodeError` on Windows systems whose ANSI code page is not UTF-8 (e.g., Japanese cp932), whenever the plugin has a populated `.venv`. The validator recursively scans every `.py` file inside `.venv` (thousands of files for numpy/scipy) and reads them with the locale's default encoding.

## Environment

- Version: v0.5.3-144-gdee66c79 (built from source at commit `dee66c79`)
- OS: Windows 11, Japanese locale (ANSI code page cp932)
- Should affect any non-UTF-8 system locale (cp932, cp936, cp949, cp1250, ...)

## Reproduction

On Japanese Windows (or any non-UTF-8 locale), run `plugin check` against a plugin whose `.venv` contains packages such as numpy or scipy:

```
LichtFeld-Studio.exe plugin check <plugin-name>
```

Result:

```
UnicodeDecodeError: 'cp932' codec can't decode byte 0x... in position ...
```

## Root cause (`src/python/lfs_plugins/validator.py` at dee66c79)

- `_check_panel_assets()` iterates `plugin_dir.rglob("*.py")` (line 111) with **no `.venv` exclusion**, so it walks every `.py` file of every installed dependency inside the plugin's `.venv`.
- Line 116 calls `py_file.read_text()` **without an encoding argument**, which uses the locale preferred encoding — cp932 on Japanese Windows. Many third-party sources (numpy, scipy, ...) are UTF-8 files containing non-ASCII characters, so the decode fails and the whole `plugin check` command crashes.
- The other `read_text()` calls in the same file (lines 37, 67, 85, 223) have the same latent issue for plugin-authored files, but the `.venv` scan is what makes the crash effectively guaranteed.

## Suggested fix

1. Exclude `.venv` from the scan, the same way `manager.py` already does at line 738:

   ```python
   for py_file in plugin_dir.rglob("*.py"):
       if ".venv" in py_file.parts:
           continue
   ```

   This also makes `plugin check` dramatically faster (it currently AST-parses the entire dependency tree).

2. Read files with an explicit encoding for robustness:

   ```python
   py_file.read_text(encoding="utf-8", errors="replace")
   ```

   (Same for the other `read_text()` call sites in validator.py.)

## Workaround

`set PYTHONUTF8=1` before running makes the command succeed.

francescofugazzi:
## Headless Python scripts

Fixes #1439.

- Wait for startup plugin loading before running CLI-provided Python scripts in direct headless training.
- Preserve the existing non-blocking plugin preload behavior for the GUI.
- Leave the TCP headless path unchanged.

## Plugin validation

Fixes #1440.

- Read validator inputs explicitly as UTF-8, independently of the Windows system code page.
- Exclude `.venv` sources from panel asset checks, avoiding dependency-tree scans and locale-dependent decode failures.
- Add regression coverage for a virtual-environment dependency containing UTF-8 source.

## Headless CUDA teardown

- Complete CUDA memory-pool cleanup before process exit in headless training.
- Prevent late cache, slab, and event-pool destruction from calling CUDA after CUDART begins unloading.

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -284,6 +284,11 @@ namespace lfs::app {
                         vis::gui::panels::PythonScriptManagerState::getInstance().setScripts(params->python_scripts);
                     }
 
+                    if (!params->python_scripts.empty() && !python::ensure_plugins_loaded(true)) {
+                        LOG_ERROR("Failed to load plugins before running headless Python scripts");
+                        return 1;
+                    }
+
                     if (const auto result = trainer->initialize(*ckpt_params_result); !result) {
                         LOG_ERROR("Failed to initialize trainer: {}", result.error());
                         return 1;
@@ -327,6 +332,11 @@ namespace lfs::app {
                     if (!params->python_scripts.empty()) {
                         trainer->set_python_scripts(params->python_scripts);
                         vis::gui::panels::PythonScriptManagerState::getInstance().setScripts(params->python_scripts);
+                    }
+
+                    if (!params->python_scripts.empty() && !python::ensure_plugins_loaded(true)) {
+                        LOG_ERROR("Failed to load plugins before running headless Python scripts");
+                        return 1;
                     }
 
                     if (const auto result = trainer->initialize(*params); !result) {
--- a/src/python/lfs_plugins/validator.py
+++ b/src/python/lfs_plugins/validator.py
@@ -21,6 +21,10 @@ except ImportError:
     import tomli as tomllib
 
 
+def _read_text(path: Path) -> str:
+    return path.read_text(encoding="utf-8", errors="replace")
+
+
 def validate_plugin(plugin_path: str | Path) -> list[str]:
     """Validate plugin structure and manifest. Returns list of errors (empty if valid)."""
     plugin_dir = Path(plugin_path)
@@ -34,7 +38,7 @@ def validate_plugin(plugin_path: str | Path) -> list[str]:
         errors.append("Missing pyproject.toml")
     else:
         try:
-            data = tomllib.loads(manifest.read_text())
+            data = tomllib.loads(_read_text(manifest))
             lf = data.get("tool", {}).get("lichtfeld", {})
             if not lf:
                 errors.append("pyproject.toml: missing [tool.lichtfeld] section")
@@ -64,7 +68,7 @@ def validate_plugin(plugin_path: str | Path) -> list[str]:
     if not init_py.exists():
         errors.append("Missing __init__.py")
     else:
-        content = init_py.read_text()
+        content = _read_text(init_py)
         if "def on_load" not in content:
             errors.append("Missing on_load() function")
         if "def on_unload" not in content:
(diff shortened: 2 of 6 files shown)


MrNeRF:
Looks good! Thank you!

PR #1443, opened 2026-07-26
Preserve training iterations when resetting the same dataset

### LINKED ISSUE #1436: (3c878a92) Training, UI/UX: Iterations count resets to 36,000 after resetting a training session
When stopping and resetting a training session, the "iterations count" gets reset to the default 36,000 instead of staying at what the user entered before starting the previous session they just stopped.
All other training parameters are not reset, only this one, which is inconsistent and frustrating since we need to enter our desired number again.

francescofugazzi:
## Summary

Preserves the auto-scale marker when resetting a training session on the same dataset.

This prevents camera-count step scaling from being applied again after reset and overwriting an explicitly chosen iteration count. Loading a dataset from a different path still resets the marker and applies automatic scaling normally.

Fixes #1436.


MrNeRF:
Lgtm. Thanks!

PR #1462, opened 2026-07-27
Video Frame Extractor Source Frame Numbering

kmeirlaen:
## Summary

This patch changes video frame extractor output naming so the numeric token in the filename pattern represents the 1-based source video frame number, not the sequence number of the extracted output image.

For the default `frame_%d` pattern, extracting every other frame from a five-frame source now writes names such as `frame_1.png`, `frame_3.png`, and `frame_5.png`. A trimmed extraction keeps numbering relative to the original video, so a trim that starts at source frame 3 writes `frame_3.png` rather than restarting at `frame_1.png`.

## Observed Symptoms

The extractor's default filename pattern is `frame_%d`. Users reasonably interpret `%d` as the actual video frame number, but the implementation passed extraction counters into the formatter. As a result, skipped frames and trimmed ranges produced compact output sequences that did not match the source video frame numbers.

Example replication steps for the original issue:

1. Open the Extract Frames from Video panel.
2. Use the default `frame_%d` filename pattern.
3. Extract every 2nd frame from a video.
4. Observe that the output files are named `frame_1`, `frame_2`, `frame_3`, even though they correspond to source frames 1, 3, and 5.
5. Repeat with a trimmed range starting later in the video and observe that numbering restarts at 1 instead of preserving the original source frame index.

## Root Cause

`formatFrameFilenameStem()` already formats the number it receives correctly. The bug was in the extraction write paths: they passed `saved_count + 1` or `written_count + 1`, which are output/extraction counters, while the extractor already tracked the selected source frame as `current_src_frame` and stored it in metadata as `source_frame`.

Sharpness window mode had the same issue through `written_count + 1`, even though the winning candidate already carried its `source_frame`.

## What This Patch Fixes

- Normal hardware decode writes now generate filenames from `current_src_frame`.
- Normal software decode writes now generate filenames from `current_src_frame`.
- Sharpness window writes now generate filenames from the selected candidate's `source_frame`.
- The `VideoFrameExtractor::Params::filename_pattern` comment now documents that `%d` means the 1-based source video frame number.
- Added deployed-path coverage that creates a tiny encoded video through the existing video encoder, runs the real extractor, and checks full interval and trimmed-range filename side effects together with the extractor return contract.

## Impact Check

- GUI: The existing video extractor dialog passes the filename pattern through unchanged, so the GUI inherits the corrected naming behavior. No new user-facing strings were added.
- CLI: No direct CLI extractor surface was found in the inspected code.
- Python API/plugin: The Python file menu operator only opens the native video extractor panel, so it inherits the GUI behavior without API changes.
- MCP: No direct video frame extraction MCP tool was found in the inspected app/MCP surfaces.
- Metadata: Existing metadata `source_frame` values are unchanged; filenames now align with those values.

### DIFF
--- a/src/io/video_frame_extractor.cpp
+++ b/src/io/video_frame_extractor.cpp
@@ -36,6 +36,7 @@ extern "C" {
 #include <limits>
 #include <stdexcept>
 #include <thread>
+#include <unordered_set>
 #include <vector>
 
 namespace lfs::io {
@@ -1328,6 +1329,7 @@ namespace lfs::io {
                     double sharpness_score;
                 };
                 std::vector<FrameSaveInfo> saved_frames;
+                std::unordered_set<std::filesystem::path> emitted_filenames;
 
                 const bool seek_timestamps_available =
                     std::isfinite(time_base) && time_base > 0.0 && std::isfinite(video_duration) &&
@@ -1395,6 +1397,29 @@ namespace lfs::io {
                     return params.output_dir / (formatFrameFilenameStem(params.filename_pattern, frame_num) + ext);
                 };
 
+                const auto source_frame_for_time = [&](const double frame_time) {
+                    return std::max(
+                        1, static_cast<int>(std::llround(frame_time * video_fps)) + 1);
+                };
+
+                auto reserve_output_filename = [&](const std::filesystem::path& filename,
+                                                   const int source_frame) {
+                    if (emitted_filenames.insert(filename).second)
+                        return true;
+                    LOG_WARN("Skipping duplicate source frame {} output: {}",
+                             source_frame, lfs::core::path_to_utf8(filename));
+                    return false;
+                };
+
+                auto finish_selected_frame = [&]() {
+                    saved_count++;
+
+                    if (params.progress_callback) {
+                        params.progress_callback(saved_count + skipped_count, estimated_total, skipped_count);
+                    }
+                    throw_if_cancelled();
+                };
+
                 auto should_extract_frame = [&](const double frame_time) {
                     bool should_extract = false;
                     if (params.mode == ExtractionMode::FPS) {
@@ -1420,6 +1445,12 @@ namespace lfs::io {
                             return a.score < b.score;
                         });
                     std::filesystem::path fname = generate_filename(best->source_frame);
+                    if (!reserve_output_filename(fname, best->source_frame)) {
+                        finish_selected_frame();
+                        window_candidates.clear();
+                        window_skip_counter = 0;
+                        return;
+                    }
                     // Apply rotation to the best window frame before writing
                     int write_w = out_width;
                     int write_h = out_height;
@@ -1467,12 +1498,9 @@ namespace lfs::io {
                                                     best->score});
                         }
                     }
-                    ++saved_count;
-                    if (params.progress_callback)
-                        params.progress_callback(saved_count, estimated_total, skipped_count);
+                    finish_selected_frame();
                     window_candidates.clear();
                     window_skip_counter = 0;
-                    throw_if_cancelled();
                 };
 
                 auto process_frame_hw = [&](AVFrame* hw_frame) {
@@ -1529,6 +1557,11 @@ namespace lfs::io {
                                     params.progress_callback(saved_count + skipped_count, estimated_total, skipped_count);
                                 return;
                             }
+                        }
+
+                        if (!reserve_output_filename(filename, current_src_frame)) {
+                            finish_selected_frame();
+                            return;
                         }
 
                         const int rot = params.rotation;
@@ -1614,6 +1647,11 @@ namespace lfs::io {
                             }
                         }
                         // --- End sharpness ---
+
+                        if (!reserve_output_filename(filename, current_src_frame)) {
+                            finish_selected_frame();
+                            return;
+                        }
 
                         // --- Rotation (hybrid HW path) ---
                         int hw_rot_w = out_width;
@@ -1729,6 +1767,12 @@ namespace lfs::io {
                     }
                     // --- End sharpness ---
 
+                    std::filesystem::path filename = generate_filename(current_src_frame);
+                    if (!reserve_output_filename(filename, current_src_frame)) {
+                        finish_selected_frame();
+                        return;
+                    }
+
                     // --- Rotation (SW path) ---
                     int sw_rot_w = out_width;
                     int sw_rot_h = out_height;
@@ -1763,8 +1807,6 @@ namespace lfs::io {
                                     static_cast<size_t>(out_width) * out_height * 3);
                     }
                     // --- End rotation ---
-
-                    std::filesystem::path filename = generate_filename(current_src_frame);
 
                     if (gpu_encoding_enabled) {
                         if (batch_encode_w == 0) {
@@ -1801,12 +1843,7 @@ namespace lfs::io {
                         LOG_WARN("Failed to write extracted frame: {}", lfs::core::path_to_utf8(filename));
                     }
 
-                    saved_count++;
-
-                    if (params.progress_callback) {
-                        params.progress_callback(saved_count + skipped_count, estimated_total, skipped_count);
-                    }
-                    throw_if_cancelled();
+                    finish_selected_frame();
                 };
 
                 if (params.mode == ExtractionMode::FPS) {
@@ -1838,8 +1875,7 @@ namespace lfs::io {
                 const auto process_selected_frame = [&](AVFrame* const selected_frame,
                                                         const double frame_time) {
                     current_frame_time = frame_time;
-                    current_src_frame = std::max(
-                        1, static_cast<int>(std::llround(frame_time * video_fps)) + 1);
+                    current_src_frame = source_frame_for_time(frame_time);
                     if (using_hw_decode)
                         process_frame_hw(selected_frame);
                     else
@@ -1869,7 +1905,7 @@ namespace lfs::io {
                     }
 
                     current_frame_time = frame_time;
-                    current_src_frame = decoded_frame_count;
+                    current_src_frame = source_frame_for_time(frame_time);
                     if (params.sharpness.enabled && params.sharpness.window_mode) {
                         const int window_index = params.mode == ExtractionMode::FPS
                                                      ? static_cast<int>(std::floor(
--- a/src/io/video_frame_extractor.cpp
+++ b/src/io/video_frame_extractor.cpp
@@ -1419,8 +1419,7 @@ namespace lfs::io {
                         [](const CandidateFrame& a, const CandidateFrame& b) {
                             return a.score < b.score;
                         });
-                    std::filesystem::path fname = generate_filename(
-                        written_count + 1);
+                    std::filesystem::path fname = generate_filename(best->source_frame);
                     // Apply rotation to the best window frame before writing
                     int write_w = out_width;
                     int write_h = out_height;
@@ -1478,7 +1477,7 @@ namespace lfs::io {
 
                 auto process_frame_hw = [&](AVFrame* hw_frame) {
                     throw_if_cancelled();
-                    std::filesystem::path filename = generate_filename(saved_count + 1);
+                    std::filesystem::path filename = generate_filename(current_src_frame);
 
                     const AVPixelFormat hw_sw_format = hardwareFrameSoftwareFormat(hw_frame);
                     if (decoded_software_format == AV_PIX_FMT_NONE &&
@@ -1765,7 +1764,7 @@ namespace lfs::io {
                     }
                     // --- End rotation ---
 
-                    std::filesystem::path filename = generate_filename(saved_count + 1);
+                    std::filesystem::path filename = generate_filename(current_src_frame);
 
                     if (gpu_encoding_enabled) {
                         if (batch_encode_w == 0) {
--- a/src/io/video_frame_extractor.hpp
+++ b/src/io/video_frame_extractor.hpp
@@ -82,7 +82,7 @@ namespace lfs::io {
             int custom_height = 0;
 
             // Output naming
-            std::string filename_pattern = "frame_%d"; // %d = frame number
+            std::string filename_pattern = "frame_%d"; // %d = 1-based source video frame number
             bool generate_metadata = false;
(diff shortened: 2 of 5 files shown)


MrNeRF (commented):
Thx for picking this up! `frame_%d` meaning the real video frame is the right call, users read it that way and we did not deliver it.

The sharpness window hunk is a clean fix. The two other ones I do not think hold yet, and I would like you to look at them again before we take this.

Short version:

1. `current_src_frame` is not a source frame number in the sequential path, it is the decode counter, and that counter restarts after the trim seek. So the trimmed case from your description still writes `frame_1.png` on any normal video. Your test is green only because the 5 frame fixture is a single GOP.
2. Source frame numbers are not unique the way the old counters were. We overwrite files without noticing and still report them as written.
3. The cuda guard from the second commit does not skip, it turns into a failure.

Everything with file and line in the comments below. None of this is hard to fix, the first two are basically one helper.

Two things I noticed while reading, both **not** for this PR, only so they do not get lost:

- `video_extractor_dialog.cpp:1262` builds the pattern preview with a hardcoded `1`, so somebody who trims to 00:30 sees `frame_1.png` in the dialog and gets `frame_901.png` on disk.
- unpadded `%d` sorts badly and gets worse with real frame numbers (`frame_1000` before `frame_2`). `formatFrameFilenameStem` already handles `%04d`, so maybe the default is worth a thought some day.

On compatibility: I grepped, `filename_pattern` only goes from the dialog into the extractor, nothing in the repo parses the number back out or expects `1..N` without gaps. So we only break external scripts. I am fine with that, it needs a line in the release notes.

Happy to help if something is unclear.

src/io/video_frame_extractor.cpp
```diff
@@ -1478,7 +1477,7 @@ namespace lfs::io {
 
                 auto process_frame_hw = [&](AVFrame* hw_frame) {
                     throw_if_cancelled();
-                    std::filesystem::path filename = generate_filename(saved_count + 1);
+                    std::filesystem::path filename = generate_filename(current_src_frame);
```
MrNeRF, line 1508: This is the one I would really like you to look at again.

`current_src_frame` is not the source frame number here, at least not on the path that matters. Two lines write it:

- `video_frame_extractor.cpp:1872` (sequential): `current_src_frame = decoded_frame_count;` - that is just how many frames we decoded so far.
- `video_frame_extractor.cpp:1841` (sparse/retained): from the timestamp, `llround(frame_time * video_fps) + 1`.

Only the second one is a real frame number. `decoded_frame_count` starts at 0 (`:1294`) and nothing rebases it after the trim seek at `:1080`, and that seek goes **backwards to the previous keyframe**. With a normal GOP of 250 and a trim starting at 60s we land at ~50s, start counting at 1 from there, and the first file is `frame_1.png` again. That is exactly the case your description says is fixed.

Simplest fix is to take the timestamp everywhere:

```cpp
const auto source_frame_for_time = [&](const double frame_time) {
    return std::max(1, static_cast<int>(std::llround(frame_time * video_fps)) + 1);
};
```

and use it in `process_selected_frame` (`:1840`) and in `process_sequential_frame` (`:1872`). Then the trim case works and both paths finally say the same thing - today they disagree on VFR material, and with your change that disagreement lands in the filenames the user sees.

src/io/video_frame_extractor.cpp
```diff
@@ -1765,7 +1764,7 @@ namespace lfs::io {
                     }
                     // --- End rotation ---
 
-                    std::filesystem::path filename = generate_filename(saved_count + 1);
+                    std::filesystem::path filename = generate_filename(current_src_frame);
```
MrNeRF: Same value as in the hw path above, so the same thing applies here. One more that is specific to the new scheme:

`saved_count + 1` was unique by construction, a source frame number is not. Nothing in `generate_filename` or `write_image_file` checks whether we wrote that name already, so we just overwrite.

The path that hits it is the fps tail fill at `:2043`:

```cpp
while (shouldFillRetainedFpsTail(...) && ... frameCoversSampleTime(retained_frame_time, ...)) {
    process_selected_frame(sparse_previous_frame, retained_frame_time);
    next_capture_time += target_interval;
}
```

It calls `process_selected_frame` over and over with the **same** `retained_frame_time`, so it computes the same number and builds the same filename every round - and that is the frame the normal loop wrote a moment ago. `target_fps` is not clamped to `video_fps` (`:1050`), so 120fps out of a 30fps video runs this three times: three writes, no new file.

The bad part is that we do not tell anybody. `written_count` counts the overwrites, so `Extracted {} frames` (`:2253`) and `written_frames` in the metadata are too high, and `saved_frames` carries the same filename three times in `extraction_metadata.json`. The user asks for N frames, gets N-3, and everything looks green.

Can you decide what a repeated source frame should be called and make it explicit please? Skipping the duplicate write is fine by me, a suffix is fine too. Only the counters and the metadata have to match what is really on disk. An assert that we did not emit that path before would be nice on top.

src/io/video_frame_extractor.cpp
```diff
@@ -1419,8 +1419,7 @@ namespace lfs::io {
                         [](const CandidateFrame& a, const CandidateFrame& b) {
                             return a.score < b.score;
                         });
-                    std::filesystem::path fname = generate_filename(
-                        written_count + 1);
+                    std::filesystem::path fname = generate_filename(best->source_frame);
```
MrNeRF, line 1447: This one is a clear win. `written_count + 1` was simply wrong here and the winning candidate carried the right number all along, so filename and the `source_frame` we write into the metadata finally agree. Thx!

src/io/video_frame_extractor.hpp
```diff
@@ -82,7 +82,7 @@ namespace lfs::io {
             int custom_height = 0;
 
             // Output naming
-            std::string filename_pattern = "frame_%d"; // %d = frame number
+            std::string filename_pattern = "frame_%d"; // %d = 1-based source video frame number
```
MrNeRF, line 85: Comment is good and I want it. Only keep in mind it is a promise we do not keep yet for trimmed extractions, see my comment in the cpp. Once that is sorted this line is accurate.

tests/test_video_hdr_logic.cpp
```diff
@@ -3,16 +3,25 @@
 
 #include "io/hdr_tonemap.hpp"
 #include "io/video_frame_extractor.hpp"
+#include "io/video/video_encoder.hpp"
```
MrNeRF: Small one: this file was a pure logic TU so far, params validation and hdr helpers, cheap and it runs anywhere. With the encoder and `cuda_runtime.h` it now drags in a real encode, a real decode and a temp dir.

Can you move the two new tests into their own file please, something like `test_video_frame_extractor_naming.cpp`? Then the cheap ones stay cheap.

tests/test_video_hdr_logic.cpp
```diff
@@ -21,6 +30,111 @@ namespace {
...
+        ~CudaFloatBuffer() {
+            if (ptr)
+                cudaFree(ptr);
+        }
+
+        float* ptr = nullptr;
+        cudaError_t status = cudaSuccess;
+    };
+
+    void skipIfCudaUnavailable() {
```
MrNeRF: This does not skip anything unfortunately.

`GTEST_SKIP()` expands to a `return` (`gtest/internal/gtest-internal.h:1301`), so it returns out of **the helper**, not out of the test. The test body keeps going, `cudaMalloc` fails and the `ASSERT_TRUE` further down fires.

And the skip does not shield that failure:

```cpp
// gtest.cc:2465
bool TestResult::Skipped() const { return !Failed() && CountIf(..., TestPartSkipped) > 0; }
```

Failed wins. So a machine without a gpu reports FAILED, not SKIPPED - the opposite of what the second commit wanted.

Everywhere else in tests/ the macro sits directly in the test body or in `SetUp()`, see `test_fused_l1_ssim.cpp:33` or `test_colmap_image_layout.cpp:216`. Easiest here is to let the helper return `bool` and put `if (!cudaAvailable()) GTEST_SKIP() << "...";` at the top of both tests.

tests/test_video_hdr_logic.cpp
```diff
@@ -21,6 +30,111 @@ namespace {
...
+        int device_count = 0;
+        if (cudaGetDeviceCount(&device_count) != cudaSuccess || device_count <= 0)
+            GTEST_SKIP() << "CUDA device required for VideoEncoder-based fixture";
+    }
+
+    bool writeTinyEncodedVideo(const std::filesystem::path& video_path, std::string& error) {
+        constexpr int width = 16;
+        constexpr int height = 16;
+        constexpr int frame_count = 5;
```
MrNeRF: Five frames is one single GOP, and that is why the trim test below is green.

The backwards seek lands on frame 1, so the decode counter happens to be the real frame number and nobody notices that it restarts. On a video with a normal keyframe distance it does not, see my comment in the cpp.

To cover what the PR promises the fixture needs more than one GOP and a trim that starts inside a later one. Something like 40 frames with a forced keyframe distance of 10 and a trim starting around frame 25. As it stands this test cannot fail, and a regression test that cannot fail will not help us the next time somebody touches this.

tests/test_video_hdr_logic.cpp
```diff
@@ -183,3 +297,47 @@ TEST(VideoFrameExtractorParams, RejectsInvalidRangesAndDimensions) {
...
+    std::filesystem::create_directories(output_dir);
+
+    std::string error;
+    ASSERT_TRUE(writeTinyEncodedVideo(video_path, error)) << error;
+
+    auto params = extractionParams(video_path, output_dir);
+    params.frame_interval = 2;
+
+    VideoFrameExtractor extractor;
+    EXPECT_TRUE(extractor.extract(params, error)) << error;
```
MrNeRF: Two small ones, they apply to both new tests.

`EXPECT_TRUE` here and then we go on checking files. If extraction fails you get five confusing failures instead of one clear one - `ASSERT_TRUE`, like you already do for `writeTinyEncodedVideo`.

And nothing says how many files we expect. A count keeps an eye on duplicates and strays, which is the class of problem from my other comment:

```cpp
EXPECT_EQ(3, std::distance(std::filesystem::directory_iterator{output_dir},
                           std::filesystem::directory_iterator{}));
```

MrNeRF:
Review above is by claude but I think it has a point

kmeirlaen:
> Review above is by claude but I think it has a point

yeah, i focussed too much on the extraction with quality filtering - i'll do some more testing on normal extraction and trimmed videos

kmeirlaen:
comment is valid, took some tinkering with the time stamps but in a 30 FPS video trimmed at 20.3s can write frame_10.jpg instead of around frame_610.jpg.

What happened:
* GUI trim start: 20.3s
* FFmpeg seeked backward to a nearby keyframe, probably around 20.0s
* extractor decoded about 10 frames after that keyframe
* sequential path used decoded_frame_count
* filename became frame_10.jpg

I'm adding a fix for this and also tests

kmeirlaen:
Update:
* now computes source-frame filenames from timestamps in both selected/sparse and sequential paths.
* now tracks emitted filenames and skips duplicate source-frame outputs with a warning instead of overwriting.
* Moved the heavy encode/decode naming coverage into seperate test

MrNeRF:
Good catch, and thanks for digging into the write paths rather than just the formatter. One thing your patch fixes that isn't even in the description: because we seek to the start keyframe whenever `start_time > 0.1`, `decoded_frame_count` was a *post-seek* ordinal, so trimmed extractions were writing wrong `source_frame` values into `extraction_metadata.json` too, not only into filenames. Nice side effect.

A few suggestions before this goes in — all improvements, nothing I'd call a blocker.

**Prefer bumping the number over dropping the frame.** `source_frame_for_time()` assumes constant frame rate. That holds for most footage, but for VFR sources (screen recordings, some Android cameras) two selected frames can round to the same index, and then `reserve_output_filename()` skips the write completely. Right now no frame is ever lost, so that would be a small step back for those files, and the only trace is a log line. What do you think about keeping the number unique instead:

```cpp
current_src_frame = std::max(source_frame_for_time(frame_time), last_emitted_src_frame + 1);
```

Numbering stays monotonic and unique, nothing gets discarded, and the dedup set still earns its keep for the real duplicate sources (the FPS tail-fill loop and the sparse-seek path re-selecting the retained frame). Same reasoning for the `video_fps <= 0` case where we substitute `30.0` — there the timestamps have no relation to the actual frame spacing, so numbering from them is a guess.

**The duplicate warning gets loud.** One `LOG_WARN` per skipped duplicate means a 30 fps request from a 1 fps timelapse prints thousands of lines. Could you count them and emit a single summary next to the existing `"Extracted {} frames from video"`?

**Output count vs. what the dialog promises.** Asking for 30 fps from a 10 fps source now yields 5 files where it used to yield 15. I think that's the better behaviour — copies of the same picture are useless as training input — but `ESTIMATED_FRAMES` in the extractor dialog and the progress denominator still quote the old number, so the user is told 15 and gets 5. Worth making those agree, or at least mentioning it in the PR description as an intended change.

**Description tweak:** "Existing metadata `source_frame` values are unchanged" isn't quite right — they do change for every sequential extraction, and for trims that's precisely the fix. Would be good to spell that out, plus the duplicate-skipping, since neither is mentioned and both are user-visible.

On the tests:

- `start_time = 2.3` sits exactly on a frame boundary, and we compare `frame_time < start_time` with no epsilon while `frame_time` comes from `pts * time_base` with a non-binary-exact mp4 time base. Whether frame 24 lands a hair above or below 2.3 can depend on the muxer, so this could go flaky on someone else's box. Picking a value between two frames (2.25) makes it deterministic.
- If FFmpeg has neither NVENC nor a software H.264 encoder, `ASSERT_TRUE(writeEncodedVideo(...))` fails for a reason that has nothing to do with naming. `GTEST_SKIP` there would be friendlier.
- Sharpness window mode is one of the three paths in your list but has no test, and the pattern parser handles `frame_%04d` which also isn't covered. Not required for this PR, just flagging it while you're in the file.

Tiny one: the dedup key is fully determined by the frame number (`formatFrameFilenameStem` always emits a number, even for pattern-less input), so `std::unordered_set<int>` would do the same work with less memory and no path hashing per frame.

Direction is right and I'd like to merge this — mainly looking for the collision handling and the log summary.

PR #1464, opened 2026-07-27
Complete locale coverage and add validation

### LINKED ISSUE #1447: Translations fall behind because nothing checks for missing keys
`en.json` has 1,711 keys and we ship 10 languages, but the other files fall behind and nothing catches it.

**What is missing today**

- **15 keys are missing in 8 of the 10 languages.** All of them belong to newer training options:
  - `training.tooltip.use_normal_loss`
  - `training.tooltip.normal_loss_weight`, `normal_consistency_weight`, `normal_flatten_weight`
  - `training.options.depth_loss.ssi`, `ssi_depth`, `ssi_disparity`
  - `training.tooltip.opt.depth_loss_mode.ssi`, `ssi_depth`, `ssi_disparity`
  - `training_params.normal_consistency_weight`
- **`tooltip.gt_compare_mode` is missing in all 9 non-English files.**
- Many values are still byte-identical to English, so they were never actually translated:

| language | untranslated values |
|---|---|
| nl | 335 |
| fr | 287 |
| it | 283 |
| es | 279 |
| pl | 268 |
| de | 259 |
| ja | 238 |
| ko | 238 |
| zh | 147 |

The pattern is clear: when a new training parameter is added, the key goes into `en.json` and nowhere else.

**What to do**

1. Add a CI check that every locale file contains every key from `en.json`. Fail the build otherwise. This is the important part — it stops the problem from coming back.
2. Fill the 15 + 1 missing keys.
3. Optionally report the untranslated-identical counts so translators know where to start.

### LINKED ISSUE #1448: Some UI text is hardcoded English and cannot be translated
Some UI text never reaches the translation system at all, because it is written straight into the markup or into C++.

**83 English strings sit directly in `.rml` files.** Examples:

- `scene_tree.rml` — the `Scene`, `History` and `Logging` tabs, `Log Level`, and the level names `Trace`, `Debug`, `Performance`, `Error`, `Critical`
- `scene_tree.rml` — `Shared History`, `No history yet`, `Clear History`, `Undo Stack`, `Redo Stack`, `Nothing recorded yet`, `No logs captured yet`
- `python_console_panel.rml` — `Reload`, `Save As`, `Format`, `Reset`
- `sequencer_overlay.rml` — `Close`, `Edit Keyframe Time`, `Cancel`
- `watch_dirs_dialog.rml` — `Remove`, `Cancel`
- `startup.rml` — `Discord`, `Donate`

**Some are set from C++ too**, so they are not just placeholders:

```
src/visualizer/gui/scene_panel_native.cpp:833
    changed |= setCachedText(history_summary_label_el_, "Shared History");
```

No translation can ever reach these strings, so those parts of the window stay English in every language.

**What to do**

Move them into `en.json` (plus the other locale files) and look them up the same way the rest of the UI does.

Note: technical strings such as training parameter names stay in English on purpose — this issue is only about labels and buttons.

francescofugazzi:
## Summary

This PR completes the current localization pass for the shipped language catalogs, improves runtime language switching, and adds dedicated safeguards to prevent localization regressions from reaching the application.

## Localization updates

- Localized remaining interface, training, dataset, Asset Manager, Mesh-to-Splat, PPISP, histogram, and video-extraction strings across:
  - German
  - Spanish
  - French
  - Italian
  - Dutch
  - Polish
  - Japanese
  - Korean
  - Simplified Chinese
- Preserved established 3DGS terminology, acronyms, formats, units, and product names where they are conventionally kept in English.
- Removed Korean Asset Manager strings that had accidentally been present in non-Korean catalogs:
  - `fr.json`
  - `it.json`
  - `pl.json`
- Removed obsolete depth-loss entries from the affected locale catalogs.
- Corrected the Japanese PLY model count string to use the compatible `std::format` placeholder `{}` instead of `%zu`.
- Normalized locale JSON formatting where needed and checked the catalogs for literal Unicode escape sequences.

## Runtime localization fixes

- Fixed Scene → Logging localization refresh so Log Level options update through RmlUi’s select API and remain visible in both the closed control and dropdown after a language change.
- Reused the already parsed English catalog as the fallback map, avoiding duplicate parsing and storage when English is active.
- Reduced fallback diagnostics to one warning per missing key, preventing repeated warnings on frequently refreshed UI paths.
- Updated `lf.ui.set_language()` to load CJK fallback fonts for Japanese, Korean, and Simplified Chinese.
- Refresh RmlUi font resources after the first CJK font load so already-open tabs, buttons, and panels recalculate text metrics and layout correctly during a runtime language change.

## Localization safeguards

- Added `tools/check_locale_completeness.py`, which validates exact locale key parity with `en.json` and compatible `std::format` placeholders.
- The checker rejects:
  - missing keys;
  - unexpected or obsolete keys;
  - malformed or mismatched `std::format` placeholders.
- Named placeholders may be reordered by translations where required by grammar.
- Added a standalone GitHub Actions workflow that runs the default locale validation on pull requests, pushes to `master`, and manual dispatches.
- Added optional diagnostics for values identical to English. These remain non-blocking because several technical terms, formats, and acronyms are intentionally shared across locales.
- Documented the validation rules, audit modes, and command examples in `CONTRIBUTING.md` and the checker’s `--help` output.

## Validation

- Verified valid JSON for all updated locale files.
- Ran the locale-completeness check successfully against all 1,770 English keys.
- Verified the optional identical-value report, detailed listing, and strict audit mode.
- Tested runtime language switching from the Python Console:

  ```python
  import lichtfeld as lf

  lf.ui.set_language("ja")
  lf.ui.set_language("en")
  lf.ui.set_language("ko")
  lf.ui.set_language("zh")
  ```

## Translation review

The translations were reviewed for consistency and terminology, but native-speaker review is still needed before treating every localized phrase as final.

Fixes #1447  
Fixes #1448


src/visualizer/gui/scene_panel_native.cpp
```diff
@@ -825,25 +827,37 @@ namespace lfs::vis::gui {
...
+            lichtfeld::Strings::Scene::LOG_TRACE, lichtfeld::Strings::Scene::LOG_DEBUG,
+            lichtfeld::Strings::Scene::LOG_INFO, lichtfeld::Strings::Scene::LOG_PERFORMANCE,
+            lichtfeld::Strings::Scene::LOG_WARN, lichtfeld::Strings::Scene::LOG_ERROR,
+            lichtfeld::Strings::Scene::LOG_CRITICAL, lichtfeld::Strings::Scene::LOG_OFF};
+        const std::array log_option_ids{"logging-level-trace", "logging-level-debug",
+                                        "logging-level-info", "logging-level-performance",
+                                        "logging-level-warn", "logging-level-error",
+                                        "logging-level-critical", "logging-level-off"};
+        for (size_t i = 0; i < log_option_keys.size(); ++i)
+            changed |= setCachedText(document_->GetElementById(log_option_ids[i]), LOC(log_option_keys[i]));
```
MrNeRF: This doesn't work, GetElementById never finds the options.

RmlUi moves all `<option>` children out of the DOM into the select's internal selectbox (`ElementFormControlSelect::MoveChildren`), and that selectbox is attached as a non-DOM child. The BFS in `ElementUtilities::GetElementById` only walks DOM children, so it goes right past them. You get nullptr and setCachedText silently does nothing.

I built it and ran it, the Log Level dropdown is empty now, english too. Before it showed Info because the text sat in the rml.

`logging_level_select_el_` is already an ElementFormControlSelect, so this should do it:

```cpp
for (size_t i = 0; i < log_option_keys.size(); ++i)
    if (auto* const opt = logging_level_select_el_->GetOption(static_cast<int>(i)))
        changed |= setCachedText(opt, LOC(log_option_keys[i]));
logging_level_select_el_->SetSelection(logging_level_select_el_->GetSelection());
```

The SetSelection at the end is needed too. Setting the option text alone does not refresh the closed box, RmlUi only redraws that when `value_rml_dirty` is set and that happens inside SetSelection. Without it the closed dropdown keeps the old language after a switch.

src/visualizer/gui/scene_panel_native.cpp
```diff
@@ -825,25 +827,37 @@ namespace lfs::vis::gui {
...
-        changed |= setCachedText(logging_copy_btn_el_, "Copy");
+        changed |= setCachedText(logging_summary_label_el_, LOC(lichtfeld::Strings::Scene::LOG_MIRROR));
+        changed |= setCachedText(logging_level_label_el_, LOC(lichtfeld::Strings::Scene::LOG_LEVEL));
+        changed |= setCachedText(logging_export_btn_el_, LOC(lichtfeld::Strings::Scene::LOG_EXPORT));
+        changed |= setCachedText(logging_copy_btn_el_, LOC(lichtfeld::Strings::Scene::LOG_COPY));
         changed |= setCachedText(logging_note_el_,
-                                 std::format("Mirrors console output. The panel shows the latest {} entries to stay responsive; copy and export use the full buffered log history.",
-                                             MAX_RENDERED_LOG_ENTRIES));
-        changed |= setCachedText(logging_empty_el_, "No logs captured yet");
+                                 std::vformat(LOC(lichtfeld::Strings::Scene::LOG_NOTE),
```
MrNeRF, line 847: Careful, this makes a translated string a format string. If a translator drops the {} or leaves a stray { in there, vformat throws and we go down in the per frame sync.

Right now all 10 locales have the {} so nothing is broken, but nothing keeps it that way either. Same pattern is already in sequencer_ui_manager.cpp:2375 and 2379. Would be nice if your new script checks this, see my other comment.

MrNeRF (commented):
Thanks a lot for this! Translations look good and the korean strings are really gone from fr/it/pl now, 56 lines each before, 0 now. Build passes, your script passes, ran the gui and there is no error noise.

One thing blocks it though: the Log Level dropdown is empty now in every language, see my comment in scene_panel_native.cpp. Rest is small stuff.

One more that is not in your diff: japanese ply_models_count in locales/ja.json is "PLYモデル: %zu" instead of {}, so japanese users see a literal %zu in the status bar. Old bug, but you are touching ja.json anyway.

tools/check_locale_completeness.py
```diff
@@ -0,0 +1,105 @@
...
+    english = load_locale(english_path)
+    english_keys = set(english)
+    failures: list[str] = []
+    identical_reports: list[tuple[str, list[str]]] = []
+
+    for locale_path in sorted(LOCALES_DIR.glob("*.json")):
+        if locale_path == english_path:
+            continue
+        locale = load_locale(locale_path)
+        missing = sorted(english_keys - set(locale))
```
MrNeRF, line 107: Two things I would add here while you are in it.

First, also report keys that a locale has but en.json doesn't. That is exactly the korean strings you cleaned out of fr/it/pl by hand. All 9 locales still carry 4 keys en.json doesn't have (training.options.depth_loss.adaptive_warped_l1, .pearson and the two matching training.tooltip.opt.depth_loss_mode.*). Old ones, not from you, but the script would stop them coming back.

Second, check that the {} placeholders match en.json. That is the only localization bug that can really crash us, see my comment in scene_panel_native.cpp. It would also have caught the %zu in ja.json.

src/core/event_bridge/localization_manager.cpp
```diff
@@ -84,6 +89,14 @@ namespace lfs::event {
         if (it != current_strings_.end()) {
             return it->second.c_str();
         }
+
+        const auto fallback_it = fallback_strings_.find(key_str);
+        if (fallback_it != fallback_strings_.end()) {
+            LOG_WARN("Missing localization key '{}' in '{}'; using English fallback",
```
MrNeRF: This warns on every single call. get() runs from the per frame sync paths, so one missing key spams the log forever, and with the fallback the ui looks fine so nobody notices it.

Can you make it warn once please? A small mutable unordered_set<std::string> next to the map is enough.

src/core/event_bridge/localization_manager.cpp
```diff
@@ -69,6 +70,10 @@ namespace lfs::event {
         const bool has_default = std::find(available_languages_.begin(),
                                            available_languages_.end(),
                                            DEFAULT_LANGUAGE) != available_languages_.end();
+        if (has_default &&
+            !parseLocaleFile(locales_dir_ + "/" + DEFAULT_LANGUAGE + ".json", fallback_strings_)) {
```
MrNeRF: nitpick, you decide. en.json gets parsed 3 times at startup now: once in the loop above into test_strings which we throw away, once here, and once more in setLanguage. And when the language is english we keep those 1770 keys twice in memory (fallback_strings_ plus current_strings_). You could reuse the parse from the loop and skip the copy when the language is en.

francescofugazzi:
Implemented the requested localization follow-up changes:

- Updated the Scene → Logging level refresh to retrieve `<select>` options through `GetOption(i)` and reapply the current selection, so translated option labels are visible both in the closed field and in the dropdown.
- Reworked localization fallback handling:
  - `en.json` is parsed once and reused as the fallback catalog;
  - selecting English no longer duplicates the catalog in the current-language map;
  - missing-key/fallback warnings are emitted once per key instead of repeatedly on per-frame paths.
- Extended `tools/check_locale_completeness.py` to fail on:
  - missing keys;
  - unexpected/obsolete keys;
  - malformed or mismatched `std::format` placeholders.
  Named placeholders may still be reordered by translations.
- Removed the obsolete depth-loss keys from the affected locale files and corrected the Japanese `status.ply_models_count` placeholder from `%zu` to `{}`.
- Updated `CONTRIBUTING.md` and the checker’s `--help` output with the current validation rules, audit modes, and runnable examples.

I also fixed an additional runtime-language-switch issue found while testing from the Python Console. Switching directly to Japanese, Korean, or Chinese now loads the CJK fallback fonts and invalidates RmlUi font resources once, so already-open tabs, buttons, and panels recalculate their text metrics and spacing correctly.

Example commands used for the runtime check:

```python
import lichtfeld as lf

lf.ui.set_language("ja")
lf.ui.set_language("en")
lf.ui.set_language("ko")
lf.ui.set_language("zh")
```

The locale checker passes in its CI-equivalent default mode, while --report-identical, --list-identical, and --fail-on-identical behave as documented.

MrNeRF:
Thank you

PR #1468, opened 2026-07-28
Fix Empty-Name Group Merge When Merging To Single PLY

### LINKED ISSUE #1458: merge to single .ply causes splats to be deleted

kmeirlaen:
Fixes #1458  

## Summary

This patch fixes the Scene Navigation `Merge to Single PLY` path for groups. The merge code now keeps a stable copy of the group name before removing the original group node, so the replacement splat is created with the intended name instead of sometimes using an invalid string.

It also adds a focused regression test that reproduces the original failure mode by passing the group node's own `name` reference into `SceneManager::mergeGroupNode()`.

Original issue: https://github.com/MrNeRF/LichtFeld-Studio/issues/1458

## Observed Symptoms

When merging a group into a single PLY, the operation could appear to delete the group and its splat children instead of replacing them with one merged splat.

The repro log showed the merge attempting to create a node with an empty name:

```text
[01:27:29.869] [warn] scene.cpp:171  Cannot add node with empty name
[01:27:29.871] [info] scene_manager.cpp:4246  Merged group '' -> ''
```

That log is the important clue: the merge was not failing because of a duplicate name. It was losing the group name during the merge.

## Replication Steps For The Original Issue

1. Load a `.ply` model.
2. Duplicate the `.ply`.
3. Move the duplicate so both models are visible separately.
4. In Scene Navigation, create a new group.
5. Move both splat nodes into that group.
6. Right-click the group and choose `Merge to Single PLY`.
7. Before this patch, the group and children could disappear, with the log showing `Cannot add node with empty name` and `Merged group '' -> ''`.

## What This Patch Fixes

- Copies `node->name` before dispatching `MergeGroupById` to avoid passing a scene-node-owned string reference into merge logic.
- Copies the `mergeGroupNode()` input into a local `group_name` immediately and uses that stable copy for lookup, history capture, removal, replacement insertion, events, logging, and return value.
- Adds `UndoHistoryTest.MergeGroupNodeHandlesNameReferenceFromGroupNode`, which intentionally passes `group->name` directly to reproduce the dangerous path.
- Keeps the small MSVC test-build fix for `tests/test_error_envelope.cpp`, splitting a hex escape so `\x80` is not parsed together with following hex-looking characters.

## Root Cause

`MergeGroupById` looked up the group node by id and called:

```cpp
mergeGroupNode(node->name);
```

`mergeGroupNode()` accepted that as `const std::string& name`. If the reference pointed to the group node's own `name` member, the function later removed that same group node:

```cpp
scene_.removeNode(name, false);
scene_.addSplat(name, std::move(merged_model), parent_id);
```

Removing the group destroys the string object backing `name`. Every later use of `name` is then a dangling reference. In the observed run, that dangling reference read back as an empty string, so `addSplat()` tried to add a node named `""` and failed with `Cannot add node with empty name`.

## Why The Issue Showed Up

The UI context-menu flow uses `MergeGroupById`, so it naturally passes through the id-to-node lookup path. That made it possible for the merge function's `name` parameter to alias the very node being removed.

Tests or callers that passed an independent `std::string` copy did not expose the bug, which is why the earlier happy-path merge tests could pass while the UI action still failed.

## Verification

The new regression test failed before the fix with the same symptom as the user log:

```text
Cannot add node with empty name
Merged group '' -> ''
```

After the fix:

```text
UndoHistoryTest.MergeGroupNodeHandlesNameReferenceFromGroupNode passed
Merged group 'group' -> 'group'
```

### DIFF
--- a/src/visualizer/scene/scene_manager.cpp
+++ b/src/visualizer/scene/scene_manager.cpp
@@ -596,8 +596,10 @@ namespace lfs::vis {
 
         cmd::MergeGroupById::when([this](const auto& cmd) {
             const auto* node = scene_.getNodeById(static_cast<core::NodeId>(cmd.node_id));
-            if (node)
-                mergeGroupNode(node->name);
+            if (node) {
+                const std::string node_name = node->name;
+                mergeGroupNode(node_name);
+            }
         });
 
         // Handle node selection from scene panel (both PLYs and Groups)
@@ -4120,13 +4122,14 @@ namespace lfs::vis {
     }
 
     std::string SceneManager::mergeGroupNode(const std::string& name) {
-        const auto* group = scene_.getNode(name);
+        const std::string group_name = name;
+        const auto* group = scene_.getNode(group_name);
         if (!group || group->type != core::NodeType::GROUP) {
             return {};
         }
 
         const auto history_options = sceneGraphCaptureOptions(true, false);
-        auto history_before = op::SceneGraphPatchEntry::captureState(*this, {name}, history_options);
+        auto history_before = op::SceneGraphPatchEntry::captureState(*this, {group_name}, history_options);
         const core::NodeId group_id = group->id;
         const core::NodeId parent_id = group->parent_id;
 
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thx, looks much better. I will add a follow up refactor that will move everything to nodeids instead of strings.

Issue #1454, opened 2026-07-27
Write down (or revisit) the one-RmlUi-context-per-panel design


MrNeRF:
Right now every panel creates its own RmlUi context, renders into its own texture, and gets composited onto the screen (`RmlPanelHost` in `src/visualizer/gui/rmlui/rml_panel_host.cpp`). On top of that there are 11 fixed contexts (shell, menu bar, status bar, right panel, toasts, modals, overlays, ...), so a normal session runs 20 or more contexts at once.

**The good part**

It caches really well. From `performance.log`:

```
gui_render.panel_layout.renderRightPanel.cached   0.08 - 0.11 ms
gui_render.rmlui_record                           0.08 - 0.40 ms
gui_render.cpu_ui_before_vulkan_begin             0.6  - 0.9  ms
```

A panel that did not change costs almost nothing. For a GPU-first app that is the right priority.

**The cost**

Because contexts do not know about each other, several things RmlUi normally does for free are written by hand:

- **Hit testing.** `GuiManager::hitTestPointer` (gui_manager.cpp:7017) is a fixed chain of 7 checks — modal, menu bar, export lock, overlay, floating panel, sequencer, viewport overlay, resize edge. Every new UI surface means editing this chain.
- **Window stacking.** `float_stack_order`, `alloc_float_stack_order_locked`, `bring_floating_panel_to_front_locked`.
- **Dropdowns.** `openDropdownBounds`, `openDropdownContainsPoint`, `openDropdownOptionAtPoint`, `setManualDropdownHover` — a popup cannot leave its panel's texture, so the hover behaviour is rebuilt by hand.
- **Tooltips, focus tracking and input forwarding**, once per host.

This is where most of our recent input bugs came from.

**What this issue is for**

Not to undo the design — the caching win is real and measured. But the trade-off should be written down in `docs/` so the next person knows *why* hit testing is a manual chain, and we should decide whether popups (dropdowns, context menus, tooltips) can move to a single shared overlay context. That alone would remove a large part of the hand-written code without giving up the per-panel caching.

PR #1474, opened 2026-07-29
Draft: RmlUI context investigation and native input cleanup

francescofugazzi:
Related to #1454.

## Summary

This draft documents and validates the current RmlUI context model, adds a lightweight GUI-only test path, fixes one render-on-demand feedback loop, and starts reducing custom UI mechanics where RmlUI can safely own them.

It is intentionally a draft. The shared-context work is an investigation and opt-in prototype, not a proposal to replace all panel-local contexts yet.

## Why

LichtFeld Studio currently uses one RmlUI context per `RmlPanelHost`, each rendered into its own texture and composited later. This is effective for unchanged panels, but contexts cannot coordinate with each other. The application has consequently accumulated custom routing for input, stacking, popups, tooltips, focus, and caching.

The goal of this work is to understand which custom paths are required editor policy, which are required by the panel-local cache boundary, and which can be handed back to RmlUI without regressing interaction or performance.

### Context-model documentation and investigation record

Added and expanded `docs/docs/development/rmlui-context-model.md`.

It now documents:

- the fixed and dynamic RmlUI context inventory;
- why a normal session reaches 20+ contexts;
- the per-panel texture/cache architecture and its measured motivation;
- manual input, tooltip, focus, popup, and stacking boundaries;
- the existing timing and demand-trace instrumentation;
- benchmark requirements and decision criteria for any shared-context migration;
- the distinction between application policy and duplicated UI mechanics;
- the RmlUI 6.2 limitation that prevents moving a live native `<select>` popup between contexts.

The documentation also records the current conclusion for panel-local selects: a separate overlay context would have to mirror options and manually route selection back to the source document. That would relocate the bridge instead of removing it. A genuinely native popup path requires the source document itself to participate in a full-window shared context.

### Lightweight GUI test configuration without LibTorch

Added `BUILD_GUI_TESTS`, independent from the existing full `BUILD_TESTS` suite.

- `BUILD_TESTS=ON` remains the full tensor/oracle suite and still requires LibTorch.
- `BUILD_GUI_TESTS=ON` builds only `lichtfeld_gui_tests`.
- The lightweight target links the GUI/runtime dependencies needed for focused RmlUI and panel-orchestration tests, but does not discover or require Torch, Python development components, or nanobind.
- The normal application build remains unchanged when both options are `OFF`.

This makes it possible to compile and run fast GUI regressions without downloading/configuring LibTorch solely for those tests.

### Panel rendering and context-demand coverage

Added lightweight coverage for:

- panel animation demand propagation;
- cached versus live direct render paths;
- panel preload/layout behavior;
- split caching of the right-panel scene header and active tab.

Added optional RmlUI demand tracing:

```text
LFS_RML_CONTEXT_DEMAND_TRACE=1
```

This logs only state transitions between immediate, delayed, and idle RmlUI demand. It is disabled by default and avoids per-frame logging.

### Python draw-hook feedback-loop fix

Removed the unconditional `lf.ui.request_redraw()` call from the Python empty-state viewport overlay draw hook.

The overlay is static while visible. Requesting another redraw from every draw created a render-on-demand feedback loop after startup dismissal: each draw scheduled the next frame indefinitely. Visibility and content changes already reach the renderer through ordinary input/state invalidation.

The demand trace now returns to idle after the relevant transition; dense later spans are attributable to user mouse input rather than queued RmlUI work.

### Shared-context ownership contract and opt-in spike

Extracted `RmlContextOwner` from `RmlPanelHost` so context lifetime is explicit and a host can safely borrow a context without destroying it.

Added a lightweight RmlUI contract covering:

- multiple documents in one context;
- native document stacking and hit testing;
- click delivery for overlapping documents;
- independent document unload;
- borrowed context lifetime;
- window-space document bounds;
- native mouseover bubbling, submenu popup traversal, and click delivery.

Added an opt-in shell/right-panel prototype:

```text
LFS_RML_SHARED_RIGHT_PANEL_SPIKE=1
```

When enabled, `RmlRightPanel` borrows the full-window `shell_frame` context instead of creating its own `right_panel` context. The right-panel document receives window-space bounds, keeps explicit document-ownership filtering for input, and the shell records the shared context after the right panel updates.

The default path is unchanged when the environment variable is not set.

This prototype confirms the topology change and provides a reversible measurement point. It is not yet a performance conclusion: the available baseline/shared captures used different interaction lengths and cannot be compared as a benchmark.

### Native RmlUI submenu hit testing

Removed `RmlMenuBar::dropdownElementAtPoint()` and its recursive C++ rectangle traversal.

The menu now uses the target resolved by RmlUI through `Context::ProcessMouseMove()` / `GetHoverElement()`.

The explicit submenu-open state remains intentional: it keeps the popup visible while the cursor crosses from the parent item into a previously hidden overflow popup. Once visible, RmlUI owns stacking-aware hover and click target resolution.

The existing Python resource regression was updated to require the native hover path and reject the removed geometry helper.

### Lightweight GUI configuration

Configured successfully with:

```text
BUILD_TESTS=OFF
BUILD_GUI_TESTS=ON
```

This configuration completed without LibTorch and produced `lichtfeld_gui_tests`.

### Focused RmlUI contract

```sh
cd build/gui-tests
tests/lichtfeld_gui_tests --gtest_filter=RmlSharedContextContractTest.* --gtest_color=yes --gtest_print_time=1
```

Result: 5/5 passed.

This includes the new:

```text
RmlSharedContextContractTest.NativeMouseEventsKeepAnExplicitSubmenuOpenAcrossItsPopupBoundary
```

### Full lightweight GUI label

```sh
ctest --test-dir build/gui-tests -C Release -L gui --output-on-failure
```

Result: 1/1 CTest target passed.

### Manual smoke test

Built and exercised the application manually:

- opened menus with submenus;
- moved the pointer from parent items into side popups;
- clicked submenu actions;
- closed menus by leaving/clicking outside;
- switched top-level menus while a dropdown was open.

No regression was observed.

## How to configure and run the lightweight GUI tests

Use a separate build directory if the existing one is configured for the normal application or full test suite.

```sh
cmake -S . -B build/gui-tests -G Ninja \
  -DCMAKE_BUILD_TYPE=Release \
  -DBUILD_TESTS=OFF \
  -DBUILD_GUI_TESTS=ON
```

Build the application and lightweight GUI tests:

```sh
cmake --build build/gui-tests --target LichtFeld-Studio --parallel
cmake --build build/gui-tests --target lichtfeld_gui_tests --parallel
```

Run the full lightweight GUI label:

```sh
ctest --test-dir build/gui-tests -C Release -L gui --output-on-failure
```

Run only the RmlUI shared-context contract:

```sh
cd build/gui-tests
tests/lichtfeld_gui_tests --gtest_filter=RmlSharedContextContractTest.* --gtest_color=yes --gtest_print_time=1
```

Run only the new submenu test:

```sh
cd build/gui-tests
tests/lichtfeld_gui_tests --gtest_filter=RmlSharedContextContractTest.NativeMouseEventsKeepAnExplicitSubmenuOpenAcrossItsPopupBoundary --gtest_color=yes --gtest_print_time=1
```

On Windows, use the executable with its `.exe` extension:

```bat
cd build\gui-tests
tests\lichtfeld_gui_tests.exe --gtest_filter=RmlSharedContextContractTest.* --gtest_color=yes --gtest_print_time=1
```

The existing full test configuration remains separate:

```text
BUILD_TESTS=ON
BUILD_GUI_TESTS=OFF
```

It still requires LibTorch and retains its existing tensor/oracle and Python test coverage.

## Reviewer feedback requested

1. Is the documented cache-versus-shared-context framing correct?
2. Is the opt-in shell/right-panel spike an acceptable basis for the next measurement experiment?
3. Should the next step continue with isolated same-document native cleanup, or should it move directly to one `RmlPanelHost` in the shared full-window context?
4. Are the current input-ownership constraints sufficient, especially the observed limitation for `Context::ProcessMouse...()` with separately positioned documents in one context?
5. Is the lightweight `BUILD_GUI_TESTS` split an acceptable long-term developer workflow?

## Non-goals

- This does not replace all panel-local contexts.
- This does not claim that a global/shared context is already performance-safe.
- This does not remove application-level editor policy such as modal/viewport precedence.
- This does not migrate native select popups into a separate overlay context.
- The shared right-panel path remains opt-in and disabled by default.


src/python/lfs_plugins/overlays/__init__.py
```diff
@@ -426,7 +426,6 @@ def draw_dashed_line(start_x, start_y, end_x, end_y):
     layout.draw_window_text(center_x - hint_w * 0.5, center_y + 70.0, hint, hint_color)
 
     layout.end_window()
```
MrNeRF, line 428: This is the one I'd really like back, but not as it was.

The overlay is animated. `:353` computes `dash_offset` from `lf.ui.get_time()` with `_ANIM_SPEED = 30.0`, that's the marching dashed border running at about 1.5 cycles per second. With render on demand, a draw hook asking for the next frame is simply how we express a continuous animation. So this call wasn't a bug, it was the animation.

I looked for anything else that could keep it alive and there is nothing. This hook is registered through `lf.ui.add_hook("viewport_overlay", "draw", ...)`, which lands in `PyUIHookRegistry`. But `FrameDemand::python_overlay` is fed by `PyViewportDrawRegistry` and `PyTransformGizmoRegistry` (`viewport_overlay.cpp:26-27`), a completely different registry, and both are false on an empty scene. `needsAnimationFrame()` only knows about `rml_viewport_overlay_`, which is the RmlUi document, not this hook.

You were right that an unconditional call pins the loop though, so please don't just revert it. Gate it on the overlay actually being visible, `:321-330` already computes that condition. And `_draw_drag_drop_overlay` needs the same treatment: it animates too (pulse and bounce), it never had a `request_redraw()`, and the empty state function returns early while dragging. So if you only fix this call site the drop overlay stays frozen.

src/visualizer/gui/rml_menu_bar.cpp
```diff
@@ -528,7 +528,7 @@ namespace lfs::vis::gui {
 
             Rml::Element* hit_element = nullptr;
             if (hovered_label < 0 && dropdown_container_)
```
MrNeRF, line 530: I like this change, it's clearly better than the old walk. The old helper ignored clipping, `overflow` and z-order, `ProcessMouseMove` already ran at `:485` before the hit test, `rml_mx`/`rml_my` are the same coordinates the helper used, and the old `hit_element ? hit_element : dropdownElementAtPoint(...)` fallback was already dead.

One thing to tighten though. `dropdownElementAtPoint()` only searched inside `dropdown_container_`. `GetHoverElement()` returns the hover element for the whole context, so `hit_element` can now be a toolbar button or a window control. That matters at `:546`, where `while (clicked && clicked != hit)` walks up from `hit_element` looking at `data-action` on the way. That walk now starts outside the dropdown.

It is harmless right now, I went through every `data-action` in `menubar.rml` (`window_close`, `window_minimize`, `window_toggle_maximize`, `window_toggle_ui`, `toggle_independent_split_view`, plus the data bound toolbar actions) and none of them are `"operator"` or `"callback"`, so we fall through to `closeDropdown()` like before. But nothing enforces that. One line brings back the old guarantee:

```cpp
hit_element = rml_context_->GetHoverElement();
if (!isDescendantOf(hit_element, dropdown_container_))
    hit_element = nullptr;
```

You already wrote that helper twice in this PR as `belongsToDocument`. Would be nice to have it once in a shared place instead of a third copy.

tests/test_rml_shared_context_contract.cpp
```diff
@@ -0,0 +1,319 @@
...
+            ASSERT_NE(context_, nullptr);
+        }
+
+        void TearDown() override {
+            if (context_)
+                Rml::RemoveContext(context_->GetName());
+        }
+
+        Rml::Context* context_ = nullptr;
+        NullRenderInterface render_interface_;
```
MrNeRF, line 110: This one is worth fixing before anything else, it's the only real crash risk I found.

`render_interface_` is a per test fixture member, but RmlUi documents the opposite lifetime right on `CreateContext` (`RmlUi/Core/Core.h:78-80`):

> the render interface and the text input handler must be kept alive until after the call to `Rml::Shutdown`. Alternatively, the render interface can be destroyed after all contexts it belongs to have been destroyed, and a subsequent call has been made to `Rml::ReleaseRenderManagers`.

`TearDown()` calls `RemoveContext` but never `ReleaseRenderManagers`, and `Rml::Shutdown()` only runs in `TearDownTestSuite`, long after each fixture and its interface are gone. So the `RenderManager` that RmlUi keeps for `&render_interface_` outlives the interface on every single test.

It only survives because the tests call `Update()` and never `Render()`, so nothing ever gets compiled through the interface. Note it is not because the handles are zero, `CompileGeometry` returns 1. Add one `Render()` call or a texture to any of the test documents and this becomes a real use after free at shutdown.

Easiest fix: make it `static` and own it in `SetUpTestSuite` so it outlives `Rml::Shutdown()`. Or call `Rml::ReleaseRenderManagers()` in `TearDown` right after `RemoveContext`.

tests/CMakeLists.txt
```diff
@@ -61,6 +63,58 @@ if(BUILD_UNICODE_TEST_ONLY)
...
+        GTest::gtest_main
+        glm::glm
+    )
+
+    add_test(NAME lichtfeld.gui
+        COMMAND $<TARGET_FILE:lichtfeld_gui_tests> --gtest_color=no
+    )
+    set_tests_properties(lichtfeld.gui PROPERTIES
+        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+        LABELS "gui;fast"
```
MrNeRF, line 107: Two small things here.

`LABELS "gui;fast"`, but the same four sources also go into `lichtfeld_tests`. So with `BUILD_TESTS=ON` and `BUILD_GUI_TESTS=ON`, which is a normal dev setup, `ctest -L fast` picks up both `lichtfeld.gui` and `lichtfeld.fast`, and `LFS_NON_FAST_TEST_FILTER` (`:532`) filters out neither `PanelRegistryRenderPathsTest.*` nor `RmlSharedContextContractTest.*`. The four files then run twice in two processes. Not harmful, just noisy and it doubles the tier. I'd use `LABELS "gui"` only.

And the bigger one: none of this runs in CI. `ubuntu.yml`, `windows.yml` and `nightly.yml` never set `BUILD_TESTS` or `BUILD_GUI_TESTS` and never call `ctest`. So `lichtfeld_gui_tests`, `test_panel_registry_render_paths.cpp` and `test_rml_shared_context_contract.cpp` have zero coverage on every platform, and the "5/5 passed" is one machine with one toolchain. Since this target needs neither Torch nor a GPU it's cheap to add, and honestly wiring it into CI is the whole point of having the switch.

src/visualizer/gui/rmlui/rml_context_owner.cpp
```diff
@@ -0,0 +1,39 @@
...
+        if (owns_context_ && manager_ && manager_->isInitialized())
+            manager_->destroyContext(context_name_);
+        context_ = nullptr;
+    }
+
+    bool RmlContextOwner::ensureContext(const int width, const int height) {
+        if (context_)
+            return true;
+
+        context_ = manager_->createContext(context_name_, width, height);
```
MrNeRF, line 35: `manager_` is `nullptr` on the borrowed path, and the only thing protecting us here is `if (context_) return true;` plus an `assert` that disappears in Release. So `RmlContextOwner(nullptr)` followed by `ensureContext()` is a null deref in a release build. Not reachable today, but one line closes it for good:

```cpp
if (!manager_)
    return false;
```

Also nice side effect: on the borrowed path `owns_context_` is already `false`, so if it ever did create a context here we would leak it. The early return covers both cases.

MrNeRF (commented):
Hi Francesco, thanks for picking this up. The context model is a real problem and I like that you started by measuring instead of rewriting. Let me be clear about what I want out of this work, because I think it points the effort in a slightly different direction.

## What I want to see proven

Two things, and only these two:

1. **One shared context is at least as fast as what we have now.**
2. **After that, we can delete the handwritten routing code completely.**

I don't want a second render path living next to the old one behind a flag. If sharing works it becomes the only way and there is nothing left to switch. If it doesn't work we drop it. A runtime flag just means we keep both paths alive forever, and that is the opposite of the goal.

So please, no new runtime flags. `LFS_RML_SHARED_RIGHT_PANEL_SPIKE` and `LFS_RML_CONTEXT_DEMAND_TRACE` should stay on your local branch. For an A/B test just build two binaries from two branches. The trace itself is nicely done by the way, only logging on state change and no allocation when off, exactly right for us. It's just measuring gear, not something we ship.

`BUILD_GUI_TESTS` is different. Build time only, never reaches users, that one is welcome. But then it also has to go into CI, otherwise it rots. Right now `ubuntu.yml`, `windows.yml` and `nightly.yml` never set `BUILD_TESTS` or `BUILD_GUI_TESTS` and never call `ctest`, so all the new tests run on your machine only.

## Good news: you can answer the speed question without building a shared context at all

I went through the RmlUi 6.2 headers we have installed. There is no partial render. `Context::Render()` always draws every visible document (`Core/Context.h:59-60`), `Element::Render()` is protected (`:592`), and there is no dirty region or damage API anywhere. `SetContextClipRect` saves GPU fill but not the CPU tree walk, and the CPU walk is what we actually measure.

That one fact decides everything. With one context, the cache is all or nothing for the whole window. If any single element changes, we re-record the whole UI.

So the question is simply:

- How often does *anything* in the UI change?
- What does one full-window record cost?
- Compare that against what we pay today: only the dirty panels re-record, everything else is a cheap blit.

All of this is measurable on the current build. We already time every context. A full-window record you can get by forcing everything dirty once (theme switch or resize) and adding up the per-context record times. "How often is something dirty" is just a counter.

If the numbers say the shared version loses, we know it before anyone writes the shared code path. That is a much stronger result than the spike can give us, and it costs an instrumentation patch instead of a parallel implementation.

## Why the current numbers can't decide it yet

**1. Our timers can't see the thing we need.** All four RmlUi cache timers use `LOG_TIMER_THRESHOLD(..., 0.25)` (`rmlui_manager.cpp:759, 792, 824, 854`), and anything under the threshold is dropped.

A cache hit never costs 0.25 ms. You can see it from the code: the outer scope `gui_render.panel_layout.renderRightPanel.cached` (`panel_layout.cpp:283`) is a normal `LOG_TIMER` and its median is 0.05 ms. Something inside a 0.05 ms scope can never trip a 0.25 ms floor. And in a 30k line log, `cached_context` shows up **zero** times.

The worse part: `cache_refresh` has the same threshold. So the refresh *count* is also too low. That means both numbers we need for the comparison are broken, not just one. Please drop the thresholds and use counters plus a histogram. We want the distribution, not a filtered average.

**2. The spike shares our two cheapest panels.** `shell.rml` is three empty divs with a background colour. The expensive one is the scene panel, your own doc says about 6.6 ms per refresh over 386 refreshes, and it stays separate. Sharing the cheap panels tells us nothing about the case that decides the architecture, because the whole risk of one context is that every extra document adds to the walk.

**3. The spike numbers aren't comparable.** `rml_right_panel.cpp:678` still runs its three pass height loop. In shared mode those become three *full-window* updates on the shell context, and `rml_shell_frame.cpp:173` adds a fourth. So it's four whole-window updates against one. The 0.46 ms median measures that, not the cache scope.

**4. We can't replay the same input.** The two captures are 44.7 s and 39.5 s with different clicking, and we have nothing in the repo that can replay an input sequence. Your doc says this itself in "What remains manual". I think this is a prerequisite, not a follow up. Without it every number is one hand-driven sample and we can't compare anything.

**5. Let's fix the pass/fail rule before we run the test, not after.** "Expand only if it removes a meaningful custom path without a material regression" doesn't bind us to anything. You can see that in the spike result: shell cache refreshes went from 1 to 63 and the doc reads it as "expected, not a win or loss". Something we can actually fail:

> p50 and p95 of `gui_render.cpu_ui_before_vulkan_begin` must not get worse at all, p99 by no more than X ms, over at least 1000 frames, same replayed input, same resolution, with the scene panel and rendering tab open. Plus the number of routing lines we delete.

## Proving we can delete the handwritten code

This is a separate proof and it has nothing to do with speed. For each piece of handwritten routing, write a headless test that shows RmlUi gives the same answer natively in one context. Your contract tests are exactly this idea and it's the best part of the PR. Driving real RmlUi headless instead of mocking it was the right call. Please keep going until every candidate has one.

What I think can go:

| Code | Where | Replaced by |
| --- | --- | --- |
| `passiveMouseMoveNeedsRender()` | `rmlui_manager.cpp:476-535` | `GetElementAtPoint(p) != GetHoverElement()`, about 60 lines |
| `trackContextFrame()` / `TrackedContextFrame` | `rmlui_manager.cpp:407-434`, 14 call sites | every field is native to one context |
| `activeOverlayContainsPoint` / `OccludesContext` | `:537-569` | native stacking and hit testing |
| `wantsCaptureKeyboard` / `wantsTextInput` / `anyItemActive` | `:571-599` | one `GetFocusElement()` |
| `RmlPanelHost` select bridge | `rml_panel_host.cpp:789-846` plus 7 users | same-document popup once the owner is window wide |

And two that stay, so nobody counts on deleting them: `GuiManager::hitTestPointer()` (`gui_manager.cpp:6084-6117`), because 6 of its 8 checks are not RML hit tests at all (capture, modal, startup, context menu, export lock, sequencer), and `isPositionOverFloatingPanel()` (`panel_registry.cpp:793-823`), which needs the outward resize margin and the rounded corner reject. The float stack order in `PanelRegistry` moves rather than disappears, we still need it for enumeration, persistence and the Python side.

One correction that actually helps your case: the "code to minimize" table lists tooltips, but RmlUi 6.2 has no tooltip API at all, zero hits in any header. So `RmlTooltipController` can't be deleted by any framework change, only the cross-context wake-up bookkeeping (`rmlui_manager.cpp:436, 445, 456`). Meanwhile the table leaves out the ~250 lines above, which are the real prize. Your case is stronger than you're making it.

## Two bugs that slipped in

These are not part of the investigation, they just came along.

**The empty state overlay is not static, and removing `request_redraw()` freezes it.** Look at `overlays/__init__.py:353`, `dash_offset` comes from `lf.ui.get_time()` with `_ANIM_SPEED = 30.0`. That's the marching dashed border, about 1.5 cycles per second. With render on demand, a draw hook asking for the next frame *is* how we do a continuous animation here. It's not a feedback loop, it's the animation.

I checked if anything else keeps it alive and no. The overlay is drawn from `lf.ui.add_hook("viewport_overlay", "draw", ...)` which goes into `PyUIHookRegistry`, but `FrameDemand::python_overlay` comes from `PyViewportDrawRegistry` and `PyTransformGizmoRegistry` (`viewport_overlay.cpp:26-27`). Different registry, both false on an empty scene. And `needsAnimationFrame()` only knows `rml_viewport_overlay_`, which is the RmlUi document, not this hook.

Your follow up trace in the doc, "one `py_redraw=true` frame, then back to `py_redraw=false` while idle", is exactly the frozen animation, written up as the success case.

Please don't just put it back unconditionally though, you were right that this pins the loop. Gate it on the overlay being visible, `:321-330` already computes that. And `_draw_drag_drop_overlay` needs the same, it animates too (pulse and bounce) and never had a `request_redraw()`, and the empty state function returns early while dragging, so fixing only one place leaves the drop overlay frozen.

**The drag and drop overlay never shows up on an idle empty scene.** `drag_drop_.pollEvents()` and `drag_drop_hovering_` are the first two lines of `GuiManager::render()` (`gui_manager.cpp:4458-4459`), and `VisualizerImpl::render()` skips `gui_manager_->render()` completely when there is no demand (`visualizer_impl.cpp:1718`). `setDragHovering` (`drag_drop_native.cpp:338`) only sets a member, it doesn't wake anything. SDL drop events set `had_event` but not `window_event` (`frame_input_buffer.hpp:64-66`), and `inputFrameRequestsRender()` wants a window, button, key or wheel event or a mouse move. During an X11 drag the pointer belongs to the drag source, so we get no motion. The drop still imports fine, the user just gets no feedback before dropping.

This one is already latent on master for a loaded scene, your change extends it to the empty scene, which is the case the overlay was written for. I'd fix it in its own commit.

Small thing on the PR text: "the default path is unchanged when the environment variable is not set" isn't quite true. Four changes ship without a gate. The menu bar hit test and the `RmlContextOwner` extraction (both intended and fine), the overlay removal, and the `belongsToDocument()` guards in `rml_right_panel.cpp`, which sit inside the commit you describe as the opt-in spike. Those guards do nothing today since the right panel context has exactly one document, but they are not gated.

## Doc fixes

The context inventory is good, I checked all 11 `createContext("...")` names against the call sites and they match. Five things to correct:

- `:428` "the overlay is static". It isn't. This is our developer reference, so as written it teaches the next person to delete `request_redraw()` from an animated hook. Please fix this line even if the code change goes away.
- `:22-24` the three performance ranges say "the available performance.log baseline", but no `performance.log` is in the repo, so I can't reproduce them. On my own capture of the same timers the numbers are quite different, for example `renderRightPanel.cached` is p50 0.05 / p90 0.08 over 268 samples against your 0.08 to 0.11, and `cpu_ui_before_vulkan_begin` is p50 0.39 / p90 1.13 / max 20.27 over 275 against your 0.6 to 0.9. Different machine and capture, that's the point. Since the doc says these "establish a baseline to preserve or improve", please give n, the percentile, the workload and the resolution, and attach the capture.
- The ownership table offers "whole shared context, safe document region, or no cache". The middle option doesn't exist in RmlUi 6.2, see above. It would need a patched RmlUi. Better to say that or remove it, because that's exactly the escape hatch someone will plan around.
- "No public operation for moving an element from one context to another" is not right as written. `Element::RemoveChild` returns `ElementPtr` (`Core/Element.h:520`) and `AppendChild` takes one (`:506`), so moving between *documents* is public. Your conclusion for the select still holds but for other reasons: the selectbox is a widget-owned child outside the normal DOM (`ElementFormControlSelect.h:97`, you need `GetNumChildren(true)` to even see it), and each context has its own `RenderManager` (`Core/Context.h:34, 245`), so geometry and texture handles end up on the wrong manager after a move. Right answer, wrong reason, and the reason is what people plan against.
- `:216` still lists `test_menubar_resources.py` as covering "retained submenu bounds", but this same PR changed that test to assert `dropdownElementAtPoint` and `GetAbsoluteOffset` are gone.

Also in the plugin doc: it tells authors to "request a core UI capability" for a dropdown or menu that crosses panels, but `lf.ui.show_context_menu(items, screen_x, screen_y, on_action)` is already public and stubbed (`py_ui.cpp:2722`, `stubs/lichtfeld/ui/__init__.pyi:994`) and goes straight into the full window `global_context_menu`. What I'd write instead: your panel `id` is your RmlUi context name and your `lf.ui.rml.get_document()` key, so prefix it. Data model names are global today and become context-global under sharing, so prefix those too. And `show_context_menu()` is the supported cross panel popup.

## What I like

- `RmlContextOwner` is the right abstraction. Own versus borrow is explicit, copy is deleted, it only destroys when it owns and the manager is still alive, and it has `LFS_VIS_API` plus `core/export.hpp` like our Windows DLL rule needs. Shutdown order in `gui_manager.cpp:3775-3776` is safe and `tooltip_` is declared after `context_owner_` so it dies first. Nice.
- The menu bar change is simply better than what it replaced. The old rectangle walk ignored clipping, `overflow` and z-order, `ProcessMouseMove` already ran before the hit test at `rml_menu_bar.cpp:485`, and the old fallback call was already redundant. This is the shape all the other deletions should have.
- The contract tests running real RmlUi headless. That's the base for the deletion proof.
- The framing in the doc, that our multi-context model is a performance workaround and not proof that one context is too expensive, is the right question to ask. And the non-goals are honest about not having answered it yet.

## Things I checked so you don't have to worry about them

- The deferred shell frame render is correct. It looks wrong because the shell queues a background cached context (`rml_shell_frame.cpp:186`, `.foreground = false`) whose `#right-panel-region` is opaque, but the insert lands between `gui_manager.cpp:4990` and `:4991`, still before `:4995-5005`, and the only call that moved over is `reg.preload_panels(PanelSpace::SidePanel, ...)` at `:4760`, which draws nothing by contract.
- Document position can't go stale in shared mode. `screen_x/screen_y` are zeroed at `gui_manager.cpp:4779-4780` and the new `last_context_w_/h_` guard closes the rest. Both `SetDimensions` callers compute the same ints and `Context::SetDimensions` returns early on equal values.
- `belongsToDocument` can't reject a valid element. RmlUi 6.2 select popups are same-document children (`WidgetDropDown.cpp:22-24`), and the debugger attaches to the first context created, which is `startup_overlay` (`gui_manager.cpp:3276` comes before the shell and right panel init), never `right_panel`.
- Two `Rml::Initialise`/`Shutdown` suites in one binary are fine, gtest never interleaves suites and `Rml::Shutdown` resets global state.
- Includes are fine, no MSVC narrowing in braced initializers, no per-frame allocation in the trace, and the trace can't touch a destroyed context.

All static review, I didn't build or run it.

src/visualizer/gui/rmlui/rml_context_owner.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "gui/rmlui/rmlui_manager.hpp"
+
+#include <cassert>
+#include <utility>
+
+namespace lfs::vis::gui {
+
+    RmlContextOwner::RmlContextOwner(RmlUIManager* manager, std::string context_name)
+        : manager_(manager), context_name_(std::move(context_name)) {
```
MrNeRF, line 16: clang-format hasn't been run. 14 spots across 5 files, all new from this PR, master is clean:

- `rml_context_owner.cpp:16, 21`
- `rml_right_panel.cpp:614-616, 643`
- `rmlui/rmlui_manager.cpp:670, 681`
- `tests/test_panel_registry_render_paths.cpp:19`
- `tests/test_rml_shared_context_contract.cpp:131-134`

Our `clang-format.yml` only runs on push to master and auto commits, so it won't block anything here, it just shows up later as a `style: apply clang-format` commit. House rule is to run it before committing.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -106,10 +109,11 @@ namespace lfs::vis::gui {
         active_tab_.clear();
         if (rml_manager_)
             rml_manager_->releaseCachedVulkanContext(direct_cache_);
-        if (rml_context_ && rml_manager_)
+        if (!shared_context_ && rml_context_ && rml_manager_)
```
MrNeRF, line 112: The shared path in `shutdown()` correctly skips `destroyContext`, but it never calls `rml_context_->UnloadDocument(document_)`. And it leaves the `right_panel_tabs` data model bound to `&tabs_` and `&active_tab_`, with `BindEventCallback` lambdas capturing `this`.

That's only safe because `gui_manager.cpp:3775-3776` runs both shutdowns back to back and the shell's `destroyContext` tears the whole context down right after. If anyone ever reorders that or shuts down only one part, we end up with a live visible document in `shell_frame` whose data model points at cleared members.

`reloadResources()` at `:130-140` unloads properly, so the asymmetry between the two is the giveaway.

src/visualizer/gui/rml_right_panel.cpp
```diff
@@ -448,6 +476,8 @@ namespace lfs::vis::gui {
             input_dirty_ = true;
         }
         auto* hover = pointer_in_context ? rml_context_->GetHoverElement() : nullptr;
```
MrNeRF, line 478: Two notes on this block.

These `belongsToDocument()` guards, here and at `:382`, `:433`, `:574`, `:581`, `:590`, `:608`, ship unconditionally, inside the commit described as the opt-in spike. They do nothing today because the `right_panel` context has exactly one document, so the check is always true. Still, they are not behind the flag, so the PR text isn't quite accurate. Either mention them or move them behind the same gate.

Second, at `:465` the condition `if (pointer_in_context || shared_context_)` makes the `else if (rml_pointer_inside_)` branch unreachable in shared mode. Small impact, since window space `ProcessMouseMove` clears hover natively and `input_dirty_` still comes from the `over_interactive` transition at `:502-504`, but worth cleaning up.

Related: when a floating panel covers the right panel, `gui_manager.cpp:4908-4919` sets `mouse_x = mouse_y = -1.0e9f`. In the normal path that lands in the leave branch. In shared mode it becomes `ProcessMouseMove(-1000000000, -1000000000)` on the shell context every frame the pointer moves, which pushes that value through `GetElementAtPoint`'s transform math for every document in the window. `ProcessMouseLeave()` would be the right call there.

src/visualizer/gui/rmlui/rmlui_manager.cpp
```diff
@@ -656,6 +662,27 @@ namespace lfs::vis::gui {
         });
     }
 
+    void RmlUIManager::traceContextUpdateDemand(const Rml::Context* const context) {
```
MrNeRF, line 665: Nice and cheap, I like that it only logs on state change and does nothing when disabled. Two accuracy caveats worth writing into the doc, since we're using this trace as evidence:

1. `GetNextUpdateDelay()` returns `next_update_timeout`, which `Context::Update()` resets. This trace fires from the queue path, including on frames where the cached path skips `Update()` entirely (`rml_shell_frame.cpp:172-174` only updates when `refresh_cache`). On those frames we report the delay from the previous update.
2. `context_update_demand_states_` is only erased in `destroyContext` and `shutdown`, so a context that stops being queued keeps its last state and then emits a fake transition when it comes back.

Neither affects behaviour, but the "none of the queued contexts changed from idle" conclusion in the doc rests on exactly this.

francescofugazzi:
Thanks again for the detailed review. I followed the “measure first, do not maintain two runtime paths” direction rather than continuing the shared-context spike.

I am not updating the draft PR with a new shared-context implementation. The investigation now has enough evidence to decide whether that implementation is worth writing.

## Native shared-context contracts

I extended the live headless RmlUi contract tests. They now cover:

- stacking, hit testing, click delivery, and document unloading;
- absolute window-space document bounds;
- document-root overflow clipping;
- native focus moving between documents;
- `GetElementAtPoint()` versus `GetHoverElement()` as the pointer-transition decision;
- ownership of the widget-created native `selectbox` popup by its source document;
- submenu hover/click behavior across a popup boundary;
- borrowed context ownership.

The focused contract suite passes 9/9 tests locally, and the complete local GUI tier passes with `ctest -L gui`.

This establishes that, inside one context, RmlUi can replace the candidate manual mechanics:

| Candidate | Native evidence |
| --- | --- |
| `passiveMouseMoveNeedsRender()` | Hit element and current hover identify a surface transition. |
| `trackContextFrame()` / overlay occlusion | Document bounds, stacking, and native hit testing provide the active surface. |
| Focus aggregators | One `GetFocusElement()` identifies the active document. |
| `RmlPanelHost` select bridge | The generated `selectbox` remains owned by the source document. |
| Submenu routing | Native hover and bubbling work once the popup remains explicitly open. |

The tooltip controller itself is intentionally not counted as removable: RmlUi 6.2 does not provide a tooltip API. Only cross-context wake-up bookkeeping would disappear in a successful shared model.

## Deterministic replay

I added a local deterministic replay runner:

- waits for a loaded model;
- snapshots the current resolution and stops if it changes;
- selects `lfs.rendering`;
- warms up for 120 frames;
- records 1,000 frames each for `idle`, `rendering_panel_hover`, and `theme_change`;
- reports p50/p95/p99/max CPU UI time plus queued/update/record/cache-hit/cache-refresh counts per context.

The run used the same loaded PLY, a maximized 2560×1540 window, and the Rendering tab.

System used:

| Item | Value |
| --- | --- |
| OS | Windows 11 Pro, build 26200 |
| CPU | AMD Ryzen 7 5800H, 8 cores / 16 logical processors |
| Memory | 64 GB installed, 59.9 GiB usable |
| GPU | NVIDIA GeForce RTX 3070 Laptop GPU |
| Build | Release, Ninja, MSVC 14.44, `x64-windows`, `BUILD_GUI_TESTS=ON`, `BUILD_TESTS=OFF` |

## Measured current topology

Latest replay capture:

| Workload | p50 | p95 | p99 |
| --- | ---: | ---: | ---: |
| `idle` | 0.770 ms | 0.939 ms | 1.271 ms |
| `rendering_panel_hover` | 7.061 ms | 7.646 ms | 8.450 ms |
| `theme_change` | 0.795 ms | 0.956 ms | 2.376 ms |

The hover workload confirms that the existing topology is selective:

| Context | Refreshes over 1,000 hover frames |
| --- | ---: |
| `lfs.rendering` | 104 |
| `right_panel` | 72 |
| `status_bar` | 57 |
| `scene_panel_native` | 9 |
| `viewport_overlay` | 1 |
| `shell_frame` | 0 |
| `menu_bar` | 0 |

So the current implementation normally refreshes only dirty surfaces and composites cached textures for the rest.

## Full-window shared-context estimate

As suggested, I measured the current-tree cost of forcing all visible contexts dirty once through the theme-change workload. The first forced refresh was repeatable:

| Context | Update | Record |
| --- | ---: | ---: |
| `lfs.rendering` | 2.122 ms | 0.485 ms |
| `viewport_overlay` | 1.582 ms | 0.040 ms |
| `scene_panel_native` | 1.485 ms | 0.079 ms |
| `menu_bar` | 0.432 ms | 0.074 ms |
| `right_panel` | 0.182 ms | 0.044 ms |
| `status_bar` | 0.121 ms | 0.044 ms |
| `shell_frame` | 0.019 ms | 0.014 ms |
| **Total** | **5.943 ms** | **0.780 ms** |

This is not presented as an actual shared-context result. It is the current-tree estimate of the work that one shared context must perform whenever a hover/style/input change requires `Context::Update()` and `Context::Render()`.

## Decision

I am stopping before implementing the hard shared-context branch.

RmlUi 6.2 has no public partial-render or damage API: a shared `Context::Render()` records all visible documents. The current hover path has a measured p50 of 7.061 ms while it updates only dirty panel contexts. A shared context would need to traverse the full visible tree on hover, introducing roughly 5.943 ms of full-tree update work plus 0.780 ms of record work, while retaining the Python-panel/layout work already present in the hover workload.

The manual routing code is worth simplifying only if the shared model meets the agreed “no p50/p95 regression” gate. These measurements make that implausible on the current UI topology, so writing the full shared path would be a large refactor to demonstrate an already-predicted regression.

The conclusion is therefore:

- retain the panel-sized cache topology;
- retain the native contract tests as future deletion evidence;
- do not revive the old shell/right-panel-only spike, since it omits the dominant Rendering and scene documents;
- revisit only if RmlUi gains partial rendering/damage tracking, or if the UI topology changes enough to materially alter the full-tree cost.

I have documented the benchmark protocol, the local measurements, native-contract coverage, and this no-go decision locally. I have intentionally not converted the draft PR into a new shared-context implementation.

PR #1504, opened 2026-07-30
Scene Graph Camera Group Active Count

kmeirlaen:
## Summary

When importing a dataset, the scene graph displays a camera group such as `Training (103)` using the total number of camera children. After disabling cameras for training, the row still displays `Training (103)`, making it look like all cameras are still active.

Camera group rows in the scene graph now show active training cameras (for example `Training (100/103)` when some cameras in the group are disabled. Fully enabled groups keep the existing compact total count.

## Reproduction Steps

1. Load a dataset with a camera group, for example `Training (103)`.
2. Disable one or more cameras for training from the scene graph context menu.
3. Observe that the camera group row still shows the original total count.

## Root Cause

The RmlUI scene graph label for `CAMERA_GROUP` nodes used `node->children.size()`. That counted direct child nodes but did not distinguish cameras enabled for training from cameras disabled for training. The core scene already tracked per-camera `training_enabled` state and exposed a whole-scene active camera count, but there was no per-camera-group active/total query for the scene tree label.

## What This Patch Fixes

- Adds `Scene::CameraTrainingCounts` and `Scene::getCameraTrainingCounts(NodeId)` to report enabled and total direct camera children for a camera group.
- Updates the scene graph camera-group label to display:
  - `Training (103)` when all direct cameras are enabled.
  - `Training (100/103)` when only 100 of 103 direct cameras are enabled.
- Ignores malformed non-camera direct children in camera group counts.
- Leaves CLI, Python API, and MCP behavior unchanged.

### DIFF
--- a/src/core/include/core/scene.hpp
+++ b/src/core/include/core/scene.hpp
@@ -150,6 +150,11 @@ namespace lfs::core {
             bool has_selection = false;
         };
 
+        struct CameraTrainingCounts {
+            size_t enabled = 0;
+            size_t total = 0;
+        };
+
         enum class MutationType : uint32_t {
             NODE_ADDED = 1 << 0,
             NODE_REMOVED = 1 << 1,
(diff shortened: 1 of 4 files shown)


MrNeRF:
lgtm! Thx!

PR #1527, opened 2026-08-01
Complete GUI localization coverage and add headless contracts

francescofugazzi:
## Summary

This pull request completes the GUI localization pass and strengthens localization handling across native C++, RML, Python panels, asynchronous operations, MCP/UI bridges, and shipped locale resources.

It also introduces headless localization contracts and hardcoded-UI audits to prevent localization regressions from being reintroduced.

The English locale currently contains 2042 localization keys, with matching key sets and placeholder contracts across all shipped locales.

## Localization coverage

The changes move remaining user-facing GUI text into the localization resources, including:

- viewport and rendering controls;
- viewport and depth controls;
- LOD controls;
- video extractor controls and dropdown options;
- sequencer labels and status text;
- asset manager messages;
- export and import panels;
- plugin marketplace status messages;
- operator labels and cancellation messages;
- asynchronous task progress and completion states;
- MCP task and export status;
- COLMAP export dialogs;
- startup and overlay UI;
- tooltips and sidebar labels;
- RML attributes and option labels.

All shipped locales are kept structurally aligned with `en.json`.

## Runtime language switching

Runtime language switching now updates:

- cached native panels;
- cached Python panels;
- live RML documents;
- RML text translation directives;
- RML `title` and `placeholder` attributes;
- translated `<option>` elements;
- cached selected values in RML select controls;
- startup language selection;
- menus, toolbars, HUD labels, and overlays.

Dynamic application-owned content is preserved during localization refreshes. Values such as paths, progress values, frame counts, FPS values, and other runtime-generated content are not overwritten by stale localized text.

The refresh path also handles the initial language generation correctly, including startup when the generation value is zero.

## Python panel localization

Python panels now use explicit localization mappings for their labels.

The Rendering Panel includes dedicated keys for:

- environment map;
- environment exposure;
- environment rotation;
- rendering and viewport controls;
- LOD controls;
- mesh and post-processing controls.

The localization contracts inspect literal Python `LOCALE_KEY` dictionaries so indirect mapping references cannot bypass validation.

Python formatting is protected with a safe formatter that supports both positional and named placeholders and gracefully handles malformed translator-controlled format strings.

## Stable task outcomes

Task state now separates stable machine-readable outcomes from localized display stages.

Export and dataset operations expose stable outcomes such as:

- `idle`;
- `running`;
- `completed`;
- `failed`;
- `cancelled`.

Localized stage text remains presentation-only and is not used for program logic.

The export outcome is propagated consistently through:

- native asynchronous task state;
- Python store bindings;
- Python overlay state;
- Python-facing UI bridges.

## COLMAP export handling

COLMAP overwrite dialogs now detect the actual source sparse format.

Binary sources report:

- `cameras.bin`;
- `images.bin`;
- `points3D.bin`.

Text sources report:

- `cameras.txt`;
- `images.txt`;
- `points3D.txt`.

The message uses a shared positional localization placeholder and is translated in every shipped locale.

## Localization validation tools

The localization validation suite now checks:

- English and shipped locale key parity;
- placeholder parity;
- plural-form compatibility;
- malformed format strings;
- one JSON key per physical line;
- JSON key indentation;
- direct C++ localization lookups;
- direct Python translation calls;
- Python `LOCALE_KEY` dictionaries;
- RML translation directives;
- RML translated attributes;
- runtime RML refresh wiring;
- dynamic-content preservation logic;
- translated select option handling;
- cached native panel refresh behavior;
- cached Python panel refresh behavior;
- operator label localization;
- stable task outcome usage;
- hardcoded UI candidates;
- declared UI fallback exceptions;
- watch-directory message formatting;
- language generation propagation.

The hardcoded UI audit supports CI enforcement through `--fail-on-candidates`.

## CI and headless contracts

The repository includes a headless localization contract target that can be run without Torch-dependent tests.

The contracts are also suitable for CI execution and fail when localization regressions are detected.

Example commands:

```bash
python tests/python/test_localization_contracts.py
python tools/check_locale_completeness.py
python tools/check_ui_hardcoded.py
ctest --test-dir build/tests --output-on-failure -R LocalizationContracts
```

## Documentation

Localization conventions, validation tools, hardcoded UI audit rules, allowlist policy, formatting rules, runtime refresh behavior, and language-specific grammar handling are documented in:

- `CONTRIBUTING.md`;
- `docs/building_and_distribution.md`;
- the localization testing documentation.

## Validation performed

The following checks pass locally:

- localization contract tests;
- locale completeness validation;
- hardcoded UI audit;
- indirect Python localization mapping validation;
- placeholder validation;
- plural-form validation;
- one-key-per-line validation;
- Python syntax checks;
- clang-format validation;
- `git diff --check`.

A full application build and runtime verification should be performed before merge, with particular attention to:

- runtime language switching while multiple panels are open;
- Video Extractor dropdown refreshes;
- Sequencer and Asset Manager refreshes;
- COLMAP binary and text overwrite dialogs;
- dynamic RML content preservation;
- startup localization warnings.


MrNeRF (commented):
Thanks for this huge effort — this is a big and genuinely useful pass. The locale data itself is in really good shape: all 10 files have exact key parity with en.json, zero placeholder mismatches, correct terminology in ja/ko/zh, and the deferred-refresh idea is the right approach. Dutch and Italian are particularly strong.

A few things need fixing before merge (details in the inline comments):

**Must fix**
- Task stage strings are now translated, but `src/app/mcp_runtime_tools.cpp` and `export_panel.py` still compare against English `"Complete"` / `"Failed"` / `"Cancelled"` — in a non-English session finished jobs report as idle and exports never register in the Asset Manager. Stage should stay a stable token, translated only for display.
- `LocalizationManager` has no locking but is now called from worker threads (exports, video encode, folder scans) — a language switch during an export can crash.
- `LOCF` is a bare `std::vformat`: one broken placeholder in a locale file (or via `lf.ui.loc_set`) throws, and two call sites sit inside catch blocks on worker threads, which terminates the app.
- The new contract test spawns `python` and fails on python3-only machines (`sys.executable`).
- `selection_controls.py` tracks `language_generation` but never dirties its labels — the stale-language bug stays in that panel.
- The startup overlay language picker doesn't trigger the new deferred refresh — first-run users keep an English UI until restart.
- Neither new check runs in CI yet — wiring `check_ui_hardcoded` and `LocalizationContracts` into the workflows would let them actually guard.

**Also worth a pass**
- de/fr/es still have 25 `runtime.*` values in plain English (it has them all translated — nice reference). A few German strings lost umlauts: `"Schluesselbild"`, `"Alle loeschen"` → `"Keyframe"`, `"Alle löschen"`.
- Small pluralization losses ("Showing 1 assets", "plugins"), and en.json `disk_space_dialog.checkpoint_save_failed` = `"Checkpoint Save Failed (Iteration"` is a fragment — `"… (Iteration {})"` would make it translatable.
- The scanner's clean result is mostly it being too forgiving (single words always skipped) — see inline notes; still valuable as a review queue.
- Not from this PR: the Ubuntu CI failures are SDL3/vcpkg infrastructure — a re-run should clear them.

With the stage-token, threading and format-fallback issues sorted, this will be a great step for the app. Thanks again!

src/visualizer/gui/async_task_manager.cpp
```diff
@@ -1196,30 +1200,30 @@ namespace lfs::vis::gui {
...
-                    error_msg = "Export cancelled by user";
+                    error_msg = LOC(lichtfeld::Strings::Runtime::EXPORT_CANCELLED);
                 }
 
                 if (success) {
                     LOG_INFO("Export completed: {}", lfs::core::path_to_utf8(path));
                     {
                         const std::lock_guard lock(export_state_.mutex);
-                        export_state_.stage = "Complete";
+                        export_state_.stage = LOC(lichtfeld::Strings::Runtime::TASK_COMPLETE);
```
MrNeRF, line 1335: These stage values are also read as plain English elsewhere: `mcp_runtime_tools.cpp` checks `stage == "Complete"` / `"Failed"` / `"Cancelled"`, and `export_panel.py:758` checks `"Complete"` before registering the export. Once translated (de: "Abgeschlossen"), non-English sessions never see a job as finished. Suggest keeping `stage` as a fixed token and translating only at display time.

src/visualizer/gui/async_task_manager.cpp
```diff
@@ -932,15 +935,15 @@ namespace lfs::vis::gui {
                         }
                     }
                 } catch (const std::exception& e) {
-                    error_msg = std::string("COLMAP export crashed with exception: ") + e.what();
+                    error_msg = LOCF(lichtfeld::Strings::Runtime::TASK_FAILED_DETAIL, e.what());
```
MrNeRF, line 1038: Two risks here: this runs in a `catch` on a worker thread, and `LOCF` is a bare `std::vformat` that throws on a broken placeholder — an uncaught throw here ends the process. Also `LocalizationManager` has no locking, so a language switch during an export races with these lookups.

src/core/event_bridge/localization_manager.hpp
```diff
@@ -53,6 +54,12 @@ namespace lfs::event {
         mutable std::unordered_map<std::string, std::string> overrides_;
     };
 
-#define LOC(key) lfs::event::LocalizationManager::getInstance().get(key)
+    template <typename... Args>
+    [[nodiscard]] inline std::string formatLocalized(const std::string_view key, Args&&... args) {
+        return std::vformat(LocalizationManager::getInstance().get(key), std::make_format_args(args...));
```
MrNeRF: `std::vformat` throws `std::format_error` if a translation has a bad placeholder (and plugins can set arbitrary strings via `lf.ui.loc_set`). A try/catch falling back to the English text would make bad locale data harmless instead of fatal. Same applies to the `fmt::runtime(LOC(...))` site in `rml_sequencer_overlay.cpp`.

tests/python/test_localization_contracts.py
```diff
@@ -0,0 +1,116 @@
...
+            if path.suffix not in {".cpp", ".hpp", ".h", ".py"}:
+                continue
+            source = path.read_text(encoding="utf-8", errors="ignore")
+            for pattern in patterns:
+                for key in pattern.findall(source):
+                    assert key in keys, f"{path}: missing {key}"
+
+
+def test_hardcoded_ui_audit_has_no_candidates():
+    result = subprocess.run(["python", str(ROOT / "tools" / "check_ui_hardcoded.py")],
```
MrNeRF: `"python"` doesn't exist on most Linux boxes (only `python3`), so this test fails with FileNotFoundError. `sys.executable` fixes it.

tests/python/test_localization_contracts.py
```diff
@@ -0,0 +1,116 @@
...
+    else:
+        yield prefix, value
+
+
+def _load(locale):
+    return json.loads((LOCALES / f"{locale}.json").read_text(encoding="utf-8"))
+
+
+def _fields(text):
+    return {name.split(".", 1)[0].split("[", 1)[0] for _, name, _, _ in string.Formatter().parse(text) if name}
```
MrNeRF: Filtering on `if name` drops auto-numbered `{}` fields, so `"Failed: {}"` vs `"Fehler: {} und {}"` pass this check. `check_locale_completeness.py` keeps them — same approach would work here.

tests/python/test_localization_contracts.py
```diff
@@ -0,0 +1,116 @@
...
+              "processed": 2, "total": 4, "added": 1, "discovered": 4,
+              "created": 1, "skipped": 3, "status": "ok", "error": "failure"}
+    for path in sorted(LOCALES.glob("*.json")):
+        messages = json.loads(path.read_text(encoding="utf-8"))["watch_dirs"]
+        for key, text in messages.items():
+            if key.startswith("scan_"):
+                text.format(**values)
+
+
+def test_language_generation_is_part_of_cached_localized_ui_state():
```
MrNeRF, line 192: This only checks the substring exists in the file, so it also passes when the wiring is broken (see the note in `selection_controls.py`). Comparing rendered labels across a simulated language switch would catch the real thing.

src/python/lfs_plugins/selection_controls.py
```diff
@@ -286,6 +287,7 @@ def _refresh_depth_state(self):
 
     def _state_items(self):
         return (
+            ("language_generation", RuntimeState.language_generation.value),
```
MrNeRF, line 290: `language_generation` goes into the state key here, but `_dirty_changed_fields()` has no entry for it — so a language switch dirties nothing and the labels stay in the old language. `depth_view_controls.py` maps it to its label fields; or use `_dirty_all()` like `active_tool` does.

src/python/lfs/py_ui.cpp
```diff
@@ -4839,6 +4839,8 @@ namespace lfs::python {
                         if (auto* const gui_manager = get_gui_manager())
                             gui_manager->ensureCjkFontsLoaded();
                     lfs::vis::publish_language_generation();
+                    if (auto* const gui_manager = get_gui_manager())
```
MrNeRF, line 4846: The startup overlay's language dropdown (`startup_overlay.cpp:73`) needs this same call — right now picking a language there leaves the shell (menubar, overlays, panels) in the old language until restart.

src/python/lfs_plugins/plugin_marketplace_panel.py
```diff
@@ -438,7 +438,11 @@ def _build_card_record(self, entry, card_id, mgr,
         status_text = ""
         status_class = "status-muted"
         if is_installed:
-            state_str = plugin_state.value if plugin_state else tr("plugin_manager.status_not_loaded")
+            state_str = (
+                tr("plugin_manager.status_active")
+                if plugin_state == PluginState.ACTIVE
```
MrNeRF: This used to show the real state; now error/installing/loading/disabled all read "not loaded", so a plugin that failed to load looks merely inactive. Per-state keys would keep the error visible.

src/python/lfs_plugins/plugin_marketplace_panel.py
```diff
@@ -937,15 +941,13 @@ def _update_catalog_status(self, doc, entry_count: int, is_loading: bool, regist
             return
 
         if is_loading and entry_count == 0:
-            text = "Fetching plugin registry..."
+            text = lf.ui.tr("plugin_marketplace.loading")
             tone = "status-info"
         elif registry_loaded:
-            noun = "plugin" if entry_count == 1 else "plugins"
-            text = f"Registry loaded: {entry_count} {noun} in the marketplace catalog."
+            text = lf.ui.tr("plugin_marketplace.registry_loaded").format(count=entry_count)
```
MrNeRF: Small one: the singular got lost — 1 plugin now reads "plugins".

src/visualizer/gui/rmlui/resources/menubar.rml
```diff
@@ -52,28 +52,28 @@
                 <div id="menu-window-split-view"
                      class="icon-btn menu-window-btn menu-window-utility-btn"
                      data-action="toggle_independent_split_view"
-                     title="Independent Split View">
+                     title="@tr:ui.independent_split_view">
```
MrNeRF, line 55: RmlUi translates text nodes only, not attributes — these `title="@tr:..."` tooltips show the raw key on hover (three get repaired later, but only after a state change). Needs the programmatic tooltip route.

src/visualizer/gui/editor/python_editor.cpp
```diff
@@ -2398,20 +2399,20 @@ namespace lfs::vis::editor {
             const bool has_selection = hasEditableSelection();
             std::vector<gui::ContextMenuItem> items;
             if (has_selection && !read_only) {
-                items.push_back(gui::ContextMenuItem{.label = "Cut", .action = "cut"});
+                items.push_back(gui::ContextMenuItem{.label = lfs::event::LocalizationManager::getInstance().get("common.cut"), .action = "cut"});
```
MrNeRF, line 2402: `common.cut` / `common.copy` / `common.paste` / `common.select_all` aren't in any locale file yet, so the context menu shows the raw keys.

src/visualizer/gui/panels/python_console_panel.cpp
```diff
@@ -458,16 +458,18 @@ namespace {
 
         std::vector<lfs::vis::gui::ContextMenuItem> items;
         if (has_selection) {
-            items.push_back(lfs::vis::gui::ContextMenuItem{.label = "Copy", .action = "copy"});
+            items.push_back(lfs::vis::gui::ContextMenuItem{
+                .label = lfs::event::LocalizationManager::getInstance().get("common.copy"),
```
MrNeRF, line 462: Same as the editor: `common.copy`, `common.copy_all`, `common.paste` don't exist in the locale files yet.

src/visualizer/gui/panels/python_console_panel.cpp
```diff
@@ -1501,7 +1503,7 @@ namespace lfs::vis::gui::panels {
 
     void PythonConsoleState::runScriptAsync(const std::string& code) {
         if (script_running_.load()) {
-            addError("A script is already running");
+            addError(LOC("python_console.running"));
```
MrNeRF: This replaced "A script is already running" with `python_console.running` = "Running..." — as an error message it now reads like normal progress.

src/python/lfs_plugins/sequencer_ops.py
```diff
@@ -41,7 +41,7 @@ def execute(self, context):
 class PlayPauseOperator(Operator):
     """Toggle sequencer playback."""
 
-    label = "Play/Pause"
+    label = "action.sequencer_play_pause"
```
MrNeRF: This key doesn't exist (`input_settings.action.sequencer_play_pause` is the real one) — and nothing translates operator labels anyway: `py_ui.cpp` stores them verbatim, so MCP `list_operators` now returns raw keys for all three. Maybe keep plain labels until the operator path can translate them.

src/visualizer/gui/scene_panel_native.cpp
```diff
@@ -314,18 +311,13 @@ namespace lfs::vis::gui {
...
-                return std::format("No buffered CLI logs · Level {}", level_label);
+                return LOCF("runtime.cli_no_logs", level_label);
             if (displayed_entry_count < entry_count) {
-                return std::format("{} buffered log {} · showing latest {} · Level {}",
-                                   entry_count,
-                                   entry_count == 1 ? "entry" : "entries",
-                                   displayed_entry_count,
-                                   level_label);
+                return LOCF("runtime.cli_logs_latest", entry_count,
+                            entry_count == 1 ? "entry" : "entries", displayed_entry_count, level_label);
```
MrNeRF: English `entry`/`entries` gets inserted into translated sentences (ja renders `5件のログ（entries）`). The `log_entry_count`/`log_entries_count` key pair used further down solves exactly this.

src/visualizer/gui/scene_panel_native.cpp
```diff
@@ -233,13 +232,11 @@ namespace lfs::vis::gui {
                     item.metadata.label.empty() ? std::string("Untitled Change") : item.metadata.label;
                 const bool is_next = index == 0;
                 const std::string stack_line = is_next
-                                                   ? std::format("NEXT {} · Top of stack",
-                                                                 kind == "undo" ? "UNDO" : "REDO")
+                                                   ? LOCF("runtime.history_next", kind == "undo" ? "UNDO" : "REDO")
```
MrNeRF: Same pattern: `UNDO`/`REDO` stay English inside a translated sentence.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -4552,6 +4556,12 @@ namespace lfs::vis::gui {
             pollDevResourceHotReload();
         }
 
+        if (pending_localization_ui_refresh_ && !rmlui_manager_.wantsTextInput() &&
+            !rmlui_manager_.anyItemActive()) {
```
MrNeRF: Two small things: this guard checks less than `shouldDeferDevResourceHotReload()` (no menu/context-menu/modal checks), and the flag is cleared before the reload runs — `RmlModalOverlay::reloadResources()` skips while a modal is active, so that refresh would be lost silently.

tools/check_ui_hardcoded.py
```diff
@@ -0,0 +1,160 @@
...
+STRING_LITERAL = re.compile(r'"((?:\\.|[^"\\])*)"')
+RML_TEXT = re.compile(r">([^<>{}][^<>{}]*[A-Za-z][^<>{}]*)<")
+UI_SINK = re.compile(
+    r"\b(SetText|SetInnerRML|body_rml|\.title\s*=|\.label\s*=|bind_func\(|"
+    r"\.stage\s*=|\.error\s*=|\bstatus\s*=|_set_scan_log_entry\(|addError\(|fail_start\(|std::unexpected\(|"
+    r"set_text\(|set_inner_rml\(|message_dialog\()"
+)
+IGNORE_LINE = re.compile(r"\b(LOG_(?:TRACE|DEBUG|INFO|WARN|ERROR)|#include)\b")
+TECHNICAL_LITERAL = re.compile(
+    r"^(?:[A-Za-z_][A-Za-z0-9_./:-]*|[A-Z0-9_]{2,}|"
```
MrNeRF: This pattern treats every single word as technical, so `Cancel`, `Export`, `Overview`, `Drill-down`… can never be reported (the VRAM HUD tabs in `viewport_overlay.rml:1024` are live examples). Together with sinks missing `_set_status(`, `on_progress(`, f-strings, class-body `label =` and RML `title=` attributes, a clean run doesn't prove coverage yet — great as a review queue, but worth tightening before it becomes a hard gate.

tools/check_ui_hardcoded.py
```diff
@@ -0,0 +1,160 @@
...
+                continue
+            if path.suffix in SOURCE_SUFFIXES:
+                findings.extend(scan_source(path, allowlist, allow_patterns))
+            elif path.suffix in RML_SUFFIXES:
+                findings.extend(scan_rml(path, allowlist, allow_patterns))
+
+    if findings:
+        print(f"Likely hardcoded UI strings: {len(findings)}")
+        for finding in findings:
+            relative_path = finding.path.relative_to(PROJECT_ROOT).as_posix()
```
MrNeRF, line 254: A relative `--root src/core` crashes in `relative_to()` — resolving the path first would fix it.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -1275,9 +1275,7 @@ def get_asset_results_summary(self) -> str:
         total = self._last_asset_match_count
```
MrNeRF, line 1276: Lost the singular and thousands separator: "Showing 1 assets", "Showing 12345 assets". A `{count:,}` plus a one/many key pair (like `watch_dirs.scan_preparing_one/_many`) restores both.

tests/CMakeLists.txt
```diff
@@ -3,6 +3,29 @@
...
+        TIMEOUT 30
+    )
+    add_custom_target(test_localization_contracts
+        COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -R LocalizationContracts
+        DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/CTestTestfile.cmake
+        USES_TERMINAL
+    )
+    message(STATUS "Registered headless localization contract tests (no LibTorch)")
+    return()
+endif()
```
MrNeRF, line 29: With `BUILD_LOCALIZATION_TESTS=ON` this early `return()` silently drops the whole gtest suite (and the unicode test below) if other test options are also on. Also worth wiring this target and `check_ui_hardcoded.py` into CI — neither runs anywhere yet.

src/visualizer/gui/rml_menu_bar.cpp
```diff
@@ -719,13 +720,13 @@ namespace lfs::vis::gui {
             struct NavButtonSpec {
```
MrNeRF, line 735: These four lookups + string builds now run every frame via `rebuildToolbarButtons()` — caching behind `language_generation` would keep the draw path clean.

src/visualizer/gui/vram_hud_overlay.cpp
```diff
@@ -990,7 +993,8 @@ namespace lfs::vis::gui {
         const auto process_total = bestProcessTotal(s);
 
         if (iteration_label_) {
-            setText(iteration_label_, cached_iteration_text_, std::format("iter {}", s.iteration));
+            setText(iteration_label_, cached_iteration_text_,
```
MrNeRF, line 1011: This formats + looks up every HUD update — and `status.iteration` is "Iteration:" (a form label with a colon), which widens the compact HUD.

src/python/lfs_plugins/rendering_panel.py
```diff
@@ -191,6 +191,10 @@ def _set_theme_vignette_style(*, intensity=None, radius=None, softness=None):
     "grid_opacity": "main_panel.grid_opacity",
     "focal_length_mm": "main_panel.focal_length",
     "render_scale": "main_panel.render_scale",
+    "environment_mode": "main_panel.environment",
+    "environment_map_path": "main_panel.preset",
+    "environment_exposure": "main_panel.ppisp_exposure",
+    "environment_rotation_degrees": "transform.rotation",
```
MrNeRF: In Chinese this key is `旋转：` (fullwidth colon) and `_entry_label` only strips ASCII `:`, so the row renders as `旋转：:`.

src/visualizer/gui/rmlui/elements/scene_graph_element.cpp
```diff
@@ -2409,13 +2409,13 @@ namespace lfs::vis::gui {
 
             const std::string export_button = "Overwrite";
             lfs::core::ModalRequest request;
-            request.title = "Export COLMAP sparse";
+            request.title = LOC(lichtfeld::Strings::Window::EXPORT);
```
MrNeRF, line 2392: The title became generic ("Export" instead of which export), while "Overwrite", "Cancel" and "This writes" in the same modal stay English — worth finishing this dialog in one go.

src/python/lfs_plugins/import_panels.py
```diff
@@ -2137,7 +2141,7 @@ def _set_scan_state(
     def _set_scan_log_entry(self, index: int, path: str, status: str) -> None:
         with self._scan_state_lock:
             while len(self._scan_log) <= index:
-                self._scan_log.append({"status": "Queued", "path": ""})
+                self._scan_log.append({"status": _tr("watch_dirs.scan_log_queued"), "path": ""})
```
MrNeRF: `_tr()` is called per padded row while holding `_scan_state_lock` — hoisting it above the loop avoids repeated FFI calls under the lock.

src/python/lfs_plugins/depth_view_controls.py
```diff
@@ -91,8 +96,15 @@ def visible(self):
...
         model.bind_func("depth_view_tool_label", lambda: _ui_label("toolbar.depth_map", "Depth Map"))
+        model.bind_func("ui_color_label", lambda: _ui_label("ui.color", "Color"))
+        model.bind_func("ui_gray_label", lambda: _ui_label("ui.gray", "Gray"))
+        model.bind_func("ui_near_label", lambda: _ui_label("ui.near", "Near"))
+        model.bind_func("ui_far_label", lambda: _ui_label("ui.far", "Far"))
         model.bind_func("depth_view_has_scene", lambda: self._has_scene)
-        model.bind_func("depth_view_disable_label", lambda: "Disable Depth Map")
+        model.bind_func(
+            "depth_view_disable_label",
+            lambda: _ui_label("toolbar.depth_mode_disable", "Disable Depth Map"),
```
MrNeRF: This borrows `toolbar.depth_mode_disable` ("Disable Depth Mode") for the Depth *Map* tool — a small mismatch, and it couples two unrelated buttons to one key.

src/python/lfs_plugins/url_downloader.py
```diff
@@ -210,7 +212,7 @@ def _download_with_progress(
                     speed_str = _format_bytes(speed) + "/s"
                     eta_str = _format_time(eta_seconds) if eta_seconds > 0 else ""
                     
-                    status = f"Downloading... {int(percent * 100)}% ({_format_bytes(downloaded)} / {_format_bytes(total_size)}) {speed_str}"
+                    status = f"{lf.ui.tr('asset_manager.import_button_downloading')} {int(percent * 100)}% ({_format_bytes(downloaded)} / {_format_bytes(total_size)}) {speed_str}"
                     if eta_str:
                         status += f" ETA: {eta_str}"
```
MrNeRF: The ` ETA: ...` suffix (and the extraction messages below: "Extracting... {i}/{total} files", "Extraction complete", "(size unknown)") are still English right next to the newly translated part.

francescofugazzi:
Thanks for the detailed review. I went through every inline thread and added follow-up commits covering the remaining functional, safety, validation, runtime-refresh, and documentation points.

## Review follow-ups addressed

### Stable task state and worker-thread safety

- Async task stages now remain stable machine-readable outcomes. Localized text is applied only at display boundaries, so MCP/export/import completion checks no longer compare translated values such as `Complete`, `Failed`, or `Cancelled`.
- `LocalizationManager` now synchronizes access for worker-thread callers and returns copied thread-local buffers, so a language switch cannot invalidate a string used by export, encoding, or scan workers.
- Localized formatting now catches invalid locale or runtime-override format strings and retries with the English fallback. If the fallback itself is malformed, the raw fallback text is returned safely instead of propagating `std::format_error`.

### Contracts, CMake, and CI

- The contract runner uses `sys.executable`, avoiding the Linux `python` versus `python3` issue.
- Placeholder validation preserves anonymous `{}` fields rather than dropping them.
- Contracts now cover real invalidation wiring for cached Python UI state, startup language selection, live RML refreshes, cached native/Python panels, translated select values, toolbar labels, HUD labels, plural forms, and safe formatting fallback.
- `BUILD_LOCALIZATION_TESTS` registers the headless CTest entry without disabling the regular test suite when other test options are enabled.
- CI now runs locale completeness, the hardcoded UI audit, and the headless localization contracts on pull requests and pushes to `master`.

### Runtime language refreshes

- Cached Python controls now dirty their localized state when `language_generation` changes.
- The startup language picker publishes the language generation and refreshes the shell immediately.
- Live RML documents are translated in place instead of being reloaded during rendering or interaction.
- RML `title` and `placeholder` directives are preserved as metadata and refreshed at runtime.
- Native select controls now refresh both their translated options and RmlUI's separate closed `selectvalue` cache.
- Video Extractor explicitly refreshes all cached select labels: extraction mode, output format, resolution, sharpness options, and window candidates.
- Navigation toolbar labels are cached by language generation rather than looked up every frame.
- The VRAM HUD caches its localized Iteration label by language generation; compact iteration labels no longer retain trailing ASCII or full-width colons.

### Remaining localization and terminology

- Restored Marketplace state visibility for unloaded, installing, loading, active, error, and disabled plugins.
- Added count-sensitive Asset Manager and Marketplace messages.
- Added Polish plural selection through `src/python/lfs_plugins/localization.py`:
  - `one` for `1`;
  - `few` for values ending in `2` through `4`, excluding `12` through `14`;
  - `other` otherwise.
- Other shipped locales retain `.one`, `.few`, and `.other` entries for key parity while currently using the applicable subset of those forms.
- Completed the runtime translations requested for German, French, and Spanish, including the German terminology corrections.
- Corrected locale-specific punctuation handling, including Chinese full-width colons.
- Completed the COLMAP export dialog, console labels, log-entry wording, downloader/extraction feedback, and dedicated Depth Map disable wording.

### Hardcoded UI audit and documentation

- Tightened the hardcoded UI scanner to inspect additional native, RML, Python, status/progress, f-string, class-label, and title-attribute sinks.
- Added regression coverage for common scanner bypasses and fixed relative-root handling.
- Kept reviewed technical/non-UI exceptions in the documented allowlist rather than hiding ordinary UI text.
- Updated `CONTRIBUTING.md` with locale conventions, one-key-per-line formatting, allowlist policy, cached-language invalidation, count-sensitive message conventions, the Polish rule, and the process for future language-specific grammar.
- Updated the build documentation with the headless contract scope and CMake/CTest entry point.

## Validation

The following checks pass:

```bash
python tests/python/test_localization_contracts.py
python tools/check_locale_completeness.py
python tools/check_ui_hardcoded.py --fail-on-candidates
clang-format --dry-run --Werror <all modified C/C++ files>
git diff --check
```

The locale checker validates 2028 English keys across all shipped locales, and the hardcoded UI audit reports no likely user-facing hardcoded strings.

## Manual validation

Runtime language switching was manually checked with open cached panels, including Video Extractor dropdowns, toolbar tooltips, Asset Manager, Sequencer, and Python panels. No stale labels, delayed refreshes, or crashes were observed.

MrNeRF:
Thanks for the thorough follow-up — most of the round-1 list is properly closed. The `LocalizationManager` locking, the `std::format` fallback, `sys.executable`, the `selection_controls` dirty map, the marketplace per-state keys, the de/fr/es `runtime.*` translations (umlauts restored), the `tests/CMakeLists.txt` `return()`, the full-width colon, and the HUD/toolbar caching all check out. The Polish rule matches CLDR and the pl strings use the right genitive forms. Locale data re-verified clean: 2028 keys across all 10 files, no key diffs, no placeholder mismatches, no empty values, and all 923 `string_keys.hpp` constants resolve.

A few things still block merge, mostly fallout from the stage/outcome split and the new RML refresh.

## Must fix

**1. Completed exports no longer register with the Asset Manager — now in every language, including English.**
`export_panel.py:758` reads `state.get("outcome")`, but the field was only added to the C++ `ExportProgressState`. Neither dict that reaches Python publishes it: `py_store.cpp:200` (`export_progress_state_to_dict`) and `py_ui.cpp:4453` (`get_export_state`) are unchanged and emit `active/progress/stage/format/error/path` only. So `.get("outcome")` is always `None`. The old bug only hit non-English sessions; this one hits everyone.

**2. MCP/CLI dataset imports never report `finished`.**
`async_task_manager.cpp:735-787` — the `DatasetLoadStarted`/`DatasetLoadCompleted` handlers set `stage` but never `outcome`, while `mcp_runtime_tools.cpp:477-480` now keys off `outcome`. That path is live via `mcp_gui_tools.cpp:1993` and `visualizer_impl.cpp:2194` → `SceneManager::loadDataset`. Also, `outcome` isn't reset on start, so a previous failure leaves `"failed"` and the next successful load reports failed.

**3. `common.cut` / `copy` / `copy_all` / `paste` / `select_all` still exist in no locale file.**
7 sites in `python_editor.cpp:2402-2418` and `python_console_panel.cpp:461-475`; `en.json`'s `common` block has none of them, so both context menus render the raw keys. These call `LocalizationManager::getInstance().get("…")` directly, which `test_literal_localization_calls_resolve` doesn't scan (it only matches `LOC(`/`LOCF(`/`lf.ui.tr(`/`_tr(`) — worth widening that pattern.

**4. `@tr:` attributes show the raw key until the user changes language.**
`gui_manager.cpp:4559` only refreshes when `language_generation` differs from `localized_rml_language_generation_`, and both start at 0 — nothing bumps the generation at startup (the only publishers are the startup picker's *change* handler, `lf.ui.set_language`, and dev locale hot-reload). A default session shows `@tr:ui.minimize`, `@tr:common.close`, `@tr:ui.independent_split_view`, `@tr:ui.hide_ui`, `@tr:ui.maximize_window` as menubar tooltips, plus `@tr:ui.search_annotations` and `@tr:common.filter` as the VRAM-HUD placeholders (`viewport_overlay.rml:1047,1067`) — those two are visible text that used to read fine in English. Priming the generation once during startup would cover it.

## Also worth a pass

- **The in-place refresh overwrites dynamic content and it doesn't heal.** `rml_document_utils.cpp:290` unconditionally re-applies the key to every `data-lfs-i18n` element. Elements that start as `@tr:` but get written programmatically snap back to their placeholder — and since the writers cache the last value in a `data-lfs-*` attribute (`video_extractor_dialog.cpp:60-77`), the next sync short-circuits and the wrong text sticks. Video extractor `#video-value`, `#output-value`, `#progress-text`, `#interval-value`, `#sharpness-mode-desc` are all in this shape. Skipping elements whose current content differs from the previous translation would fix it.
- **RML `<option>` labels don't live-refresh.** RmlUi parents options under the non-DOM `selectbox`, and `refreshLocalizedContent` recurses with DOM-only `GetNumChildren()` — which is why the video extractor needed `setCachedSelectLabels`. `dataset_import_panel.rml:90-92`, `export_panel.rml:70-71`, `plugin_marketplace.rml:18-27` and the new `rendering.rml` options stay in the old language. Recursing with `GetNumChildren(true)` would make the generic path cover them.
- **`get()` lifetime changed silently.** It now returns a pointer into an 8-slot thread-local ring; it used to be stable until `setLanguage`. Nothing violates it today (worst case is two live results at `error_event_bridge.cpp:164`), but there's no comment or assert marking the trap. Returning `std::string` would remove it outright.
- **`runtime.logs_copied` still interpolates English `entry`/`entries`** (`scene_panel_native.cpp:1036`) — de renders "5 entries in die Zwischenablage kopiert.". The same file already does it right at `:312` with `runtime.log_entry_word` / `log_entries_word`.
- **Two operator labels are raw keys with no resolver** (`sequencer_ops.py:22,33`): `mcp_operator_tools.cpp:299` returns `descriptor.label` verbatim, and `:231` composes "sequencer.add_keyframe_here could not be performed". This contradicts the comment kept on `PlayPauseOperator`.
- **Dead code from the COLMAP dialog rework**: `colmapSparseOutputFormat` / `colmapSparseOutputFileNames` / `colmapSparseOutputFileList` (`scene_graph_element.cpp:133-177`, anonymous namespace → `-Wunused-function`) and `export_panel.py:399`. Both dialogs also lost the format-dependent file list — the new static "Writes cameras.bin, images.bin, and points3D.bin" is wrong for a text-format source.

## Minor

- The CI "Audit hardcoded UI text" step runs without `--fail-on-candidates`, so it can never fail; the real gate is `test_localization_contracts.py:89`.
- `string_keys.hpp` `ALL = "all"` points at no key.
- `preserveTranslationDirectives` runs two `std::regex` passes over every RML document at load — that's on the startup path.
- Key reuse in `rendering_panel.py:194-197`: `environment_map_path` → `main_panel.preset`, `environment_exposure` → `main_panel.ppisp_exposure`, `environment_rotation_degrees` → `transform.rotation`.
- `status.iteration` lost its colon for the HUD; `training_panel.py:1132` shares the key and now reads "Iteration 5,000 (…)".
- `localized_count`, `_tr_format` and `installer._localized_progress` call `str.format` on translator output unguarded — a plugin `loc_set` override with a stray brace raises, where the C++ side now degrades gracefully.
- `common.filter` drops the old "…"; `download_progress_known` formatted with `speed=""` leaves a trailing space.
- Unrelated churn: `rml_panel_host.cpp` drops 4 blank lines; `image_preview.rml:125` and `rendering.rml:518-519` picked up odd indentation.
- `gui_manager.cpp:155,273` localizes two LOD strings and leaves their siblings ("Waiting for frame", "Tree loaded, off", "leaf-only reference") English.

Items 1–4 are the merge blockers; the rest can follow. Thanks again for keeping at this — the locale data itself is in great shape.

francescofugazzi:
I went through all the listed points and hopefully addressed the remaining issues in the latest local commits.

The implemented fixes include:

- Export outcomes are now propagated consistently through the native task state, Python overlay state, and both Python-facing bridges.
- Dataset loading explicitly resets and updates the stable `outcome` field instead of relying on localized task stage text.
- The missing common editing commands were added to every shipped locale:
  - `common.cut`
  - `common.copy`
  - `common.copy_all`
  - `common.paste`
  - `common.select_all`
- Localization contracts now detect direct `LocalizationManager::get(...)` calls.
- Initial RML localization refreshes now run correctly even when the initial language generation is zero.
- RML refreshes preserve application-owned dynamic content instead of overwriting it with stale localized text.
- Translated RML `<option>` elements are refreshed recursively, including non-DOM children, and the cached selected label is updated.
- Attribute and text translation directives are processed through a single regex traversal.
- The lifetime behavior of `LocalizationManager::get()` is documented explicitly.
- Operator labels and operator cancellation/error messages now resolve through localization keys.
- COLMAP overwrite warnings now report the actual sparse format:
  - `cameras.bin`, `images.bin`, and `points3D.bin` for binary sources;
  - `cameras.txt`, `images.txt`, and `points3D.txt` for text sources.
- Python localized formatting now supports both positional and named placeholders while failing safely on malformed translator-controlled strings.
- The CI hardcoded-UI check now runs with `--fail-on-candidates`.
- The unused localization key constant was removed.
- The iteration label was separated from the value-only status key to avoid punctuation conflicts across languages.
- The remaining LOD and common UI labels were added to all shipped locales.
- The Rendering Panel environment map, exposure, and rotation labels are now explicitly defined in every locale.
- The localization contract scanner now inspects literal Python `LOCALE_KEY` dictionaries, preventing indirect mappings from bypassing key validation.
- Unrelated whitespace churn was removed from the affected RML and C++ files.

The English locale currently contains 2042 localization keys. All shipped locales are checked against the English key set, placeholder signatures, plural forms, and one-key-per-line formatting.

Validation completed locally:

- localization contract tests;
- locale completeness checks;
- hardcoded UI audit;
- indirect Python localization mapping checks;
- one-key-per-line locale formatting check;
- placeholder and plural-form validation;
- Python syntax checks;
- clang-format validation;
- git diff validation.

PR #1528, opened 2026-08-01
ci: pin CMake to 4.4.0 (fixes the real cause of the Ubuntu break)

MrNeRF:
Supersedes #1526, which fixed a symptom. #1526 is already merged, so this PR also removes what it added.

## Symptom

Ubuntu CI went red on `master` with no relevant code change:

| Run | Commit | Date | Result |
|---|---|---|---|
| Ubuntu CUDA Build | `32ded4bd8` | 2026-07-29 | success |
| Ubuntu CUDA Build | `eb3b4ca04` | 2026-07-30 | failure |
| Ubuntu CUDA Build | `16047487a` | 2026-07-31 | failure |

The job dies in the **Configure** step with a misleading headline error:

```
CMake Error: CMake was unable to find a build program corresponding to "Ninja".
             CMAKE_MAKE_PROGRAM is not set.
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
```

That is fallout, not the cause. Further up the log:

```
error: building sdl3:x64-linux failed with: BUILD_FAILED

SDL_gtk.c:90:19: error: static declaration of 'getresuid' follows non-static declaration
SDL_gtk.c:98:19: error: static declaration of 'getresgid' follows non-static declaration
```

vcpkg failed to build `sdl3`, so configure never obtained a compiler.

## Root cause

`.github/workflows/ubuntu.yml` installs CMake from Kitware's apt repo **without a version**:

```yaml
echo 'deb [...] https://apt.kitware.com/ubuntu/ noble main' > /etc/apt/sources.list.d/kitware.list
apt-get install -y git curl unzip cmake gcc-14 g++-14 ...
```

Kitware's `noble` index currently offers:

```
4.3.4-0kitware3ubuntu24.04.1
4.4.0-0kitware1ubuntu24.04.1
4.4.1-0kitware1ubuntu24.04.1   <- newest, installed automatically
```

**CMake 4.4.1 regressed `check_symbol_exists`**: it no longer honours `CMAKE_REQUIRED_FLAGS`,
so the probe compile loses `-D_GNU_SOURCE=1`, and with `-std=c17` (rather than `gnu17`) the
compiler cannot see glibc's prototypes. Every feature check that depends on those flags now
reports "not found".

Diagnosis credit goes to the SDL maintainers in libsdl-org/SDL#16072:

> As far as I can see, `-D_GNU_SOURCE=1` is missing in cmake 4.4.1 case, and with `-std=c17`
> on the command line -- not `-std=gnu17` -- compiler extensions are not enabled and compiler
> isn't seeing the prototype. — sezero

> I can reproduce by downloading a CMake 4.4.1 release. I think `check_symbol_exists` does not
> honor `CMAKE_REQUIRED_FLAGS`. The compile command does not have `-D_GNU_SOURCE=1` in 4.4.1.
> — madebr

Reported to Kitware: https://gitlab.kitware.com/cmake/cmake/-/work_items/28001
Also hit by freedesktop-sdk: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/33486

**Why it surfaced exactly on 2026-07-30 with no code change:** vcpkg includes tool versions in
its package ABI hash. A new CMake invalidates the binary cache, so every port rebuilds. sdl3 had
been restored from cache for months; once it was actually rebuilt, the broken detection showed
up. With `HAVE_GETRESUID` undetected, `SDL_gtk.c` compiles a `static inline` fallback under
libc's own name, which collides with glibc's non-static declaration in `<unistd.h>`.

## Why #1526 was the wrong fix

#1526 added a vcpkg overlay port patching sdl3 so its fallbacks no longer collide. That makes
Ubuntu green, and it is what this repo already does for genuine upstream defects
(`libplacebo`, `libsystemd`). But it treats one visible instance of a general failure.

Detection breakage under 4.4.1 is **not limited to `getresuid`/`getresgid`**. The same SDL issue
records `fdatasync` and `gethostname` also failing detection. Those do not fail to compile —
they silently disable functionality. Patching sdl3 therefore converted one loud failure into a
green build in which an unknown number of ports may be misconfigured, across every dependency
that uses `check_symbol_exists`. A green-but-silently-wrong toolchain is worse than red CI.

sdl3 was the messenger, not the problem.

## This PR

1. Pins `cmake` and `cmake-data` to `4.4.0-0kitware1ubuntu24.04.1`, the newest release before
   the regression, with a comment explaining why and linking the upstream report.
2. Deletes `.github/overlays/sdl3/` (added by #1526). No port carries a local patch again.

Merging this changes the CMake version, which changes vcpkg's ABI hash, which forces a full
rebuild of the dependency set. **That is intended**: packages built by CI since 2026-07-30 were
configured by CMake 4.4.1 and may have features silently disabled.

## Verification

- Ubuntu CUDA Build on this PR is the test — it must go green *without* any sdl3 patch present.
- Windows is unaffected and needs no pin: `.github/workflows/windows.yml` has no CMake install
  step and uses the Visual Studio toolchain's CMake via `vcvars64.bat`. Windows CUDA Build was
  green on `master` throughout the Ubuntu outage.

## Removing the pin

Drop it once a fixed CMake is published (i.e. a release after 4.4.1 in which Kitware work item
28001 is resolved). Check with:

```bash
apt-cache madison cmake      # available Kitware versions
cmake --version              # what is actually installed
```

Then delete the two pinned versions from the `apt-get install` line and let CI prove the newer
release configures correctly.

## Upstream

libsdl-org/SDL#16092 proposed renaming SDL's fallbacks so they can never collide with a libc
declaration regardless of detection (and fixes the fallback `getresgid()` taking `uid_t *`
instead of `gid_t *`). That is hardening on its own merits, not a fix for this break; the SDL
maintainers can take it or close it. Nothing here depends on it.


miriameng:
The cmake regression looks like it's fixed in the latest 4.4.2, so no more pinning to 4.4.0 should be required

MrNeRF:
Pinning it will prevent such issues. Not sure if the latest update is required. Rather not I think

PR #1535, opened 2026-08-02
Fix PPISP Checkpoint Resume Color Regularizer State

### LINKED ISSUE #1469: Resume checkpoint with PPISP does not work
Resuming from a checkpoint with PPISP and a disabled camera does not work.

Repro:
* start LFS
* load dataset
* enable PPISP
* start training
* pause training after 1000 steps
* save checkpoint
* exit LFS 
* --> error visible in logs (not critical)
* start LFS
* load checkpoint
* resume training
* --> fails with error 

Image

Log files:

[lfs-resume-checkpoint.log](https://github.com/user-attachments/files/30459563/lfs-resume-checkpoint.log)
[lfs-save-checkpoint.log](https://github.com/user-attachments/files/30459562/lfs-save-checkpoint.log)

kmeirlaen:
fixes #1469

## Summary

This patch fixes a crash when resuming a training checkpoint created with PPISP enabled. The non-serialized ZCA pseudo-inverse matrix used by PPISP color-mean regularization is now rebuilt on every construction/restore path, and checkpoint adoption no longer swaps that constant as if it were checkpoint state.

## Observed Symptoms

Resuming a `--ppisp` checkpoint loads successfully, then fails on the first optimizer commit step. The failure report shows:

```text
Failed expression: is_valid()
Context: device transfer requires a valid tensor
Tensor::to(Device, CUstream_st*) <- Tensor::cpu()
Training step failed: iteration=4726, step_phase=OptimizerCommit
```

Replication for the original issue:

1. Load a checkpoint trained with `--ppisp`.
2. Resume training.
3. Observe that the first step after resume enters PPISP regularization and crashes while reading `color_pinv_block_diag_.cpu()`.

## Root Cause

`PPISP::color_pinv_block_diag_` is a constant CUDA tensor used by the color-mean regularizer. It is initialized during `finalize()` via `allocate_tensors()`, but it is intentionally not serialized.

The checkpoint load path constructs a temporary `PPISP`, calls `deserialize()`, and then calls `adopt_checkpoint_state()` on the live finalized instance. Before this patch:

- `deserialize()` rebuilt gradients and scratch buffers, then set `finalized_ = true`, but never rebuilt `color_pinv_block_diag_`.
- `adopt_checkpoint_state()` swapped `color_pinv_block_diag_`, moving the live valid constant into the temporary object and moving the invalid default tensor into the active PPISP.
- `reg_loss_gpu()` and `reg_backward()` then attempted `color_pinv_block_diag_.cpu()` and hit the tensor validity assertion.

## What This Patch Fixes

- Adds `PPISP::init_color_pinv_block_diag()` to centralize the constant tensor initialization.
- Calls the helper from `allocate_tensors()` for fresh runs.
- Calls the helper from `deserialize()` before marking the object finalized.
- Calls the helper from `deserialize_inference()` before marking the object finalized.
- Removes the `color_pinv_block_diag_` swap from `adopt_checkpoint_state()` because the tensor is derived constant state, not checkpoint state.
- Adds `PPISPCheckpointRoundtripTest`, which serializes, deserializes, adopts checkpoint state into a live PPISP, and exercises both `reg_loss_gpu()` return behavior and `reg_backward()` side effects.

### DIFF
--- a/src/training/components/ppisp.cpp
+++ b/src/training/components/ppisp.cpp
@@ -148,6 +148,10 @@ namespace lfs::training {
                                             color_params_.ptr<float>(), crf_params_.ptr<float>(), num_cameras_,
                                             num_frames_, nullptr);
 
+        init_color_pinv_block_diag();
+    }
+
+    void PPISP::init_color_pinv_block_diag() {
         // ZCA pinv block-diagonal matrix for color mean regularization
         // 8x8 block-diagonal: [Blue 2x2, Red 2x2, Green 2x2, Neutral 2x2]
         // From Python: _COLOR_PINV_BLOCK_DIAG
@@ -1169,6 +1173,7 @@ namespace lfs::training {
         ctrl_bwd_rgb_h_ = 0;
         ctrl_bwd_rgb_w_ = 0;
 
+        init_color_pinv_block_diag();
         finalized_ = true;
     }
 
@@ -1208,7 +1213,6 @@ namespace lfs::training {
         uid_to_frame_idx_.swap(loaded.uid_to_frame_idx_);
         uid_to_camera_id_.swap(loaded.uid_to_camera_id_);
         std::swap(finalized_, loaded.finalized_);
-        std::swap(color_pinv_block_diag_, loaded.color_pinv_block_diag_);
     }
 
     void PPISP::serialize_inference(std::ostream& os) const {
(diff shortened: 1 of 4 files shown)


MrNeRF:
built, checked, and reapplied tests on top of master. All good. Thank you!

PR #1542, opened 2026-08-03
fix: pre-flight GPU check — refuse cards below SM 7.5 with clear message

### LINKED ISSUE #1540: Pre-flight GPU check: refuse cards below SM 7.5 with a clear message instead of crashing
## Motivation

A user on Discord tried the free Windows build on a **GTX 1070** and got a silent crash — no window, no explanation, just trace logs ending after CUDA init. A helper had to diagnose it by hand ("you will need an RTX card, 2xxx or higher") and the user reasonably asked for the requirement to be documented.

## Current behavior

- The portable/release builds are PTX-only with `BUILD_CUDA_MIN_SM` defaulting to **75** (`CMakeLists.txt:330`), with a hard floor of 70 (`CMakeLists.txt:393`). Kernels can never run on Pascal (SM 6.1).
- At startup, `application.cpp` already queries the device: it logs `GPU: ... (SM x.y, ... MB)` around `src/app/application.cpp:247-248` — but never compares SM against the build's minimum. Initialization continues and the app dies later with no user-facing message.
- The CUDA driver version gets a `LOG_WARN` (`application.cpp:231`, requires 12.8+ / driver 570+) but startup also continues.

## Proposed fix

1. Bake the build's minimum SM into the binary (e.g. a compile definition set from `BUILD_CUDA_MIN_SM`), and right after `cudaGetDeviceProperties` succeed/fail fast: if `prop.major*10 + prop.minor < MIN_SM`, show a plain-language error (message box on Windows, stderr otherwise):
   > This PC's GPU (NVIDIA GeForce GTX 1070) doesn't meet the minimum requirement — an NVIDIA RTX 20-series card or newer is needed.
2. Same treatment for the driver check: below driver 570, tell the user to update instead of warning into a crash.
3. Add a **Requirements** section to the README: NVIDIA RTX 20-series or newer (compute capability 7.5+), NVIDIA driver 570+, Windows; GTX 10-series and older, AMD, and Intel are not supported.

The download portal now shows these requirements before purchase/download, but the binary itself should say why it won't run — most affected users never read a requirements table.

Refs: the memory-arena trace from the report ends at `memory_arena.cu:643` after `Committed 512 MB`; GPU line logged: `GPU: NVIDIA GeForce GTX 1070 (SM 6.1, 8191 MB)`.

jacobvanbeets:
Fixes #1540

## Problem

Users on GTX 10-series (Pascal, SM 6.1) get a silent crash at startup - no window, no explanation. The app logs GPU: NVIDIA GeForce GTX 1070 (SM 6.1, 8191 MB) but never compares against the build's minimum SM 7.5, then dies later with no user-facing message.

### CMakeLists.txt
- Bake BUILD_CUDA_MIN_SM into binary via add_compile_definitions(LFS_MIN_SM=...) so runtime can fail-fast on unsupported GPUs

### src/app/application.cpp
- After cudaGetDeviceProperties, compute device_sm = prop.major * 10 + prop.minor; if device_sm < LFS_MIN_SM, show a MessageBoxW dialog naming the GPU and exit immediately
- Replace LOG_WARN for old driver with blocking MessageBoxW dialog requiring driver 570+

### README.md
- Add Requirements section listing RTX 20-series+, driver 570+, Windows as minimums; explicitly note GTX 10-series/AMD/Intel are unsupported

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -329,9 +329,6 @@ endif()
 # Minimum SM for PTX-only (75=Turing, 80=Ampere, 86=GA102, 89=Ada, 90=Hopper)
 set(BUILD_CUDA_MIN_SM "75" CACHE STRING "Minimum SM for PTX-only builds")
 
-# Bake the minimum SM into the binary so runtime can fail-fast on unsupported GPUs.
-add_compile_definitions(LFS_MIN_SM=${BUILD_CUDA_MIN_SM})
-
 set(CMAKE_CXX_STANDARD 23)
 set(CMAKE_CUDA_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
@@ -398,6 +395,7 @@ if(BUILD_CUDA_PTX_ONLY)
     endif()
 
     set(LichtFeld-Studio_CUDA_ARCH "${BUILD_CUDA_MIN_SM}-virtual")
+    set(LFS_RUNTIME_MIN_SM "${BUILD_CUDA_MIN_SM}")
 
     math(EXPR _maj "${BUILD_CUDA_MIN_SM}/10")
     math(EXPR _min "${BUILD_CUDA_MIN_SM}%10")
@@ -407,8 +405,16 @@ if(BUILD_CUDA_PTX_ONLY)
 else()
     # Native GPU only
     set(TORCH_CUDA_ARCH_LIST "${DETECTED_COMPUTE_CAP}" CACHE STRING "" FORCE)
+    string(REPLACE "." "" LFS_RUNTIME_MIN_SM "${DETECTED_COMPUTE_CAP}")
     message(STATUS "CUDA: native (${LichtFeld-Studio_CUDA_ARCH})")
 endif()
+
+if(NOT LFS_RUNTIME_MIN_SM MATCHES "^[0-9]+$")
+    message(WARNING "Could not derive a runtime SM floor from '${LFS_RUNTIME_MIN_SM}'; "
+        "falling back to 70. The startup GPU check will be permissive.")
+    set(LFS_RUNTIME_MIN_SM "70")
+endif()
+message(STATUS "Runtime GPU floor: SM ${LFS_RUNTIME_MIN_SM}")
 
 enable_language(CUDA)
 find_package(CUDAToolkit 12.8 REQUIRED)
@@ -644,6 +650,10 @@ add_executable(${PROJECT_NAME}
     src/app/converter.cpp
     resources/lichtfeld-studio.rc
 )
+
+# The binary only carries code for these architectures; the app refuses to start on anything
+# lower instead of dying inside the first kernel launch (#1540).
+target_compile_definitions(${PROJECT_NAME} PRIVATE LFS_MIN_SM=${LFS_RUNTIME_MIN_SM})
 
 foreach(_lfs_git_version_consumer IN ITEMS
         ${PROJECT_NAME}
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thanks @jacobvanbeets — the diagnosis and the placement were right, and the gate does fire exactly where the GTX 1070 report died (before `warmup_kernels` JIT). I've pushed a maintainer commit on top rather than sending you round another review cycle. Summary of what changed and why:

**The driver dialog didn't actually stop anything.** `checkCudaDriverVersion()` returns `bool` and all five call sites discarded it, so the new modal was followed by the same crash — item 2 of #1540 was still open. The driver and SM checks are now one `preflightGpu()` called once from `Application::run`, so headless, TCP and `--render-path` runs are covered too, not just the GUI.

**`std::exit(1)` isn't safe here.** It runs static destructors and never flushes diagnostics, so the `LOG_ERROR` could be lost. `crash_handler.hpp` documents `teardown_gpu_before_exit()` + `flush_and_exit(1)` as the sanctioned sequence for exactly this; that idiom was already in the same file ~100 lines below.

**`BUILD_CUDA_MIN_SM` is a PTX-only value.** `BUILD_CUDA_PTX_ONLY` defaults OFF, where the arch comes from `nvidia-smi` and that cache var is inert. Baking it unconditionally cut both ways: a native `sm_70` build would refuse the machine it was built on, and a native `sm_86` build on an RTX 2080 would pass `75 >= 75` and then die as before. `LFS_MIN_SM` is now derived from the architectures actually embedded and scoped to the app target, with the hardcoded 75 fallback replaced by an `#error`.

Also folded in: modal suppressed for `--headless`/`--render-path` (a `MessageBoxW` in a non-interactive run blocks it forever), log before the dialog rather than after, `utf8_to_wstring` instead of the byte-wise widen, a failing `cudaGetDeviceProperties` is now fatal instead of silently skipping the gate, dead `warmupCudaSync` removed, and `clang-format`.

One correction worth flagging: **Turing isn't RTX-only.** The GTX 16-series and Tesla T4 are also SM 7.5, so "an RTX 20-series card is needed" would have told a GTX 1660 owner their supported card wasn't. Both the message and the README now say "GTX 16-series, RTX 20-series and up". The README requirements also moved into the existing "Current project notes" list — the standalone section listed Windows as a hard requirement, which contradicts the supported Linux source build.

**Verified:** build + `clang-format` clean; 200-iteration headless train passes with the gate logging once; GUI starts and loads a PLY with no error noise; and the reject path was exercised for real by temporarily forcing the floor above this card — correct message, teardown ran, exit code 1, no dialog in headless.

**Deliberate behaviour change:** driver < 570 is now a hard stop instead of a warning, per item 2 of the issue. Note that CUDA minor-version compatibility means a *native SASS* build can technically run on an older 12.x driver; PTX-only builds (what we ship) genuinely cannot. Since configure already `FATAL_ERROR`s below 570, refusing at runtime is consistent — easy to soften to PTX-only if you'd rather.

**Pre-existing gaps this does not close** (none are regressions, all worth their own issues):
- `external/CMakeLists.txt:106` sets nvImageCodec's `CUDA_TARGET_ARCHS` from `DETECTED_COMPUTE_CAP` regardless of `BUILD_CUDA_PTX_ONLY`, so a portable build can ship a codec compiled for a different arch than the app advertises.
- `PreprocessMode` never reaches `Application::run` but can launch project kernels via `computeRawDepthAnchors` (`preprocess.cpp:1006`), so it stays ungated.
- `main.cpp:53` uses `setenv(..., overwrite=0)`, so a pre-set `CUDA_MODULE_LOADING=EAGER` is not overridden; under EAGER the module load can precede the gate.
- `GuiManager::checkCudaVersionAndNotify` is now unreachable for the app; left in place rather than pulling the visualizer into this fix.

PR #1550, opened 2026-08-03
Remove misleading video frame orientation helper

### LINKED ISSUE #1546: `orientVideoExportFrameForEncoder` promises orientation handling and does nothing
**What's wrong**

The function is named as though it fixes frame orientation before encoding. Its entire
body is a validity check and `return image.contiguous();`.

This actively misleads. While debugging an upside-down exported video, this was the first
place to look for the flip — and it was a dead end; the real bug was elsewhere.

**Where**

`src/visualizer/gui/async_task_manager.cpp:336-341`

**How to fix**

Delete it and call `.contiguous()` at the call site, or give it a name describing what it
does.

noor-ahmadi:
Closes #1546.

`orientVideoExportFrameForEncoder` no longer performed any orientation work—it only made the tensor contiguous. This removes the misleading helper and keeps that behavior at the export call site.

Checked with `clang-format --dry-run --Werror src/visualizer/gui/async_task_manager.cpp` and `git diff --check`.

### DIFF
--- a/src/visualizer/gui/async_task_manager.cpp
+++ b/src/visualizer/gui/async_task_manager.cpp
@@ -332,12 +332,6 @@ namespace lfs::vis::gui {
         return state.cached_environment_resolved_path;
     }
 
-    lfs::core::Tensor orientVideoExportFrameForEncoder(const lfs::core::Tensor& image) {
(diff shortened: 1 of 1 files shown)


MrNeRF:
Good catch! Thank you!

PR #1553, opened 2026-08-04
#1475 Fix drifted parameter defaults in the property registry

### LINKED ISSUE #1475: "Reset to default" gives the wrong value for three training parameters
**What's wrong**

There are two places that say what a parameter's default is, and they disagree. The struct has one
value; the property registry has another. When a user clicks "reset", the code reads the *registry*
value — so the user gets a number that was never the real default.

Three parameters are affected, and two of them are off by 2×:

| Parameter | Real default (struct) | What reset gives | Effect |
|---|---|---|---|
| `opacity_lr` | `0.025` | `0.05` | 2× learning rate |
| `init_rho` | `0.0005` | `0.001` | 2× |
| `use_ppisp` | `false` | `true` | feature silently turns on |

A doubled learning rate does not crash. It just trains differently, and nobody would suspect the
reset button.

**Where**

- Struct defaults: `src/core/include/core/parameters.hpp:94` (`opacity_lr = 0.025f`), `:163` (`use_ppisp = false`), `:206` (`init_rho = 0.0005f`)
- Registry defaults: `src/python/lfs/py_params.cpp:63` (`0.05f`), `:308` (`0.001f`)
- The reset path that reads the registry: `src/python/lfs/py_params.cpp:667-701` — `PyOptimizationParams::reset` calls `PropertyRegistry::instance().get_property(...)` then uses `meta->default_value`

**How to fix**

1. Correct the three registry defaults to match the struct.
2. Add a gtest that walks **every** property in the `optimization` group and asserts its registry
   default equals the matching member of a default-constructed `OptimizationParameters`.

The test is the real fix — it prevents the next one. There are 96 parameters and nothing currently
stops them drifting again.

**How to check it worked**

The new test fails before the change and passes after.

dusancubik:
Fixes #1475

## Change
-  Fix three wrong defaults in `register_optimization_properties()`
  (`src/python/lfs/py_params.cpp`) so they match `OptimizationParameters`
  in `src/core/include/core/parameters.hpp`:

    | Parameter | Was | Now |
    |---|---|---|
    | `opacity_lr` | `0.05f` | `0.025f` |
    | `init_rho` | `0.001f` | `0.0005f` |
    | `use_ppisp` (id `"ppisp"`) | `true` | `false` |
- Add `tests/test_training_parameters.cpp`, which walks **every** property in the
  `optimization` group and asserts its registry default equals the corresponding
  member of a default-constructed `OptimizationParameters`.

**Note:** `OptimizationParameters` has 96 members, but only **82** are registered in the `optimization` property group — those are the ones the test covers. The other 14 have no registry entry:
`auto_train`, `bg_color`, `bg_image_path`, `config_file`, `debug_python`, `debug_python_port`, `enable_save_eval_images`, `eval_steps`, `no_splash`, `ppisp_lr`, `ppisp_reg_weight`, `ppisp_sidecar_path`, `ppisp_warmup_steps`, `save_steps`

## Verification
**Before the fix**
defaults_test_fail

**After the fix:**
defaults_test_pass

### DIFF
--- a/src/python/lfs/py_params.cpp
+++ b/src/python/lfs/py_params.cpp
@@ -60,7 +60,7 @@ namespace lfs::python {
                         "Learning rate for spherical harmonics")
             .flags(PROP_LIVE_UPDATE)
             .float_prop(&OptimizationParameters::opacity_lr,
-                        "opacity_lr", "Opacity LR", 0.05f, 0.0f, 1.0f,
+                        "opacity_lr", "Opacity LR", 0.025f, 0.0f, 1.0f,
                         "Learning rate for opacity")
             .flags(PROP_LIVE_UPDATE)
             .float_prop(&OptimizationParameters::scaling_lr,
@@ -254,7 +254,7 @@ namespace lfs::python {
                        "mip_filter", "Mip Filter", false,
                        "Enable mip filtering (anti-aliasing)")
             .bool_prop(&OptimizationParameters::use_ppisp,
-                       "ppisp", "PPISP", true,
+                       "ppisp", "PPISP", false,
                        "Per-pixel image signal processing")
             .bool_prop(&OptimizationParameters::ppisp_use_controller,
                        "ppisp_use_controller", "Controller", false,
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thank you for contributing! Lgtm!

PR #1557, opened 2026-08-05
ci: run CPU test suite on Linux

MrNeRF:


MrNeRF:
Superseded: master's Ubuntu workflow already builds and runs the CPU-only format tests on every PR, and the 'cpu' CTest label this change relied on no longer exists (labels are now fast/format/gpu).

PR #1560, opened 2026-08-05
Splat simplify target incorrect

### LINKED ISSUE #1470: Simplify not working as expected
latest master:
* load splat (1M splats)
* use splat fimplify too 500000
--> results in splat of 107 splats 

different splat:
* load splat (5M splats)
* simplify to 2.5M -> OK
* simplify to 1.250.000 -> OK
* simplify to 625K -> NOK -> only 96 (no 96M or 96K, just 96)

Also reported in discord channel:

Image

kmeirlaen:
fixes #1470: 

## Summary

This patch prevents splat simplification from collapsing dense voxel groups far below the requested target count. The fix keeps the public simplify API unchanged and constrains the voxel merge pass so it can reach the requested per-pass target without overshooting below it.

## Original Symptoms

Issue #1470 reported that simplifying a dense splat could produce dramatically fewer Gaussians than requested. The local reproduction asset `data/kerstbol-isolated-rotated_137502.ply` contains 137,502 splats. With `ratio = 0.5` and opacity pruning disabled, the target is 68,751 splats, but the unchanged implementation returned 5,218 splats.

## Reproduction Steps

1. Load `data/kerstbol-isolated-rotated_137502.ply`.
2. Run the deployed `lfs::core::simplify_splats` path with `SplatSimplifyOptions::ratio = 0.5` and `opacity_prune_threshold = 0.0f`.
3. Compare the returned splat count with `ceil(input_count * ratio)`.

The regression test `SplatSimplify.RealPlySimplifyDoesNotUndershootRequestedTarget` captures this using the real PLY asset rather than generated data.

## Root Cause

The simplifier estimated a global voxel size for the desired pass target, grouped all splats into voxels, and merged every multi-splat voxel group in the pass. Dense or unevenly distributed splats could create very large voxel groups. Merging all of those groups replaced each group with one splat, so a single pass could remove far more splats than the remaining target allowed.

## What This Patch Fixes

- Oversized voxel groups are recursively split by their longest spatial axis before merging.
- Final merge groups are capped deterministically.
- Each pass trims the selected merge groups so the pass cannot reduce the output below its target.
- If a selected group would remove too many rows, it is partially split into one smaller merge group plus singletons, preserving the exact remaining savings needed.
- Existing merge behavior, progress reporting, and history tracking continue to use the same deployed core simplifier path.

## Return Contract and Side Effects

The changed production method still returns `std::expected<std::unique_ptr<SplatData>, std::string>` through `simplify_splats` and keeps the input `SplatData` immutable. The new regression asserts both the returned output count and that the source splat count remains unchanged after the call.

`simplify_splats_with_history` continues to return the simplified splat and merge tree through the same contract. Existing history tests passed after the change.

## Test Results

Pre-fix regression command:

```powershell
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && cd /d build-tests-release\tests && set PATH=..;.;%PATH% && lichtfeld_tests.exe --gtest_filter=SplatSimplify.RealPlySimplifyDoesNotUndershootRequestedTarget --gtest_color=no'
```

Pre-fix result: failed as expected. Returned 5,218 splats; expected 68,751.

Post-fix regression command:

```powershell
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && cd /d build-tests-release\tests && set PATH=..;.;%PATH% && lichtfeld_tests.exe --gtest_filter=SplatSimplify.RealPlySimplifyDoesNotUndershootRequestedTarget --gtest_color=no'
```

Post-fix result: passed, 1 test.

Simplify suite command:

```powershell
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && cd /d build-tests-release\tests && set PATH=..;.;%PATH% && lichtfeld_tests.exe --gtest_filter=SplatSimplify.* --gtest_color=no'
```

Result: passed, 10 tests.

Additional focused check:

```powershell
& cmd.exe /c 'call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64 && cd /d build-tests-release\tests && set PATH=..;.;%PATH% && lichtfeld_tests.exe --gtest_filter=TrainingParametersTest.* --gtest_color=no'
```

Result: 15 passed, 1 failed. The failure is unrelated to this patch: `TrainingParametersTest.OldNewToJsonParity` compares generated LF JSON against CRLF fixture bytes.


MrNeRF (approved):
Ask for 50%, you get exactly 50% now. Before you got 4% on dense splats. Verified on the real 137k asset, the new test passes in 165 ms. I pushed a master sync and a timeout on the asset download. Good to merge.

Issue #1532, opened 2026-08-01
OPF import (Open Photogrammetry Format) as a training dataset source


MrNeRF:
OPF (Open Photogrammetry Format, Pix4D, spec 1.0.5: https://pix4d.github.io/opf-spec/) is an
open JSON-based container for complete photogrammetry projects: a `.opf` project file
referencing camera lists (with per-camera UIDs), calibrated camera parameters (intrinsics,
distortion, poses), and point clouds. PIX4Dmatic exports it natively, and it is designed
precisely for cross-vendor exchange of calibrated reconstructions — the same role COLMAP
output plays for us today.

## Why

Users coming from Pix4D pipelines currently must re-run SfM or convert through COLMAP to train
in LichtFeld. OPF import makes a Pix4D reconstruction directly trainable: calibrated cameras +
sparse points in, training starts — no third-party conversion, no re-solving.

## Scope — import as a dataset/SfM source (peer of the COLMAP and transforms.json loaders)

- [ ] Parse the `.opf` project container and resolve its referenced components (relative
      resolution per spec; JSON schemas are published for validation).
- [ ] Cameras: input camera list + calibrated cameras → our camera model (intrinsics,
      distortion model mapping, poses). Unsupported distortion models fail with a typed error
      naming the model — no silent approximation.
- [ ] Point cloud: calibrated/densified cloud → training initialization points (sparse is
      sufficient; dense optional).
- [ ] Image resolution: OPF references images by URI — reuse the existing recursive
      case-insensitive image index (`filesystem_utils`), including the masks/depth sibling
      conventions where present.
- [ ] Dataset detection: `_looks_like_dataset`/scanner and the import panel recognize an OPF
      project directory as one dataset asset.
- [ ] GPS/geo metadata: OPF carries georeferencing — map it onto the PROJ georeference block
      (CRS + f64 world_origin) introduced by the .licht format work, instead of dropping it.
- [ ] Fixtures: at least one real PIX4Dmatic export (Pix4D publishes samples) + schema-invalid
      and missing-component cases with typed errors.
- [ ] Per-vendor extensions in OPF (the spec is extensible): unknown extensions are preserved
      or ignored with a log line, never a parse failure.

## Non-goals for the first cut

OPF *export* (LichtFeld is a consumer of reconstructions here; our exchange output remains
COLMAP), and OPF's mesh/texture components.

PR #1572, opened 2026-08-06
Add initial OPF dataset loading and training integration

francescofugazzi:
## Description

Related to [issue #1532](https://github.com/MrNeRF/LichtFeld-Studio/issues/1532).

This PR starts the OPF integration in LichtFeld-Studio.

The current implementation adds:

- OPF project discovery when selecting a dataset folder.
- OPF resource and camera parsing.
- Camera and coordinate-system conversion to LichtFeld conventions.
- Sparse point-cloud glTF loading.
- Scene-graph creation for OPF cameras and point clouds.
- Training initialization from imported point clouds.
- Localized OPF GUI messages.
- PLY importing as a point-cloud scene node from the Training panel.
- Initial support for projects containing multiple point-cloud resources.
- Regression tests for parsing, validation, camera conversion and training initialization.

The implementation is functional but not complete. It still requires additional testing with different OPF projects, camera models, coordinate systems, point-cloud layouts and depth/confidence resources.

Manual validation was performed with:

- [Olympic Flame dataset]
- [Catch Rainwater / PIX4Dcatch dataset]
- [OPF example datasets]
- [Rivaz dataset]

The Catch Rainwater project contains RGB images together with depth and confidence resources. The RGB images are correctly used as the camera dataset, while the additional resources remain separate project data.

A complete training run was successfully performed with the current implementation.

## Important status

This is an early implementation and should be considered a prototype.

Further work is still needed for:

- More extensive OPF dataset testing.
- Additional camera and reference-frame validation.
- Final sparse/dense/depth point-cloud priority rules.
- Better handling of unsupported resources.
- Complete CLI validation.
- Complete localization and GUI validation.
- Clang/static-analysis validation.
- Cross-platform testing.
- Final documentation of the supported OPF subset.

## Note for @MrNeRF

The latest commit is the prototype corresponding to the private discussion about projects containing multiple point clouds.

The prototype currently:

- Loads supported point clouds as separate scene nodes.
- Selects one point cloud as the training initialization source.
- Shows the selected source with a green check mark in the Scene Graph.
- Synchronizes that selection with the Training panel dropdown.
- Allows changing the selected point cloud from either interface.
- Keeps the other point-cloud nodes available.
- Preserves the selected point cloud in the scene after training initialization.
- Hides the selected point cloud after it is converted into the training model.
- Allows importing an external PLY as a point-cloud node from the Training panel.

This part is intentionally not final. If the general direction is considered useful, the priority rules, scene representation and user interface can be refined in a follow-up iteration.

To test it, load an OPF project containing multiple point clouds, select different PCD nodes from the Scene Graph and Training panel, and verify that both controls remain synchronized and that training uses the selected source.

## Validation

The current tests cover OPF parsing, resource validation, camera conversion, point-cloud loading, folder-based discovery, scene integration, selected point-cloud training initialization and retention of scene nodes.

The implementation still requires broader testing before merge.


MrNeRF:
what's the state here? SHould probably also go first into dev

PR #1585, opened 2026-08-08
Fix newer GCC builds and remove inert Python UI APIs

### LINKED ISSUE #1450: 47 Python UI functions exist but silently do nothing
`src/python/lfs/rml_im_mode_layout.cpp` exposes **47 methods with empty bodies**. A plugin can call them, nothing happens, and there is no warning.

Some of them:

```
push_style_var_float, push_style_var_vec2, pop_style_var
push_style_color, pop_style_color
set_cursor_pos, set_cursor_pos_x
set_next_window_pos, set_next_window_size, set_next_window_focus,
set_next_window_bg_alpha, set_next_window_pos_centered, set_next_window_pos_center,
set_next_window_pos_viewport_center
set_keyboard_focus_here, capture_keyboard_from_app, capture_mouse_from_app
set_scroll_here_y, begin_group, end_group, tree_pop, set_next_item_width
```

`get_cursor_pos()` also always returns `(0, 0)`.

These are leftovers from the old ImGui-shaped API. Only 2 places call `warn_unsupported` (`menu` and `popover`); the other 45 fail silently.

**We already call some of them ourselves and get nothing back:**

- `src/python/lfs_plugins/overlays/__init__.py:332` and `:439` — `layout.set_next_window_pos(...)`
- `src/python/lfs_plugins/windows/layout_utils.py:11` — `layout.set_cursor_pos_x(layout.get_cursor_pos()[0] + indent)`

So overlay positioning and that indent helper are quietly doing nothing right now.

**What to do**

Pick one per method:

1. implement it,
2. make it log a warning once through `warn_unsupported`, or
3. remove it from the binding.

Silently doing nothing is the worst of the three, because plugin authors cannot tell the difference between a bug in their code and a stub in ours.

shadygm:
## Summary

- make the Python LSP client use `std::uint32_t` with an explicit `<cstdint>` include so it builds reliably with newer GCC versions
- remove unused and silently inert compatibility methods from `RmlUILayout` and duplicate tree aliases from `RmlSubLayout`
- preserve the separate, functional viewport-overlay `UILayout` drawing and window APIs
- remove the unused `lfs_plugins.windows.layout_utils` helper module
- update Python stubs, API documentation, migration guidance, and regression coverage

## Root cause

The Python LSP client relied on an unqualified fixed-width integer type being available through transitive includes, which newer GCC toolchains do not guarantee.

Separately, `RmlUILayout` exposed an ImGui-shaped compatibility surface whose retained-mode implementation consisted of no-op bodies or inert return values. No in-tree panel code used those methods, while similarly named overlay operations belong to the distinct and functional `UILayout` API. Keeping both surfaces made unsupported calls appear successful.

## Impact

Unsupported RML panel methods now fail visibly through normal missing-attribute behavior instead of silently doing nothing. Existing viewport overlays retain their working `UILayout` drawing and window methods. The generated stubs and documentation now describe the actual API.

## Validation

- `cmake --build cmake-build-release -j6`
- `cmake --build cmake-build-release --target lfs_py -j6`
- `cmake --build cmake-build-release --target check_python_stubs -j6`
- direct Python API contract check: 63 removed RML methods absent; working overlay methods retained
- `git diff --check`

The bundled release-tree Python does not include pytest, so the new pytest cases were not invoked directly.

Fixes #1450


MrNeRF:
lgtm

PR #1503, opened 2026-07-30
[Superseded — reference only] Add persistent preferences, recovery controls, MCP management, optional scene upscaling, and video export controls

francescofugazzi:
> [!IMPORTANT]
> This PR has been superseded by a sequence of smaller, independently reviewable PRs.
>
> Replacement PRs:
>
> - #1629
> - #1658
> - #1672
> - #1725 
> - #1895 
> - #1939 
>
> This original PR remains open temporarily as a historical and implementation reference. Please review the smaller replacement PRs instead.

## Summary

This PR establishes a single, persistent home for user-level application preferences, adds crash-safe recovery and safe-mode controls, improves startup and confirmation behavior, introduces configurable MCP runtime management, and provides a backend-neutral scene-upscaling and temporal-rendering architecture.

It now also exposes the renderer's 16-bit/32-bit splat quantization and extends the same Native, Spatial, Temporal, NVIDIA DLSS, and AMD FSR 3.1 rendering paths to Sequencer video export.

Native remains the normal LichtFeld Studio renderer and the safe default. Optional vendor SDKs remain explicit developer opt-ins and are never downloaded, bundled, or required by a standard build.

## User-owned storage

- Centralize application-owned paths through `UserPaths`.
- Use `%USERPROFILE%\.lichtfeld` on Windows and XDG-compatible locations on Linux.
- Honor compiled portable-storage defaults and the explicit `LFS_HOME` override.
- Remove current-working-directory fallbacks for GUI-owned files.
- Keep global application preferences separate from `.licht` project state.
- Do not inspect, migrate, archive, or continuously reread legacy GUI settings.
- Provide dedicated locations for configuration, window state, keymaps, backups, presets, asset data, cache, logs, plugins, and virtual environments.

## Preferences panel

- Add **Edit > Preferences** and a configurable `Ctrl+,` shortcut.
- Expose the shortcut through Input Settings and the Edit menu.
- Add a localized, theme-aware sidebar with responsive form rows and a fixed bottom action bar.
- Add General, Appearance, Input, Interface, and MCP sections.
- Persist language, theme, UI scale, camera-navigation mode, axis and view snapping, independent remember options, window state, performance HUD visibility, MCP configuration, scene-upscaler settings, and viewport splat quantization.
- Preserve the distinction between automatic UI scale and explicit numeric scale.
- Apply preference changes immediately.
- Keep Preferences, menus, shortcuts, toolbars, the viewport, and status controls synchronized.
[...]

PR #1586, opened 2026-08-08
Add Schema Versioning for Saved Layouts

### LINKED ISSUE #1486: Layout is saved on every frame of a window drag, and has no version field
**What's wrong**

Two small problems in the same file.

1. `LayoutState::save()` does a `create_directories` plus a full JSON write. It is called directly
   rather than through a debounce timer, so dragging a window edge writes the file every frame.
2. `layout.json` has no `version` field and no migration path. Any future change to its shape will
   either silently misread old files or require throwing them away.

**Where**

`src/visualizer/gui/layout_state.cpp:59-98` (save), `:100-146` (load — no version check)

**How to fix**

Add a debounce timer (save at most every ~500 ms after the last change), and add a `"version"`
field with a migration switch on load.

For reference, the input-bindings code in this repo already does versioned persistence properly
(`migrateLoadedProfile`, stable ids with a comment explaining why they must not be renumbered). That
is the pattern to copy — it should be the house standard for every persisted file.

shadygm:
## Summary

- add schema version 1 to `layout.json` writes
- treat existing unversioned layouts as version 0 and migrate them through an explicit sequential migration path
- reject malformed, negative, and future schema versions without mutating in-memory state or overwriting the file
- make read-modify-write callers honor layout load failures
- add focused coverage for versioned round-trips, legacy migration, unsupported versions, malformed versions, and first-save behavior

Closes #1486 

## Root cause and impact

Layout persistence previously had no schema marker. Older and newer documents were parsed solely by field name, so a future incompatible schema could be silently misread and then overwritten by the current application. Existing layout files remain compatible: a missing version is interpreted as the original version 0 schema, whose fields migrate directly to version 1.

## Why this PR does not add a 500 ms debounce

The issue also proposed debouncing layout writes based on the premise that resizing writes `layout.json` every frame. Static and runtime investigation did not reproduce that behavior:

- panel resize callbacks only update in-memory dimensions
- per-frame layout-change detection updates cached geometry and schedules render-settle frames; it does not persist
- `PanelLayoutManager::saveState()` is called during GUI shutdown
- VRAM HUD drag/resize persistence occurs on `Dragend` or other discrete actions
- a runtime trace using an isolated config issued 20 compositor window resizes at 50 ms intervals; file size, inode, and nanosecond mtime remained unchanged throughout, followed by one write during shutdown

A 500 ms timer would therefore introduce a new layout autosave/crash-recovery policy rather than eliminate current write amplification. That behavioral enhancement is intentionally left for a separate change with its own scheduling and shutdown-flush design.

## Validation

- `cmake --build cmake-build-release --target lfs_visualizer -j6`
- `cmake --build cmake-build-release --target LichtFeld-Studio -j6`
- standalone C++23 syntax check for `tests/test_layout_state.cpp`
- isolated runtime round-trip: an unversioned fixture loaded as version 0, preserved its values, and saved with `"version": 1`
- `git diff --check`

The monolithic GTest binary was not run locally because the checkout does not contain the optional `external/libtorch` SDK required to configure `BUILD_TESTS=ON`.


shadygm:
Awaiting #1503 to keep working on this

shadygm:
@MrNeRF, do we still want this?

MrNeRF:
@shadygm I close it. This is superseded. Sorry :(

Issue #1507, opened 2026-07-31
Export provenance stamps — every export should record what produced it


MrNeRF:
Part of #1505.

**Urgent by nature: every artifact exported without a stamp is permanently unattributable — this cannot be fixed retroactively.**

Users end up with `splat_30000 (1).ply` collections that cannot be attributed or reproduced. This issue adds a provenance stamp to every export with a native metadata slot.

**Design (as implemented on `feat/export-provenance`)**

Single-line JSON stamp, native slots only — no sidecar files (a sidecar detaches exactly when provenance matters; formats without a slot stay unstamped and documented):

```json
{"lichtfeld_provenance":1,"export_id":"<uuid4>","project":"…","commit":"…","node":"…","dataset":"…","iteration":30000,"strategy":"mcmc","app_version":"v0.5.3-…","exported_at":"2026-08-12T17:45:00Z"}
```

Fields are omitted when unknown. `project`/`commit`/`node`/`dataset` ship empty behind a single seam (`populate_project_identity`) until the .licht project format (#1525) merges; they are validated to a UUID alphabet so future wiring cannot leak paths or names. `export_id` is a fresh UUIDv4 per export — every artifact gets a unique identity immediately.

**Slots**

| Format | Slot |
|---|---|
| PLY | header `comment` line (binary + ASCII writers) |
| SOG | `meta["asset"]["lichtfeld_provenance"]` |
| .rad | `RadMeta::comment` (in-memory, streaming LOD, and rechunk writers) |
| USD/USDA/USDC | `customLayerData` |
| SPZ v4 | typed extension `0x4C465001` (since #1610; foreign readers skip it via length framing) |
| MP4 | container `comment` tag (GUI sequencer + headless camera-path) |
| HTML viewer | embedded JSON `<script>` block + the embedded SOG payload |
| PNG/JPEG | image `Comment` metadata (viewport/screenshot exports) |
| SPZ v3 escape hatch | — none (legacy readers reject trailing bytes); documented |
| COLMAP | — no slot; toggle hidden |

**Opt-out everywhere, default on**: export-panel checkbox (hidden where it would be a no-op), `--no-provenance` on convert/mesh2splat/train/render, `include_provenance=` on all Python `lf.io.save_*` / `lf.export_scene` / video export, and on the MCP export tools.

[...]

PR #1611, opened 2026-08-12
Export provenance stamps: every export records what produced it

MrNeRF:
Implements #1507 (issue body updated to match this design).

## What

A single-line JSON stamp embedded in the **native metadata slot** of every export — no sidecar files:

```json
{"lichtfeld_provenance":1,"export_id":"<uuid4>","iteration":30000,"strategy":"mcmc","app_version":"v0.5.3-…","exported_at":"2026-08-12T17:45:00Z"}
```

Fields are omitted when unknown. Reserved identity fields (`project`/`commit`/`node`/`dataset`) ship empty behind one seam function (`populate_project_identity`) so the #1525 `.licht` wiring is a ~15-line follow-up; they are validated to a UUID alphabet so that wiring can never leak a path or username into exports. `export_id` gives every artifact a unique identity from day one.

| Format | Slot |
|---|---|
| PLY | header `comment` (binary + ASCII writers) |
| SOG | `meta["asset"]` |
| .rad | meta `comment` — in-memory, **streaming LOD, and rechunk** writers |
| USD/USDA/USDC | `customLayerData` |
| SPZ v4 | typed extension `0x4C465001` — foreign readers skip it via the length framing (verified); v3 never carries it |
| MP4 | container `comment` tag — GUI sequencer **and** headless `--render-camera-path` |
| HTML viewer | embedded JSON block + the embedded SOG payload |
| PNG/JPEG | `Comment` metadata on viewport/screenshot exports |
| COLMAP | — no slot, exempt (toggle hidden) |

## Opt-out (default on)

Export-panel checkbox (auto-hidden where it would be a no-op: COLMAP, SPZ v3, with a tooltip listing exactly what gets written), `--no-provenance` on `convert` / `mesh2splat` / `train` / camera-path render, `include_provenance=` on all Python `lf.io.save_*` / `lf.export_scene` / `lf.ui.export_video`, and on the MCP export tools. With the toggle off, output bytes match pre-branch exports.

GUI exports stamp iteration + strategy from the live trainer; `convert` recovers both from `.resume` checkpoint headers. Trainer checkpoint PLYs (`splat_<iter>.ply`) stamp with exact iteration. GUI default filenames gain an `_<iteration>` suffix.

## Review hardening

The implementation went through a five-lane review (correctness, dead code, privacy/robustness, API consistency, plan audit); all findings were fixed:

- `export_id` now draws its full 128 bits from `std::random_device` (was: one 32-bit seed behind a Mersenne Twister — collision-prone at scale).
- The SPZ provenance extension caps payloads at 4 KiB and **skips** oversized ones — a hostile v4 file can no longer force a ~1 GiB allocation through the extension reader.
- RAD meta reads are bounded (16 MiB meta / 64 KiB comment; previously a raw `uint32` was trusted before allocating — pre-existing, now closed).
- The HTML embed escapes `<` (a hostile field value could otherwise truncate the page at `</script>`).
- Dead-code cleanup in touched files (dead `ExportState` member, unused includes, dead string-key constant, misplaced tinyply call) and the PLY header parser now skips spec-style `comment` lines deliberately rather than by accident.

## Tests

`tests/test_provenance.cpp` (14 tests, all runtime-generated data, no binary fixtures): JSON schema/omission/uniqueness, embed-and-reload per format (PLY binary/ASCII, SOG, RAD, USDA, NuRec USDZ, HTML, SPZ v4 incl. typed-extension load-back, PNG tEXt), stamp-absent checks for default-off paths and SPZ v3, plus CLI flag parsing and Python coverage. `lichtfeld.fast` and `lichtfeld.python.fast` pass; smoke-tested on a real 137k-gaussian scene (stamped PLY/SPZ verified byte-level, `--no-provenance` verified clean, stamped files round-trip).

## Notes

- MP4 metadata has no automated test (no headless encoder harness in the suite); verified by inspection of both encoder init paths.
- Follow-ups filed in the issue: #1525 identity wiring; asset-scanner RAD/SOG metadata parsing (pre-existing gaps).


MrNeRF:
Update: stamping is now non-optional — every slot-bearing export carries at least a minimal stamp (`lichtfeld_provenance`, `app_version`, `build_commit` = short git hash of the build). The toggle/flags now strip identifying fields (export id, timestamp, training info) rather than omitting the stamp; enforcement is in the writers themselves (empty-provenance falls back to minimal), verified by an adversarial audit over every export path. Training timelapse/eval images now carry the stamp too. PR description caveat: the earlier 'toggle-off output is byte-identical to pre-branch' note no longer applies — toggle-off now embeds the minimal build stamp.

PR #1615, opened 2026-08-13
PLY infinite opacity import repair

kmeirlaen:
# PLY infinite opacity import repair

## Summary

This patch keeps valid Gaussian splats when a PLY stores raw opacity logits as
`+inf` or `-inf`. The loader now repairs those opacity-only infinities to large
finite logits before tensors are created:

- `opacity == +inf` becomes `20.0f`
- `opacity == -inf` becomes `-20.0f`
- `opacity == NaN` remains invalid

All other Gaussian fields keep the existing non-finite validation behavior.

## Observed symptoms

Some valid exported PLY files encode fully opaque splats with
`opacity = +inf`, because Gaussian PLY opacity is stored as a raw logit and
`logit(1.0)` is positive infinity. Import validation treated every non-finite
float field the same way, so these fully opaque rows were discarded. When many
opaque rows were present, the imported scene could lose splats or fail entirely,
which changed the visual result.

## Reproduction steps

1. Create or export a Gaussian PLY where the vertex `opacity` property is a raw
   opacity logit and one splat has `opacity = +inf`.
2. Import the PLY through the common PLY loader.
3. Observe that the row is reported as invalid/discarded, even though
   `sigmoid(+inf)` represents valid fully opaque alpha.

## Root cause

The PLY Gaussian loader validates fields before uploading runtime tensors. That
validation rejected any non-finite float value without checking which property
it came from. For opacity this was too strict because the field is not linear
alpha; it is already the raw logit opacity value.

The slower compact fallback path also copied opacity directly from the original
row after validation, so it needed the same repair as the fast fused decode
path.

## What this patch fixes

- Adds opacity-specific infinity repair in the fast fused decode path.
- Adds the same repair in compact fallback extraction when invalid rows are
  present.
- Leaves NaN opacity invalid.
- Leaves non-finite position, scale, rotation, and SH validation unchanged.
- Preserves zero-length rotation validation.
- Keeps non-finite values out of runtime tensors.
- Adds aggregate repaired `+inf` and `-inf` opacity counts.
- Keeps discarded-row warnings limited to rows that are still actually
  discarded.

## Impact

GUI, CLI, Python API, and MCP imports all use the same PLY loading path, so the
behavior changes consistently across those surfaces without separate UI or API
changes.

## Tests

Added focused public-loader tests in `tests/test_ply_error_taxonomy.cpp`:

- `RepairsPositiveInfiniteOpacityInFusedDecode`
- `RepairsNegativeInfiniteOpacityInFusedDecode`
- `RepairsInfiniteOpacityWhenCompactingInvalidRows`
- `RejectsNaNOpacity`
- `RejectsNonFiniteGaussianFieldsExceptOpacityInfinity`
- `ExportImportOfRepairedOpacityRemainsFiniteAndOpaque`

Verification performed:

- To confirm the tests fail before the implementation the focused filter ran 6 tests:
  2 passed and 4 failed. The failing tests were the intended opacity-infinity
  cases:
  - `RepairsPositiveInfiniteOpacityInFusedDecode`
  - `RepairsNegativeInfiniteOpacityInFusedDecode`
  - `RepairsInfiniteOpacityWhenCompactingInvalidRows`
  - `ExportImportOfRepairedOpacityRemainsFiniteAndOpaque`
- The expected unaffected validation tests passed in the before run:
  - `RejectsNaNOpacity`
  - `RejectsNonFiniteGaussianFieldsExceptOpacityInfinity`
- After the changes, the same focused filter passed: 6 tests run, 6 passed.

### DIFF
--- a/src/io/formats/ply.cpp
+++ b/src/io/formats/ply.cpp
@@ -89,6 +89,8 @@ namespace lfs::io {
         constexpr size_t VALIDATION_CANCEL_INTERVAL = 65536;
         constexpr size_t MAX_HEADER_BYTES = 1024 * 1024;
         constexpr float MIN_ROTATION_NORM_SQUARED = 1.0e-12f;
+        constexpr float REPAIRED_POSITIVE_INFINITY_OPACITY_LOGIT = 20.0f;
+        constexpr float REPAIRED_NEGATIVE_INFINITY_OPACITY_LOGIT = -20.0f;
         // Load decode + save pack: cap below full HW concurrency to leave headroom
         // for GPU driver / other workers, but allow more than the old 6-thread limit.
         constexpr int MAX_DECODE_THREADS = 16;
@@ -367,9 +369,16 @@ namespace lfs::io {
         size_t invalid_count = 0;
         size_t non_finite_value_count = 0;
         size_t zero_rotation_count = 0;
+        size_t repaired_positive_infinite_opacity_count = 0;
+        size_t repaired_negative_infinite_opacity_count = 0;
 
         [[nodiscard]] size_t output_count(const size_t vertex_count) const {
             return valid_rows.empty() ? vertex_count : valid_rows.size();
+        }
+
+        [[nodiscard]] size_t repaired_opacity_count() const {
+            return repaired_positive_infinite_opacity_count +
+                   repaired_negative_infinite_opacity_count;
         }
     };
 
@@ -794,6 +803,25 @@ namespace lfs::io {
         return value;
     }
 
+    [[nodiscard]] float repair_infinite_opacity_logit(
+        const float value,
+        size_t* const positive_repair_count,
+        size_t* const negative_repair_count) {
+        if (std::isinf(value)) {
+            if (value > 0.0f) {
+                if (positive_repair_count != nullptr) {
+                    ++(*positive_repair_count);
+                }
+                return ply_constants::REPAIRED_POSITIVE_INFINITY_OPACITY_LOGIT;
+            }
+            if (negative_repair_count != nullptr) {
+                ++(*negative_repair_count);
+            }
+            return ply_constants::REPAIRED_NEGATIVE_INFINITY_OPACITY_LOGIT;
+        }
+        return value;
+    }
+
     [[nodiscard]] tbb::task_arena& ply_decode_arena() {
         static const int concurrency = std::max(
             1,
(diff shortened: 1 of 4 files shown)


src/io/formats/ply.cpp
```diff
@@ -89,6 +89,8 @@ namespace lfs::io {
         constexpr size_t VALIDATION_CANCEL_INTERVAL = 65536;
         constexpr size_t MAX_HEADER_BYTES = 1024 * 1024;
         constexpr float MIN_ROTATION_NORM_SQUARED = 1.0e-12f;
+        constexpr float REPAIRED_POSITIVE_INFINITY_OPACITY_LOGIT = 20.0f;
```
MrNeRF: One short constant is enough, e.g. `OPACITY_LOGIT_CLAMP = 20.0f`, and `std::copysign` covers the negative case.

src/io/formats/ply.cpp
```diff
@@ -367,10 +369,17 @@ namespace lfs::io {
         size_t invalid_count = 0;
         size_t non_finite_value_count = 0;
         size_t zero_rotation_count = 0;
+        size_t repaired_positive_infinite_opacity_count = 0;
```
MrNeRF: One combined `repaired_opacity_count` is enough. We never act on the +inf/-inf split, and it doubles the plumbing in the assert, error and warning strings.

src/io/formats/ply.cpp
```diff
@@ -794,6 +803,25 @@ namespace lfs::io {
         return value;
     }
 
+    [[nodiscard]] float repair_infinite_opacity_logit(
```
MrNeRF, line 799: With one counter and `copysign` this helper shrinks to a couple of lines or can be inlined at its two call sites.

src/io/formats/ply.cpp
```diff
@@ -960,6 +988,23 @@ namespace lfs::io {
             }
             return value;
         };
+        const auto read_opacity = [&](const char* row, const size_t offset, bool& invalid) {
```
MrNeRF, line 977: This debug assert duplicates the one in `read_field` right above — drop it. Same in the fused path below.

src/visualizer/window/vulkan_image_barrier_tracker.hpp
```diff
@@ -4,13 +4,15 @@
 
 #pragma once
 
+#include "core/export.hpp"
```
MrNeRF: Unrelated to the PLY fix — please move to its own PR.

tests/CMakeLists.txt
```diff
@@ -378,7 +378,8 @@ add_executable(lichtfeld_tests ${TEST_SOURCES})
 
 # USD (pxr) headers from vcpkg pull in deprecated <ext/hash_set> and emit
 # #warning (-Wcpp). Third-party; suppress only for this translation unit.
-set_source_files_properties(test_usd_format.cpp PROPERTIES COMPILE_OPTIONS "-Wno-cpp")
+set_source_files_properties(test_usd_format.cpp PROPERTIES
+    COMPILE_OPTIONS "$<$<CXX_COMPILER_FRONTEND_VARIANT:GNU>:-Wno-cpp>")
```
MrNeRF: Same here — unrelated build fix, separate PR please.

MrNeRF (commented):
Thanks, the fix itself is right: repair +/-inf opacity, keep rejecting NaN and zero rotations. Verified on a 13.4M-splat scene: no splats lost and load gets ~4x faster because the fast path stays active. Please trim as commented below so this stays a small fix.

PR #1629, opened 2026-08-15
Add persistent user preferences, recovery controls, and resilient window state

francescofugazzi:
## Summary

This PR introduces a unified, crash-safe system for application preferences and
user-owned storage, together with a complete Preferences surface, recovery
controls, safe-mode behavior, resilient desktop-window restoration, and a clear
ownership boundary between global application state and `.licht` project state.

The result is that settings now have one documented location and one persistence
policy, Preferences changes are reflected immediately throughout the interface,
projects retain their own workspace without taking ownership of desktop or
appearance settings, and damaged or unsuitable saved state can be recovered
without deleting unrelated user data.

## Unified user storage

The new `UserPaths` contract resolves every user-owned path from a single policy
instead of allowing individual subsystems to derive their own locations.

- Windows uses `%USERPROFILE%/.lichtfeld` as the standard root.
- Linux uses the same `~/.lichtfeld` unified root.
- `LFS_HOME` provides an explicit unified root on every supported platform.
- Portable builds use a `.lichtfeld` tree next to the executable.
- Explicit roots are supported for application callers and isolated tests.
- Legacy keymaps, theme/language choices, UI preferences, and `layout.json` are
  imported once without overwriting destinations or deleting their sources.

The resolved tree provides dedicated locations for:

- `config/preferences.json`;
- `config/ui_preferences.json`;
- `config/window.json`;
- `config/project_lifecycle.json`;
- keymaps;
- presets;
- Asset Manager data;
- reset backups;
- caches and Vulkan cache data;
- logs and session breadcrumbs;
- user plugins and their virtual environment.

The resolved directories are also published to the Python/plugin environment so
native and Python components use the same storage contract. Logger output,
session breadcrumbs, keymaps, Python tools, package management, user plugins,
the Asset Manager catalog, and related GUI state now follow these resolved paths.
The native application resolves the canonical log directory through `UserPaths`
before logger initialization; environment publication remains centralized in the
application entry point and exists only for Python/plugin consumers.

## Crash-safe persistence and recovery

User-owned text and JSON files are published through same-directory atomic
replacement. Writes create the destination directory as needed, use unique
temporary files, flush and close the payload before replacement, clean failed
temporaries, and preserve the previous valid destination when publication
fails. Windows writes use binary output so persisted bytes are not changed by
newline translation.

Preferences are held in a dedicated in-memory store instead of being reread for
every control change. If `preferences.json` is malformed, the original file is
closed and moved to a recovery backup before built-in defaults are used. The
condition is warned once and normal persistence resumes after a successful
backup. If the backup itself fails, the original file is preserved and writes
remain disabled rather than silently replacing user data.

Reset operations are recoverable rather than destructive:

- the previous file is moved to a collision-safe timestamped backup;
- preferences are restored to explicit built-in defaults;
- layout, UI state, window state, and project-lifecycle state can be reset
  independently;
- Reset All attempts every applicable target instead of stopping after the first
  failure;
- missing files are handled as a successful no-op;
- plugins, keymaps, caches, datasets, projects, and exported files are not
  removed.

The corresponding startup controls are available through:

| Option | Behavior |
| --- | --- |
| `--safe-mode` | Starts with built-in settings, user plugins disabled, and automatic user-state persistence disabled for the process. |
| `--reset-preferences` | Backs up application preferences and restores their built-in defaults before startup. |
| `--reset-layout` | Backs up and resets transitional layout and user-global UI state before startup. |
| `--reset-all-settings` | Applies preference and layout recovery and also resets window and project-lifecycle state. |

These flags are process/application controls and are not treated as training
configuration options. Reset failures are surfaced instead of being silently
ignored.

## Preferences interface

The new Preferences window is available from **Edit > Preferences** and from a
regular, rebindable `Ctrl+,` input action. The action is exposed consistently in
C++, Python, the Input Settings panel, and the shipped Python stubs.

Preferences uses a localized, theme-aware retained-mode interface with General,
Appearance, Input, and Interface sections. It provides immediate controls for:

- application language;
- theme;
- automatic or explicit UI scale;
- camera navigation mode;
- axis/view snapping;
- independent remember-navigation and remember-snap behavior;
- section reset;
- live interface-layout reset;
- desktop-window reset;
- complete settings reset.

Selections refresh from the active application state whenever the panel opens.
Changes are applied immediately and remain synchronized with menus, toolbar
buttons, input bindings, the startup language selector, and other controls that
expose the same state.

All new visible strings and recovery messages are present in the ten bundled
locale catalogs. Safe mode deliberately starts from built-in English without
overwriting the user's saved language.

## Safe mode

Safe mode is a process-local recovery environment. It can be requested through
`--safe-mode` or an existing `LFS_SAFE_MODE=1` environment value; a normal launch
does not clear an externally supplied safe-mode request.

While active, safe mode:

- disables user plugin loading;
- disables automatic writes for preferences, UI preferences, keymaps, window
  geometry, project-lifecycle state, and transitional layout state;
- isolates the Vulkan pipeline cache;
- prevents Asset Manager write probes or catalog migration;
- uses built-in application defaults;
- remains visibly identified in the window title and status bar;
- leaves explicit user actions such as keymap export separate from automatic
  persistence.

This allows the application to start without consuming or overwriting a broken
normal-session configuration.

## Desktop window state and multi-monitor recovery

Main-window position, size, and maximized state are stored exclusively in
`config/window.json`. Fullscreen is transient: when the application closes from
fullscreen, it persists the last valid windowed rectangle rather than the
fullscreen display dimensions.

Restored geometry is validated against the usable bounds of all currently
connected displays:

- valid positions on a secondary monitor are preserved;
- a window stranded by a disconnected monitor is recovered onto the primary
  display;
- a saved window larger than the currently available display is clamped and
  centered within that display;
- very small usable display areas are handled without producing invalid sizes;
- malformed or incomplete window state falls back safely;
- Reset Window restores a centered 1280x720 window;
- the first launch, or any launch without a valid `window.json`, starts with the
  same centered 1280x720 default and clamps it if the primary usable area is
  smaller.

Window-state persistence follows the same safe-mode and atomic-write policy as
the other user files.

## `.licht` project-format integration

This PR makes the GUIL ownership boundary explicit and enforceable.

Project-owned GUIL state continues to include the workspace itself:

- dock and panel arrangement;
- panel visibility;
- dock dimensions;
- active tabs;
- Sequencer visibility;
- scene-tree and panel-registry state;
- supported panel payloads and console workspace state.

User-global and desktop state is kept outside the project:

- main-window position and size;
- maximized and fullscreen state;
- theme;
- language;
- UI/DPI scale;
- user-global HUD preferences.

The canonical GUIL fixed-arrangement contract no longer emits or retains a
`window` member. Readers remove that legacy field during normalization before
validating the canonical payload. Validation still rejects `dpi` and
`dpi_scale`; these global fields are not silently stripped from malformed
project data. Other known user-global appearance/HUD fields are removed during
capture and normalization. Opaque plugin-owned payloads remain preserved rather
than being rewritten indiscriminately.

Opening a `.licht` project can therefore restore its saved workspace without
moving the desktop window or replacing the user's language, theme, or UI scale.
Creating a new project clears project content while preserving the live
workspace and desktop geometry in which the command was invoked. A live layout
reset restores the default docks, panels, tabs, Sequencer, scene-tree chrome,
and HUD state; when a project is open, that workspace becomes its next saved
GUIL state.

The session-state regression reported during review has been corrected. The
project-owned VIEW and GUIL payloads continue to round-trip selected nodes,
Gaussian selection masks, the active tool and selection submode, dock and panel
state, Sequencer visibility, and opaque extension payloads. Only the explicitly
user-global window and appearance fields are excluded. Existing GUIL v1 files
that still contain the former project-owned `window` member are accepted and
normalized without making an otherwise clean project appear modified. Their
next Save or Save As materializes the sanitized GUIL chapter, and Save As keeps
that pending normalization across document rebinding so the legacy bytes cannot
be copied back into the destination.

Project saving was also hardened so a new or otherwise frame-less project can be
saved when no viewport thumbnail is available. THMB remains optional, and an
existing project can retain its previous thumbnail instead of failing the whole
save operation.

The additional round-trip testing exposed two independent defects in the
existing training-save lifecycle; the affected code paths were unchanged from
`master` before these fixes:

- converting a paused or completed training session to Edit Mode removed the
  SCNG training binding but could leave its CKPT chapter behind, causing later
  manual or background saves to fail validation with an unbound checkpoint;
- the trainer's final snapshot writer could publish a newer `.licht` generation
  immediately before Edit Mode released the trainer, leaving the live
  `ProjectDocument` with clean proofs from the preceding generation and causing
  the next save to fail its append-authority check.

A full non-training synchronization now retires checkpoints after the explicit
conversion to an editable splat, while active-training autosaves still retain
their resumable checkpoint. The Edit Mode command also adopts any completed
trainer-owned generation before releasing the trainer and defers the transition
while the final snapshot writer is still in flight.

## Modal and input boundaries

Confirmation dialogs are centered against the complete application window and
use a theme-aware backdrop, so enlarged side or bottom docks do not displace
them.

The backdrop is an input boundary as well as a visual treatment. While a modal,
a pending modal, or a context menu blocks the underlay, pointer and
keyboard capture is established before native viewport controls are processed.
This prevents click-through to scene interaction and native overlays such as the
axis selector. Clicking a modal backdrop no longer cancels the dialog; explicit
cancel actions still invoke Python `on_cancel("")` callbacks.

`GuiFocusState` is now a single symbol exported by `lfs_visualizer` instead of an
inline header-local singleton. On Windows this ensures that the executable,
`lfs_visualizer.dll`, tests, and Python modules observe the same keyboard, text,
and pointer capture state rather than independent DLL-local copies.

## Additional fixes

The work also resolves several shell, lifecycle, and platform defects uncovered
while integrating and testing the new state model:

- camera-navigation and view-snap selections now keep both toolbar highlight
  layers synchronized when changed from Preferences;
- retained top-menu dropdowns materialize their updated rows before the next
  pointer event and ignore stale hidden layout boxes, preventing fast hover/click
  sequences from activating entries from the previous menu;
- **New Project** no longer opens panels, resets dock geometry, hides the
  Sequencer, or changes the main-window placement;
- the Asset Manager launcher remains visible in the default scene-panel header
  at high DPI by allowing the tab strip to shrink and elide safely;
- keymap reset handles an unavailable optional configuration directory and fixes
  the corresponding MSVC compilation failure;
- input-focus tests no longer load or depend on the developer's real persisted
  keymap;
- session breadcrumbs use the shared atomic writer and preserve exact bytes on
  Windows;
- visualizer teardown destroys project and GUI owners before the services and
  registries they reference, preventing post-test Windows SEH failures;
- floating-panel and hidden-dock state restore follows the unified panel and
  render-demand contracts;
- Asset Manager catalog publication preserves the prior catalog if an atomic
  replacement fails;
- user-path and preference persistence failures propagate through typed
  `lfs::Error`, `Result`, and `Status` contracts instead of string-based error
  channels, with a localized interface-preference persistence diagnostic in
  every shipped language.

## Documentation

The development documentation now describes:

- the unified default, overridden, and portable user roots;
- the resolved directory tree and ownership of each file;
- atomic publication and reset backups;
- safe-mode behavior;
- Preferences controls;
- window/fullscreen and multi-monitor recovery rules;
- first-launch geometry;
- the boundary between user-global settings and `.licht` GUIL state;
- legacy GUIL v1 normalization and its Save/Save As materialization contract;
- resumable CKPT ownership, Edit Mode conversion, and final trainer-generation
  adoption;
- New Project and live layout-reset behavior;
- full-window modal and backdrop input contracts.

## Verification

The branch was built successfully on Windows with MSVC/Ninja.
Focused validation covered:

- user-path resolution, portable and explicit roots, directory creation, atomic
  writes, backups, and resets;
- argument-parser safe-mode and reset contracts;
- theme, language, UI-scale, navigation, snap, and Preferences behavior;
- main-window restore, fullscreen persistence, disconnected/smaller-monitor
  recovery, reset, and first-launch centering;
- GUIL validation, global-state stripping, session restore, empty-tool restore,
  and New Project workspace preservation;
- legacy GUIL v1 window normalization on open, Save, and Save As without a
  spurious dirty prompt or reintroduction of the removed window payload;
- Gaussian selection-mask, selected-node, active-tool, and selection-submode
  restoration after Save, close, and reopen;
- paused resumable checkpoint restoration, checkpoint retirement after the
  explicit Edit Mode conversion, and adoption of an externally advanced final
  trainer generation before subsequent project saves;
- project-lifecycle settings, recent-project persistence, optional thumbnail
  behavior, and session breadcrumbs;
- input focus, text-entry suppression, rebindable Preferences actions, and
  backdrop input capture;
- menu resources, high-DPI scene-header behavior, Preferences Python contracts,
  and Asset Manager path migration;
- deterministic visualizer teardown on Windows.

The final combined window-state and input-focus run completed 72 tests. The
broader user-path, preference, session, window-state, and input-contract run
accounted for all 125 cases as passing after rerunning the corrected malformed
preferences regression. A separate logger, session-breadcrumb, and user-path run
completed all 32 tests. The selected Python contract suite completed all 35
tests across localization, Preferences, menu resources, and Asset Manager
migration.

The focused `.licht` persistence regressions above were all reported passing.
Manual validation also completed the full training workflow: stop and save a
training session, enter Edit Mode, select Gaussians, save from the close prompt,
and reopen the project with the editable state restored. The final training
generation was adopted before compaction and the previous clean-proof and
unbound-checkpoint failures did not recur.

`refresh_python_stubs` completed successfully. The committed `keymap.pyi`,
`packages.pyi`, `plugins.pyi`, and `ui/__init__.pyi` files match the generated
outputs exactly.

The repository quality gates additionally report:

- locale completeness across 2,139 English keys and every shipped locale;
- all 20 localization contract checks passing;
- all 45 error-debt census self-tests passing;
- no new error-debt violations, with the baseline regenerated after removing
  18 tracked `expected<string>` occurrences;
- a clean `clang-format --dry-run --Werror` check across all 66 changed C++/CUDA
  sources;
- a clean `git diff --check` result.


src/visualizer/gui/gui_manager.cpp
```diff
@@ -5610,10 +5700,21 @@ namespace lfs::vis::gui {
...
-                                           viewport_layout_.size.x,
-                                           viewport_layout_.size.y);
+                                           panel_input.screen_x,
+                                           panel_input.screen_y,
+                                           static_cast<float>(panel_input.screen_w),
+                                           static_cast<float>(panel_input.screen_h));
+            }
+            if (startup_overlay_.isVisible()) {
+                LOG_TIMER_THRESHOLD("gui_render.menu_context_modal_render.startup_overlay", 0.25);
+                startup_overlay_.render({
```
MrNeRF: Please go back to the master overlay: keep it as the `native.startup_overlay` viewport panel, centered in the viewport, no full-window backdrop. Right now it does not show up at all in some launches (`-v <ply>`), and the design must stay as it was.

src/visualizer/gui/startup_overlay.cpp
```diff
@@ -248,10 +255,7 @@ namespace lfs::vis::gui {
     }
 
     bool StartupOverlay::blocksUnderlayInput() const {
-        if (!visible_)
-            return false;
-        std::lock_guard lock(plugin_load_mutex_);
-        return !plugin_load_state_started_;
+        return visible_;
```
MrNeRF: Restore the master rule here too (stop blocking once plugin loading starts). Same behaviour as before, please.

src/core/user_paths.cpp
```diff
@@ -0,0 +1,444 @@
...
+        if (!home)
+            return userPathError(
+                lfs::ErrorCode::Unavailable,
+                "The current user's home directory is unavailable.",
+                "Neither the platform home variable nor a user-path override is available");
+
+#ifdef _WIN32
+        return fromUnifiedRoot(*home / ".lichtfeld");
+#else
+        const auto config_home = xdgOrHome("XDG_CONFIG_HOME", *home, ".config");
```
MrNeRF: Let's drop the XDG branch and use `fromUnifiedRoot(*home / ".lichtfeld")` on Linux as well. One root on every OS, `~/.lichtfeld/{config,data,cache,logs,plugins,venv}`. Plugins and venv already live there anyway.

src/core/logger.cpp
```diff
@@ -413,13 +420,42 @@ namespace lfs::core {
         }
 
         fs::path resolve_default_log_directory(const std::string& user_dir_override) {
-            return user_dir_override.empty()
-                       ? lichtfeld_home_directory() / ".lichtfeld"
-                       : fs::path(user_dir_override);
+            if (!user_dir_override.empty())
+                return fs::path(user_dir_override) / "logs";
+            if (const char* resolved = std::getenv("LFS_RESOLVED_LOG_DIR");
```
MrNeRF: This re-implements the path policy a second time (env var, `LFS_HOME`, portable, XDG). Please call `UserPaths::resolve()` here and delete the copy. Env vars should only be for the Python side.

src/app/main.cpp
```diff
@@ -55,6 +56,28 @@ namespace {
 #endif
     }
 
+    void publishResolvedUserPaths() {
```
MrNeRF, line 59: `LFS_RESOLVED_LOG_DIR` is also set in `argument_parser.cpp`. One place is enough. If the env publish is only for Python, say so in a one-line comment.

src/core/argument_parser.cpp
```diff
@@ -576,6 +581,14 @@ namespace {
                     filter_pattern = ::args::get(log_filter);
                 }
 
+                if (const auto paths = lfs::core::UserPaths::resolve()) {
+                    const auto log_directory = lfs::core::path_to_utf8(paths->logDir());
+#ifdef _WIN32
+                    (void)_putenv_s("LFS_RESOLVED_LOG_DIR", log_directory.c_str());
```
MrNeRF: Duplicate of `publishResolvedUserPaths()` in `main.cpp`. Pick one.

src/io/project/session_chapters.cpp
```diff
@@ -247,19 +247,9 @@ namespace lfs::io::project {
...
-                     {"fullscreen", false},
-                     {"maximized", false},
-                     {"restore_x", 0},
-                     {"restore_y", 0},
-                     {"restore_width", 1280},
-                     {"restore_height", 720},
-                 }},
+                // Kept as an empty compatibility object for GUIL v1 readers.
+                // Main-window state is user-global and lives in window.json.
+                {"window", Json::object()},
```
MrNeRF: `.licht` is brand new, no v1 compat needed. Please remove the `window` key entirely instead of keeping an empty object.

src/io/project/session_chapters.cpp
```diff
@@ -474,49 +464,97 @@ namespace lfs::io::project {
...
+        }
+
+        void strip_user_global_gui_state(Json& value) {
+            strip_user_global_gui_keys(value);
+            for_each_fixed_arrangement_payload(value, [](Json& payload) {
+                strip_user_global_gui_keys(payload);
+            });
+        }
+
+        void strip_main_window_state(Json& value) {
```
MrNeRF: Can go away together with the compat shim. Validation should just reject `window` if it shows up.

src/io/project/session_chapters.cpp
```diff
@@ -544,6 +582,12 @@ namespace lfs::io::project {
                         "The fixed GUIL arrangement is missing window state",
                         "GUIL.layouts.areas.spaces.fixed_arrangement");
                 }
+                if (!payload["window"].empty()) {
```
MrNeRF: "must exist but must be empty" is an odd rule. With no compat: reject if present, done.

src/io/project/session_chapters.cpp
```diff
@@ -474,49 +464,97 @@ namespace lfs::io::project {
...
-                strip_user_global_gui_state(
-                    item.value());
-                ++item;
             }
         }
 
+        bool contains_user_global_gui_state(const Json& value) {
+            if (object_contains_user_global_gui_state(value))
+                return true;
+            auto copy = value;
```
MrNeRF: This deep-copies the whole GUIL just to satisfy the non-const template. A const overload of `for_each_fixed_arrangement_payload` avoids the copy. Also `session_state.cpp` has the same walker written by hand; please share one.

src/visualizer/theme/theme.cpp
```diff
@@ -1050,44 +1053,70 @@ namespace lfs::vis {
...
+                return std::nullopt;
+            }
+            return paths->preferencesFile();
+        }
+
+        [[nodiscard]] bool preferencesDisabled() {
+            return lfs::core::environment::flag("LFS_SAFE_MODE", false);
+        }
+
+        [[nodiscard]] nlohmann::json loadPreferences() {
```
MrNeRF: If the file is corrupt, `file >> preferences` throws and every save above swallows it, so nothing persists any more and the user never learns why. Please: unparsable file -> back it up, start from `{}`, warn once.

src/visualizer/theme/theme.cpp
```diff
@@ -1096,34 +1125,171 @@ namespace lfs::vis {
...
+                        return scale;
+                }
             }
         } catch (const std::exception& e) {
             LOG_WARN("Failed to load UI scale preference: {}", e.what());
         }
         return 0.0f;
     }
 
+    void saveLanguagePreference(const std::string& language_code) {
```
MrNeRF: Language, camera navigation and view snap are not theme. Please move these into a `preferences.hpp/.cpp` (a small in-memory store that reads the file once and writes on change would also stop the read-twice-then-fsync pattern per click).

src/visualizer/theme/theme.cpp
```diff
@@ -3,11 +3,14 @@
 
 #include "theme.hpp"
 #include "core/config_paths.hpp"
```
MrNeRF: Dead include now. `core/config_paths.hpp` itself has no production users any more; please delete it and its test.

src/visualizer/theme/theme.hpp
```diff
@@ -223,10 +223,28 @@ namespace lfs::vis {
...
 
-    // UI scale preference (0.0 = auto from OS)
+    // UI scale preference. The API represents automatic OS scaling as 0.0;
+    // preferences.json serializes that selection as the explicit string "auto".
     LFS_VIS_API void saveUiScalePreference(float scale);
     [[nodiscard]] LFS_VIS_API float loadUiScalePreference();
 
+    // Language preference. An empty result means that the user has not chosen
+    // a language yet and the first-run startup selector should be shown.
+    LFS_VIS_API void saveLanguagePreference(const std::string& language_code);
```
MrNeRF: Same as in theme.cpp: these belong in a preferences header, not theme.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -6831,6 +6932,68 @@ namespace lfs::vis::gui {
         }
     }
 
+    void GuiManager::openPreferences() {
+        PanelRegistry::instance().set_panel_enabled("lfs.preferences", true);
+    }
+
+    std::expected<void, std::string> GuiManager::resetLayout() {
```
MrNeRF, line 6939: The `window_states_` defaults here are the same as in the constructor. One small `applyDefaultWindowStates()` please.

src/visualizer/project/session_state.cpp
```diff
@@ -1060,16 +1053,47 @@ namespace lfs::vis::project {
...
+                    lfs::io::project::
+                        kUserGlobalGuiFieldKeys
+                            .end()) {
                     return true;
                 }
             }
             return false;
         }
 
+        bool contains_gui_global_field(const Json& root) {
```
MrNeRF, line 1064: Same nested walk as `for_each_fixed_arrangement_payload` in `session_chapters.cpp`. Please share it instead of writing it twice.

src/visualizer/project/session_state.cpp
```diff
@@ -1038,17 +1037,11 @@ namespace lfs::vis::project {
...
-                    value,
-                    contains_gui_global_field);
-            }
+        bool object_contains_gui_global_field(const Json& value) {
             if (!value.is_object())
                 return false;
-            for (const auto& [key, child] :
-                 value.items()) {
+            for (const auto& [key, child] : value.items()) {
+                (void)child;
```
MrNeRF: Iterate keys instead of `(void)child`.

src/visualizer/gui/rml_modal_overlay.cpp
```diff
@@ -562,7 +562,9 @@ namespace lfs::vis::gui {
         const auto& id = target->GetId();
 
         if (id == "modal-backdrop") {
-            overlay->cancel();
+            // A modal choice must be made explicitly through one of its
+            // buttons. Clicking the backdrop is intentionally inert.
```
MrNeRF, line 566: Behaviour change (backdrop click used to cancel). Fine by me, but please add it to the PR description.

src/python/lfs/py_ui_modals.cpp
```diff
@@ -88,13 +88,12 @@ namespace lfs::python {
                             LOG_ERROR("Modal callback error: {}", e.what());
                         }
                     };
-                    const std::string cancel_label = modal.buttons.size() >= 2
-                                                         ? modal.buttons.back()
-                                                         : "";
-                    req.on_cancel = [py_cb, cancel_label]() {
+                    // Cancellation is state cleanup, not an implicit click on
+                    // whichever button happens to be last.
+                    req.on_cancel = [py_cb]() {
```
MrNeRF, line 93: Plugin-facing change (`on_cancel` now gets `""` instead of the last label). Please mention it in the description.

src/core/tensor/cuda_event_pool.cpp
```diff
@@ -98,7 +98,11 @@ namespace lfs::core {
         std::lock_guard<std::mutex> lock(mutex_);
         for (cudaEvent_t event : pool_) {
             const cudaError_t destroy_status = cudaEventDestroy(event);
-            if (destroy_status != cudaSuccess) {
+            // The runtime may already be unloading during process teardown.
+            // At that point the event is unusable and cleanup is effectively
+            // complete; keep reporting every other destruction failure.
+            if (destroy_status != cudaSuccess &&
+                destroy_status != cudaErrorCudartUnloading) {
```
MrNeRF, line 105: Fine, but this is unrelated to preferences. Own commit please so it is easy to find later.

src/python/stubs/lichtfeld/keymap.pyi
```diff
@@ -77,6 +77,8 @@ class Action(enum.Enum):
 
     TOGGLE_PERFORMANCE_HUD = 77
 
+    OPEN_PREFERENCES = 78
```
MrNeRF: Hand-edited: the generator puts this at the end. `cmake --build build --target refresh_python_stubs` and commit the result (this file plus `packages.pyi`, `plugins.pyi`, `ui/__init__.pyi`), otherwise the build fails on `python_stub_check`.

src/visualizer/input/input_bindings.cpp
```diff
@@ -235,14 +252,28 @@ namespace lfs::vis::input {
...
+        const auto path = *config_dir / (name + ".json");
         saveProfileToFile(path);
     }
 
-    std::filesystem::path InputBindings::getConfigDir() {
-        return lfs::core::user_config_dir() / "input_profiles";
+    std::optional<std::filesystem::path> InputBindings::getConfigDir() {
+        const auto paths = lfs::core::UserPaths::resolve();
+        if (paths)
+            return paths->keymapDir();
```
MrNeRF, line 269: Old keymaps live in `<old config>/input_profiles/Default.json`. Without a copy every user loses their bindings on upgrade. A one-shot import here would be nice.

docs/docs/development/preferences-and-user-storage.md
```diff
@@ -0,0 +1,140 @@
...
+same-directory atomic replacement so an interrupted write cannot expose a
+partially written destination.
+
+## User storage roots
+
+The default root follows the platform policy implemented by `UserPaths`.
+Explicit application roots use one isolated unified storage tree. Portable
+builds use a `.lichtfeld` tree next to the executable. `LFS_HOME` remains an
+explicit override on every platform and in portable builds. The application
+does not scan or import old settings directories automatically. The Asset
```
MrNeRF: If we keep "no import", this needs to be very loud in the release notes: keymaps, theme, language and layout reset for everyone.

MrNeRF (changes requested):
Thanks Francesco, this is solid work. Clean diff, real dead code removed, tests are real, typed errors everywhere. A few things I want changed before merge, all of them in this PR please (no follow-ups):

**1. Startup overlay: restore the master design and make it always show.** Right now it does not appear in some launches (`LichtFeld-Studio -v <ply>` for example). It must be centered in the viewport like before, no full-window backdrop, same input rules as master. Please put the `native.startup_overlay` panel back and drop the direct render in the modal pass.

**2. One settings root on every OS.** `~/.lichtfeld/{config,data,cache,logs,plugins,venv}` on Linux too, same as Windows. Drop the XDG split; today Linux ends up with five locations. `fromUnifiedRoot` already does the right thing for `LFS_HOME`/portable, so this is deleting code (XDG branch in `UserPaths::resolve`, XDG fallback in `logger.cpp`, XDG candidates in `asset_index.py`). One folder a user can zip for a bug report beats XDG purity.

**3. No `.licht` compat shim.** The format is brand new, we owe v1 nothing. Remove the empty `window` object, reject `window` if present, delete `strip_main_window_state`, and reject `dpi`/`dpi_scale` instead of stripping.

**4. Corrupt `preferences.json` must not silently kill saves.** `loadPreferences()` throws on a parse error and every `save*Preference` swallows it, so theme/language/scale changes stop persisting with no signal. Treat an unparsable file as empty (back it up first) and warn once.

**5. Migration.** Keymaps, theme, language, layout move to new folders with no import, so every user starts from defaults after upgrade. Please add a one-shot copy of `input_profiles/Default.json` and the old prefs, or at least call it out loudly.

**6. Regenerate the Python stubs.** The build fails on `python_stub_check` here: `cmake --build build --target refresh_python_stubs` and commit the 4 files.

Smaller ones are inline. Also please mention the behaviour changes in the description (backdrop click no longer cancels modals, Python modal `on_cancel` now passes `""`).

francescofugazzi:
Thanks for the detailed review. I addressed the requested changes in the latest update:

- restored the upstream startup-overlay contract: it is once again a viewport overlay, remains viewport-centered, has no full-window backdrop, appears for startup file loads as well, and stops blocking the underlay once plugin loading begins;
- unified the default user-storage layout under `~/.lichtfeld` on Windows and Linux, and removed the duplicated XDG path policies from `UserPaths`, the logger, and the Python Asset Index;
- retained a dedicated, one-shot legacy-settings migration without mixing legacy discovery into the current path-resolution contract;
- removed the GUIL `window` compatibility object entirely and made `window`, `dpi`, and `dpi_scale` invalid in fixed-arrangement project payloads;
- replaced the duplicated GUIL traversal logic with a shared const/mutable walker that does not copy the JSON DOM;
- separated application preferences from the theme implementation and introduced an in-memory preference store with atomic persistence;
- corrupt `preferences.json` files are now backed up before being treated as empty, produce a single warning, and allow persistence to resume only after a successful backup;
- removed the obsolete `config_paths.hpp` implementation and its tests;
- consolidated repeated default window-state initialization in a shared helper;
- updated the user-storage documentation and PR description to match the final contracts;
- regenerated the error-debt baseline after removing obsolete exception-handling sites;
- removed the unrelated CUDA shutdown fix from this PR.

I also updated the focused tests for the unified paths, legacy migration, corrupt-preference recovery, strict GUIL validation, and the revised startup contract. The generated Python stubs will be refreshed with the repository target and included in the update.

francescofugazzi:
Fixed a `.licht` regression affecting project-state restoration.

Project-owned state now round-trips correctly again: selected nodes, Gaussian selection masks, the active tool and selection submode, workspace state, and paused resumable training state survive Save, close, and reopen. Main-window geometry, theme, language, and UI scale remain correctly separated as user-global state.

Legacy GUIL v1 projects containing the former `window` member are now accepted and normalized without creating a false dirty state. The sanitized chapter is materialized on the next Save or Save As, including across Save As document rebinding.

While adding the missing format-level regression coverage, I also found and fixed two separate `.licht` save-lifecycle problems in code paths that were already present unchanged in `upstream/master`:

- switching a paused or completed training session to Edit Mode could remove its SCNG binding while leaving the CKPT chapter behind, causing subsequent saves to fail with an unbound-checkpoint validation error;
- the trainer could publish a newer final `.licht` generation immediately before Edit Mode released it, leaving the live document with clean proofs from the previous generation and causing the next save to fail its append-authority check.

Edit Mode now retires the former checkpoint only during the full post-conversion synchronization, preserving resumable active-training autosaves. It also adopts any completed trainer-owned generation before releasing the trainer and defers the transition while the final writer is still active.

Supplemental regression coverage now exercises:

- legacy GUIL normalization through open, Save, and Save As;
- selected Gaussian masks, selected nodes, active tools, and selection submodes across reopen;
- paused checkpoint restoration as a resumable training session;
- CKPT retirement after conversion to Edit Mode;
- adoption of a trainer-published newer generation before the following save.

All focused tests passed, and I manually verified the complete training stop/save → Edit Mode → Gaussian selection → Save and Close → reopen workflow without either persistence failure recurring.

MrNeRF:
Merge. Looks fine!

PR #1632, opened 2026-08-15
Fix CUDA event-pool cleanup during runtime shutdown

francescofugazzi:
Treat cudaErrorCudartUnloading as an expected terminal condition when destroying pooled tensor events during process shutdown. At that point the CUDA runtime is already unavailable and the event cannot be reused, so reporting the condition as a cleanup failure only adds noise and can destabilize teardown-focused tests.

Continue routing every other cudaEventDestroy failure through the existing diagnostic path, preserving error visibility for genuine resource-lifecycle defects.


MrNeRF:
Same as here, right? https://github.com/MrNeRF/LichtFeld-Studio/pull/1629/changes#diff-1d7e3ca57963e07249c179a868030ebfac766fe36084d875d4340fba1ab1c1a0

I can merge it here already. Thank you.

francescofugazzi:
yes i removed from #1629 and soon i'm pushing an update

PR #1634, opened 2026-08-16
fix: accept partial optimization JSON; find uv on PATH

### LINKED ISSUE #1541: Allow subset of configuration parameters when loading JSON config.
In our workflow the default config parameters work fine and I mainly want to adjust some of the optimization parameters like "grad_threshold", "stop_refine" etc.
The issue is that LFS expects a full config where no defined key is missing as it seems.
To make our workflow future proof, I would always have to export the current config from LFS first, then load it, change key values and export again in the case that new config keys were added, breaking potentially existing config files.

It would be much more convenient if the config parser would accept subsets of key-value pairs instead of completely stopping to parse when a key is missing. 
Preferrably just skip the keys that were not found and use the defined default value.

TLDR:
Make this config (Example) usable as config input:
```
{
    "optimization": {
        "grad_threshold": 0.002,
        "stop_refine": 50000
    }
}
```

### LINKED ISSUE #1569: Source builds silently omit uv.exe, breaking all plugin installs with "UV not found"
When LichtFeld Studio is built from source and the FetchUV download step fails during CMake configure, the build still completes successfully, but every subsequent plugin install fails with a UV error, and nothing at build time or run time tells the user why or how to fix it.

**Observed/Reported behavior:**

1. cmake/FetchUV.cmake downloads uv 0.10.2 at configure time. On download failure it emits only message(WARNING "FetchUV: uv will not be bundled. Users can install it manually.") and returns. That warning is easy to miss in a long configure log, and it is non-fatal unless BUILD_PORTABLE is set (src/python/CMakeLists.txt).
2. At runtime, PackageManager::uv_path() (src/python/package_manager.cpp) looks only in <exe_dir>/bin/uv.exe and <exe_dir>/uv.exe. It never falls back to the system PATH, so even a user who has uv installed globally (pip/winget/standalone) still gets "UV not found".
3. The failure surfaces only when the user tries to install a plugin, far from the cause. A real-world report: a user built from source, hit the UV error on plugin install, and had to discover by trial and error that copying a pip-installed uv.exe into the build output folder fixes it.

**LLM Ideas:**

* In PackageManager::uv_path(), fall back to searching PATH when no bundled uv is found (optionally logging that an unpinned uv version is in use).
* Make the runtime error message actionable: "uv.exe not found next to the LichtFeld Studio executable. Download uv 0.10.2 from https://github.com/astral-sh/uv/releases and place uv.exe beside the executable."
* Promote the FetchUV download failure to FATAL_ERROR (or a prominent end-of-configure summary) so source builders learn about the gap at build time, when it's cheap to fix.

fredcrawford:
## Summary
- Allow partial optimization JSON: missing keys keep strategy defaults instead of requiring every `json_required` field.
- Fall back to `uv` on `PATH` when no bundled binary is next to the executable, with a clearer error message (#1569).

## Test plan
- [ ] Load a JSON with only `optimization.grad_threshold` / `stop_refine` and confirm training starts with other defaults
- [ ] Build without bundled `uv`, install `uv` on PATH, confirm plugin install finds it
- [ ] Full configs and bundled `uv` still work

Fixes #1541
Fixes #1569

### DIFF
--- a/src/core/parameters.cpp
+++ b/src/core/parameters.cpp
@@ -99,7 +99,8 @@ namespace lfs::core {
 
                 for (const auto& meta : group.properties) {
                     const std::string key(optimization_json_key(meta));
-                    if (!meta.json_required && !json.contains(key))
+                    // Missing keys keep strategy defaults (partial config overlays).
+                    if (!json.contains(key))
                         continue;
 
                     const auto& value = json.at(key);
--- a/src/python/package_manager.cpp
+++ b/src/python/package_manager.cpp
@@ -12,10 +12,12 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <cstdlib>
 #include <optional>
 #include <regex>
 #include <sstream>
 #include <vector>
+#include <string>
 
 #ifdef _WIN32
 #include <windows.h>
@@ -268,10 +270,46 @@ namespace lfs::python {
             bundled = true;
         }
 
-        if (cached.empty())
-            LOG_WARN("Bundled uv not found");
-        else if (bundled)
+        if (cached.empty()) {
+            // Fall back to PATH when FetchUV did not bundle uv next to the binary.
+#ifdef _WIN32
+            wchar_t found[MAX_PATH];
+            if (SearchPathW(nullptr, L"uv.exe", nullptr, MAX_PATH, found, nullptr) > 0) {
+                cached = found;
+            }
+#else
+            if (const char* path_env = std::getenv("PATH")) {
+                std::string paths = path_env;
+                size_t start = 0;
+                while (start <= paths.size()) {
+                    const size_t end = paths.find(':', start);
+                    const auto dir = paths.substr(start, end == std::string::npos ? std::string::npos : end - start);
+                    if (!dir.empty()) {
+                        const auto candidate = std::filesystem::path(dir) / UV_BINARY;
+                        std::error_code ec;
+                        if (std::filesystem::exists(candidate, ec) && !ec) {
+                            cached = candidate;
+                            break;
+                        }
+                    }
+                    if (end == std::string::npos)
+                        break;
+                    start = end + 1;
+                }
+            }
+#endif
+        }
+
+        if (cached.empty()) {
+            LOG_WARN(
+                "uv not found beside the executable or on PATH. "
+                "Download uv from https://github.com/astral-sh/uv/releases and place it next to the binary, "
+                "or install uv so it is available on PATH.");
+        } else if (bundled) {
             LOG_INFO("Using bundled uv: {}", lfs::core::path_to_utf8(cached));
+        } else {
+            LOG_INFO("Using PATH uv (unpinned): {}", lfs::core::path_to_utf8(cached));
+        }
 
         return cached;
     }


src/core/parameters.cpp
```diff
@@ -99,7 +99,8 @@ namespace lfs::core {
 
                 for (const auto& meta : group.properties) {
                     const std::string key(optimization_json_key(meta));
-                    if (!meta.json_required && !json.contains(key))
+                    // Missing keys keep strategy defaults (partial config overlays).
+                    if (!json.contains(key))
```
MrNeRF, line 103: Please revert this. #1541 is closed on purpose: config loading should stay strict. This also breaks `TrainingParametersTest.MissingRequiredJsonKeyStillThrows` in `tests/test_training_parameters.cpp:301` and leaves `json_required` in the registry with nothing reading it.

src/python/package_manager.cpp
```diff
@@ -12,10 +12,12 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <cstdlib>
 #include <optional>
 #include <regex>
 #include <sstream>
 #include <vector>
+#include <string>
```
MrNeRF, line 20: Includes must be sorted, `clang-format` will move this above `<vector>`. Please run `clang-format` on the file.

src/python/package_manager.cpp
```diff
@@ -268,10 +270,46 @@ namespace lfs::python {
...
         }
 
-        if (cached.empty())
-            LOG_WARN("Bundled uv not found");
-        else if (bundled)
+        if (cached.empty()) {
+            // Fall back to PATH when FetchUV did not bundle uv next to the binary.
+#ifdef _WIN32
+            wchar_t found[MAX_PATH];
+            if (SearchPathW(nullptr, L"uv.exe", nullptr, MAX_PATH, found, nullptr) > 0) {
```
MrNeRF, line 277: When the buffer is too small, `SearchPathW` returns the needed size (larger than `MAX_PATH`) and does not fill `found`. `> 0` treats that as success. Please check `n > 0 && n < MAX_PATH` (or use the file's `MAX_PATH_LEN`).

Small note: with `lpPath = nullptr` this also searches the current directory before PATH, unlike the Linux branch below which skips it. Walking `PATH` with `;` on Windows would keep both branches the same.

src/python/package_manager.cpp
```diff
@@ -268,10 +270,46 @@ namespace lfs::python {
...
+            if (const char* path_env = std::getenv("PATH")) {
+                std::string paths = path_env;
+                size_t start = 0;
+                while (start <= paths.size()) {
+                    const size_t end = paths.find(':', start);
+                    const auto dir = paths.substr(start, end == std::string::npos ? std::string::npos : end - start);
+                    if (!dir.empty()) {
+                        const auto candidate = std::filesystem::path(dir) / UV_BINARY;
+                        std::error_code ec;
+                        if (std::filesystem::exists(candidate, ec) && !ec) {
```
MrNeRF, line 290: `exists()` also matches a folder or a non-executable file named `uv`, and the result is cached for the whole run. Please check `is_regular_file` plus `access(path, X_OK)`.

src/python/package_manager.cpp
```diff
@@ -268,10 +270,46 @@ namespace lfs::python {
...
-        else if (bundled)
+        if (cached.empty()) {
+            // Fall back to PATH when FetchUV did not bundle uv next to the binary.
+#ifdef _WIN32
+            wchar_t found[MAX_PATH];
+            if (SearchPathW(nullptr, L"uv.exe", nullptr, MAX_PATH, found, nullptr) > 0) {
+                cached = found;
+            }
+#else
+            if (const char* path_env = std::getenv("PATH")) {
```
MrNeRF, line 281: Optional: `src/visualizer/gui/editor/python_lsp_client.cpp` already has the same PATH lookup (`resolve_program_path` + `is_executable_file`). One small shared helper, for example `find_program_on_path()` in `core/executable_path.hpp` (already included here), used by both would be nicer than a second copy. It would also fix the two points above in one place.

src/python/package_manager.cpp
```diff
@@ -268,10 +270,46 @@ namespace lfs::python {
...
+                        break;
+                    start = end + 1;
+                }
+            }
+#endif
+        }
+
+        if (cached.empty()) {
+            LOG_WARN(
+                "uv not found beside the executable or on PATH. "
```
MrNeRF, line 305: Nice message, but it only goes to the log once. The errors users actually see still say plain "uv not found": `src/python/lfs/py_packages.cpp` (5 throws), `src/python/uv_runner.cpp:22`, `src/python/plugin_runner.cpp:68`. Please put the message in one place on `PackageManager` and use it there too, so #1569 is fixed where the reporter hit it.

src/python/package_manager.cpp
```diff
@@ -268,10 +270,46 @@ namespace lfs::python {
             bundled = true;
         }
 
-        if (cached.empty())
-            LOG_WARN("Bundled uv not found");
-        else if (bundled)
+        if (cached.empty()) {
+            // Fall back to PATH when FetchUV did not bundle uv next to the binary.
```
MrNeRF, line 274: Nit: the log line below already says this, the comment can go.

MrNeRF (changes requested):
Thanks for the PR! Two things:

**Please drop the `parameters.cpp` change (#1541).** I closed that issue on purpose, we want config loading to be strict, not looser. This change also breaks an existing test: `TrainingParametersTest.MissingRequiredJsonKeyStillThrows` (CI did not run it because the Ubuntu job died at configure).

**The uv-on-PATH part is welcome.** A few fixes before it can go in, see the inline comments. Short version:

- run `clang-format`
- Windows: handle the `SearchPathW` "buffer too small" return
- Linux: check the candidate is really an executable file
- route the new helpful message to the errors users actually see
- optional: share one PATH lookup helper with `python_lsp_client.cpp`

Happy to merge once the uv part is cleaned up.

MrNeRF:
Closing due to inactivity. Feel free to reopen or post here.

PR #1635, opened 2026-08-16
Restore training datasets when reopening projects

francescofugazzi:
## Summary

Projects saved immediately after importing a dataset could reopen with the persisted preview scene, but without a usable trainer.

This change stores a durable reference to the external dataset and reloads it when reopening a dataset project saved before its first checkpoint.

## What changed

- Store the live dataset path in the project parameters.
- Add a durable dataset reference to `PROJ` and `REFS`.
- Require both a persisted SCNG `dataset` node and a resolvable dataset reference for no-checkpoint restoration.
- Reload the complete dataset asynchronously instead of creating a trainer from preview-only geometry.
- Preserve the existing checkpoint restoration path.
- Avoid carrying stale dataset paths into PLY-only projects.
- Log that the hydrated scene is rebuilt from the external dataset.

The dataset remains external and is not embedded in the `.licht` file. Projects without a durable dataset reference are not recovered through directory heuristics.

## Testing

Added regression coverage for:

- reopening a dataset project saved before its first checkpoint;
- refusing implicit recovery when the dataset reference is missing;
- preserving a PLY-only project stored inside a valid dataset directory;
- preventing stale dataset parameters from being bound to a non-dataset project.

Manually verified by importing a dataset, changing training settings, saving before training, reopening the project, and confirming that the dataset and trainer were restored.


src/visualizer/project/project_lifecycle.cpp
```diff
@@ -239,13 +239,110 @@ namespace lfs::vis::project {
...
-                        "Project has a training model but no checkpoint to resume");
+                const auto dataset =
+                    document.project().dataset_reference();
+                const auto project_root =
+                    projectRootFor(document);
+                const bool has_explicit_dataset_source =
+                    (dataset && *dataset) ||
+                    !report.pending_parameters.dataset
+                         .data_path.empty();
+                if (has_explicit_dataset_source ||
```
MrNeRF: Blocker: `project_root` is essentially always non-empty, so this branch runs for every project without a checkpoint, including PLY-only projects. The reload below goes through `applyLoadedDataset` -> `clear()` and replaces the whole scene. Keep master's gate here: only proceed when `document.scene_graph().training_model_uuid()` is set. Only dataset imports call `setTrainingModelNode`, PLY loads never do, so that is the reliable "this was a dataset project" signal.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -239,13 +239,110 @@ namespace lfs::vis::project {
...
+                        viewer.getTrainerManager();
+                    if (!parameter_manager ||
+                        !trainer_manager) {
+                        notifyTrainerRestoreFailure(
+                            viewer,
+                            "Project has no trainer manager or parameter manager");
+                        return;
+                    }
+
+                    std::optional<std::filesystem::path>
```
MrNeRF: This REFS lookup + `pending_parameters.dataset.data_path` fallback duplicates `resolveDatasetRootForTrainer` a few lines above. Reuse it and only add the containing-dir fallback on top.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -239,13 +239,110 @@ namespace lfs::vis::project {
...
+                            dataset_root = *resolved;
+                        }
+                    }
+                    if (!dataset_root &&
+                        !report.pending_parameters.dataset
+                                 .data_path.empty()) {
+                        dataset_root = report.pending_parameters
+                                           .dataset.data_path;
+                    }
+                    // Projects saved before dataset references were captured
```
MrNeRF: Without the training-model gate this heuristic also fires for a PLY-only `.licht` that happens to live inside a COLMAP folder: the PLY scene gets replaced by a dataset import. With the gate in place this fallback is fine.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -239,13 +239,110 @@ namespace lfs::vis::project {
...
+                            LOG_INFO(
+                                "Recovered project dataset from its containing directory ({})",
+                                lfs::core::path_to_utf8(
+                                    project_root));
+                        }
+                    }
+                    if (!dataset_root ||
+                        dataset_root->empty() ||
+                        !std::filesystem::exists(*dataset_root)) {
+                        if (has_explicit_dataset_source) {
```
MrNeRF: When nothing resolves and there is no explicit source this now returns silently. Keep the old "Project has a training model but no checkpoint to resume" toast for that fallthrough.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -239,13 +239,110 @@ namespace lfs::vis::project {
...
+                    data_loader->setParameters(params);
+                    scene_manager.setDatasetPath(*dataset_root);
+                    lfs::core::events::cmd::LoadFile{
+                        .path = *dataset_root,
+                        .is_dataset = true,
+                        .output_path = params.dataset
+                                           .output_path,
+                    }.emit();
+                    LOG_INFO(
+                        "Queued project dataset reload for trainer restoration (dataset={})",
```
MrNeRF: Note: this reimport clears the scene and re-mints node UUIDs, so the project goes dirty right after opening and any node edits made between import and save (renames, transforms, extra PLYs, selection) are dropped. Acceptable for the target case, but the log/toast should say the scene was rebuilt from the dataset so this is not a surprise.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -3484,6 +3581,62 @@ namespace lfs::vis::project {
...
+            if (dataset_path.empty()) {
+                if (const auto* trainer =
+                        viewer_.getTrainer()) {
+                    dataset_path = trainer->getParams()
+                                       .dataset.data_path;
+                }
+            }
+            if (dataset_path.empty()) {
+                dataset_path =
+                    snapshot->dataset.data_path;
```
MrNeRF: This fallback is the leak: `dataset_config_.data_path` in the parameter manager survives a scene clear (only `resetProjectState`/`clearSession` resets it). Import a dataset, then open a PLY, then save: the scene has no dataset, but this still mints a `dataset` REFS entry, and on reopen the PLY scene is replaced by the old dataset. Drop this fallback here, or bind it to the scene actually having a training model.

src/visualizer/project/project_lifecycle.cpp
```diff
@@ -4784,9 +4937,7 @@ namespace lfs::vis::project {
                                             now();
                                     // Trainer restore is soft: display
                                     // hydration already succeeded.
-                                    if (report
-                                            .trainer_state_pending &&
-                                        epoch_.load(
+                                    if (epoch_.load(
```
MrNeRF, line 4945: Removing the `trainer_state_pending` guard is what makes the no-checkpoint branch reachable at all, so every project open now goes through it. That is fine, it just makes the training-model gate above mandatory.

MrNeRF (changes requested):
Storing the dataset path in REFS and re-importing on open is the right shape.

One blocker: the reload branch runs for any project without a checkpoint and does a full scene clear + reimport (`applyLoadedDataset` -> `clear()`). That can wipe projects that were never dataset projects, e.g. import a dataset, then open a PLY in the same session, save (the stale `dataset.data_path` in the parameter snapshot still mints a dataset ref), reopen -> PLY is gone, replaced by the old dataset. Same for a PLY-only `.licht` saved inside a COLMAP folder via the containing-dir heuristic. Please keep master's gate, `scene_graph().training_model_uuid()` set, in front of the whole reload block; only dataset imports set that.

Smaller things:
- When nothing resolves, keep the old "training model but no checkpoint" toast instead of returning silently.
- The REFS / pending-params lookup duplicates `resolveDatasetRootForTrainer` just above; reuse it and only add the containing-dir fallback.
- Since reimport re-mints node UUIDs, the project goes dirty right after opening and any node edits made before saving are lost. Fine for the target case, but worth a log line saying the scene was rebuilt from the dataset.

Details inline.

francescofugazzi:
I addressed the restoration and stale-state issues:

- No-checkpoint restoration now requires both a persisted SCNG `dataset` node and a resolvable `PROJ.dataset_reference`.
- I intentionally use the dataset node instead of relying only on `training_model_uuid`, because a dataset saved before training starts may not have a training-model node yet.
- Removed the containing-directory recovery heuristic entirely; projects without a durable dataset reference are no longer recovered implicitly.
- Reused `resolveDatasetRootForTrainer()` for reference resolution.
- Removed the stale `snapshot->dataset.data_path` fallback and only bind dataset references while the live scene is actually in Dataset mode.
- Preserved the existing missing-checkpoint diagnostic and made the rebuild explicit in the log.
- Added regression coverage for no-checkpoint restoration, missing references, PLY projects inside dataset directories, and stale dataset parameters.

MrNeRF:
Thank you! Looks good to me!

PR #1637, opened 2026-08-16
Add close buttons to Asset Manager and Python Console

francescofugazzi:
## Summary

This is a small quality-of-life update adding quick close buttons to the Asset Manager and Python Console.

- adds a compact close button to the Asset Manager toolbar
- adds the corresponding close action to the Python Console
- keeps the close controls available when the panels become narrower
- uses the existing panel theme colors and button states

These are intentionally small, localized additions that make both panels quicker to dismiss without changing their broader behavior or docking logic.


MrNeRF:
lgtm! Thanks!

PR #1638, opened 2026-08-16
Spark-Compatible Stream RAD Export

kmeirlaen:
## Summary

This patch updates the existing streamable RAD export path to follow Spark `build-lod --quality` behavior instead of adding a second RAD profile. Non-stream RAD keeps LichtFeld's existing optimized/bucketed path.

## Observed Symptoms

- Stream RAD exported by LichtFeld was Spark-readable, but did not match Spark `build-lod` structure.
- Large PLY inputs could take LichtFeld's bucketed out-of-core LoD path, producing a different global tree from Spark.
- Stream RAD metadata only guaranteed `lodOpacity`, while Spark also writes global encoding bounds such as `rgbMin`, `rgbMax`, `lnScaleMin`, `lnScaleMax`, and SH band maxima.
- CLI combinations like stream RAD plus tiling or octree LoD could request behavior that cannot match Spark.

## Replication Steps For Original Issue

1. Convert the same PLY to stream RAD through LichtFeld.
2. Convert the same PLY through Spark `build-lod --quality`.
3. Compare LoD node count, child ordering, chunk ranges, and `splatEncoding` metadata.
4. Before this patch, the LichtFeld result could differ because it used LichtFeld's 1.25 Bhatt base, bucketed conversion, BFS-oriented ordering, and sparse encoding metadata.

## Root Cause

The mismatch was not the RAD container. The container was readable by Spark. The difference came from the export pipeline feeding the container:

- Stream RAD could still route large PLY files through LichtFeld's bucketed converter.
- The default in-memory Bhatt LoD base was 1.25 instead of Spark's quality base of 1.75.
- LoD nodes were emitted in LichtFeld-oriented order rather than Spark chunk order.
- Encoding metadata was not resolved globally in the same shape Spark writes.

The issue showed up most clearly on larger scenes because bucket boundaries change the LoD hierarchy before encoding even begins.

## What This Patch Fixes

- Stream RAD now uses a global Bhatt LoD build with base 1.75.
- Stream RAD bypasses the bucketed PLY converter and existing fast RAD rechunk shortcut, so it can rebuild/reorder the global LoD.
- Stream RAD remaps LoD nodes into Spark-style chunks and updates child links against the remapped order.
- Stream RAD emits Spark-style top-level and per-chunk `splatEncoding` fields.
- Non-stream RAD keeps the existing LichtFeld optimized/bucketed behavior.
- CLI parsing rejects stream RAD combinations that cannot be Spark-compatible, such as `--tiles` and `--lod-builder octree`; these still work with `--none-stream`.
- Legacy RAD loading remains supported for files with only `{ "lodOpacity": true }`.
- The out-of-core LoD test now releases the returned sidecar-backed scene before removing its temp directory, matching the Windows file-handle lifetime of the deployed loader.

## Interface Impact

- GUI: no new UI text or new profile dropdown was added. Existing stream RAD selection now means Spark-compatible stream export.
- CLI: existing stream RAD defaults are preserved; invalid stream-only combinations now fail early with clear parser errors.
- Python API: existing RAD save options continue to use `chunk_size == 65536` for stream behavior; that path now produces Spark-compatible stream RAD.
- MCP: no schema change was needed. Existing scene export options flow through the same RAD save path.

### DIFF
--- a/src/app/converter.cpp
+++ b/src/app/converter.cpp
@@ -419,15 +419,17 @@ namespace lfs::app {
                     std::println(stderr, "  Error: {}", lod_chunk_size.error());
                     return false;
                 }
-                if (*lod_chunk_size) {
+                if (*lod_chunk_size && params.rad_export_mode == param::RadExportMode::NonStream) {
                     return rechunkRadFile(input, output, radChunkSizeForMode(params.rad_export_mode),
                                           make_convert_provenance(params.include_provenance, input));
                 }
             }
 
-            // An explicit non-default builder always takes the bucketed
-            // converter; the monolithic in-memory path is bhatt-only.
+            // The bucketed out-of-core converter is LichtFeld-native and is
+            // kept behind non-stream RAD. Stream RAD follows Spark build-lod:
+            // one global Bhatt tree, Spark chunk ordering, 65,536-node chunks.
             if (params.format == param::OutputFormat::RAD && isPlyExtension(input) &&
+                params.rad_export_mode == param::RadExportMode::NonStream &&
                 (tiled || params.lod_builder != param::LodBuilder::BHATT ||
                  shouldStreamLodConvert(input))) {
                 return streamLodConvertFile(input, output, params);
--- a/src/core/argument_parser.cpp
+++ b/src/core/argument_parser.cpp
@@ -1528,6 +1528,15 @@ namespace {
         }
         params.rad_export_mode = rad_none_stream ? param::RadExportMode::NonStream
                                                  : param::RadExportMode::Stream;
+        if (params.format == param::OutputFormat::RAD &&
+            params.rad_export_mode == param::RadExportMode::Stream) {
+            if (params.tiles_x > 1 || params.tiles_y > 1) {
+                return std::unexpected("--tiles requires --none-stream; streamable RAD uses Spark-compatible global LOD");
+            }
+            if (params.lod_builder == param::LodBuilder::OCTREE) {
+                return std::unexpected("--lod-builder octree requires --none-stream; streamable RAD uses Spark build-lod ordering");
+            }
+        }
 
         return core_args::ConvertMode{params};
     }
--- a/src/io/formats/rad.cpp
+++ b/src/io/formats/rad.cpp
@@ -2473,20 +2473,16 @@ namespace lfs::io {
             });
         }
 
-        float normalized_quat_component(const float* q, const int component) {
+        SparkAabb spark_splat_aabb(const float* center, const float* scale, const float* rotation) {
             double sum = 0.0;
             for (int i = 0; i < 4; ++i) {
-                sum += q[i] * q[i];
+                sum += rotation[i] * rotation[i];
             }
             const float inv = 1.0f / std::max(std::sqrt(static_cast<float>(sum)), 1.0e-12f);
-            return q[component] * inv;
-        }
-
-        SparkAabb spark_splat_aabb(const float* center, const float* scale, const float* rotation) {
-            const float x = normalized_quat_component(rotation, 0);
-            const float y = normalized_quat_component(rotation, 1);
-            const float z = normalized_quat_component(rotation, 2);
-            const float w = normalized_quat_component(rotation, 3);
+            const float x = rotation[0] * inv;
+            const float y = rotation[1] * inv;
+            const float z = rotation[2] * inv;
+            const float w = rotation[3] * inv;
             const glm::mat3 rmat = glm::mat3_cast(glm::quat(w, x, y, z));
             const glm::vec3 r{
                 std::max(scale[0], 1.0e-3f) * 1.5f,
@@ -3013,9 +3009,8 @@ namespace lfs::io {
                         }
                     }
                 }
-                std::vector<float> high = values;
                 const float low_value = percentile(values, 0.05f);
-                const float high_value = percentile(high, 0.95f);
+                const float high_value = percentile(values, 0.95f);
                 return std::max({std::abs(low_value), std::abs(high_value), 1.0f});
             }
 
@@ -3027,9 +3022,8 @@ namespace lfs::io {
                 }
 
                 std::vector<float> rgb(packed.sh0, packed.sh0 + packed.count * 3u);
-                std::vector<float> rgb_high = rgb;
                 const float rgb_min = std::min(percentile(rgb, 0.01f), 0.0f);
-                const float rgb_max = std::max(percentile(rgb_high, 0.99f), 1.0f);
+                const float rgb_max = std::max(percentile(rgb, 0.99f), 1.0f);
 
                 std::vector<float> scale_values;
                 scale_values.reserve(packed.count * 2u);
@@ -3043,9 +3037,8 @@ namespace lfs::io {
                     scale_values.push_back(splat_scales[1]);
                     scale_values.push_back(splat_scales[2]);
                 }
-                std::vector<float> scale_high = scale_values;
                 const float scale1 = percentile(scale_values, 0.01f);
-                const float scale99 = percentile(scale_high, 0.99f);
+                const float scale99 = percentile(scale_values, 0.99f);
                 const float ln_scale_min = std::min(std::log(std::max(scale1, 1.0e-30f)), -12.0f);
                 const float ln_scale_max = std::max(std::log(std::max(scale99, 1.0e-30f)), 9.0f);
 
--- a/src/io/formats/rad.cpp
+++ b/src/io/formats/rad.cpp
@@ -19,6 +19,7 @@
 
 #include <cuda_runtime.h>
 #include <libdeflate.h>
+#include <glm/gtc/quaternion.hpp>
 #include <nlohmann/json.hpp>
 #include <tbb/blocked_range.h>
 #include <tbb/enumerable_thread_specific.h>
@@ -49,6 +50,7 @@
 #include <mutex>
 #include <numeric>
 #include <optional>
+#include <queue>
 #include <string>
 #include <utility>
 #include <vector>
@@ -819,7 +821,9 @@ namespace lfs::io {
             float max_val;
         };
 
-        Ln0R8Result encode_ln_0r8(const float* data, size_t dims, size_t count) {
+        Ln0R8Result encode_ln_0r8(const float* data, size_t dims, size_t count,
+                                  std::optional<float> forced_min = std::nullopt,
+                                  std::optional<float> forced_max = std::nullopt) {
             // First pass: compute ln of all positive scales and find range
             std::vector<float> log_values;
             log_values.reserve(count * dims);
@@ -837,8 +841,10 @@ namespace lfs::io {
                 }
             }
 
-            // Handle edge case: all zeros or single value
-            if (!std::isfinite(ln_min) || !std::isfinite(ln_max) || ln_max - ln_min < 1e-7f) {
+            if (forced_min.has_value() && forced_max.has_value()) {
+                ln_min = forced_min.value();
+                ln_max = forced_max.value();
+            } else if (!std::isfinite(ln_min) || !std::isfinite(ln_max) || ln_max - ln_min < 1e-7f) {
                 ln_min = -10.0f; // Default ~exp(-10) = 4.5e-5
                 ln_max = 2.0f;   // Default exp(2) = 7.4
             }
@@ -1395,7 +1401,10 @@ namespace lfs::io {
                 return result;
             }
 
-            static EncodedProperty encode_rgb(const float* data, size_t dims, size_t count, RadRgbEncoding encoding) {
+            static EncodedProperty encode_rgb(const float* data, size_t dims, size_t count,
+                                              RadRgbEncoding encoding,
+                                              std::optional<float> forced_min = std::nullopt,
+                                              std::optional<float> forced_max = std::nullopt) {
                 EncodedProperty result;
 
                 switch (encoding) {
@@ -1412,7 +1421,7 @@ namespace lfs::io {
                     break;
 
                 case RadRgbEncoding::R8: {
-                    auto r8_result = encode_r8(data, dims, count);
+                    auto r8_result = encode_r8(data, dims, count, forced_min, forced_max);
                     result.data = std::move(r8_result.data);
                     result.min_val = r8_result.min_val;
                     result.max_val = r8_result.max_val;
@@ -1422,7 +1431,7 @@ namespace lfs::io {
                     break;
 
                 case RadRgbEncoding::R8Delta: {
-                    auto r8d_result = encode_r8_delta(data, dims, count);
+                    auto r8d_result = encode_r8_delta(data, dims, count, forced_min, forced_max);
                     result.data = std::move(r8d_result.data);
                     result.min_val = r8d_result.min_val;
                     result.max_val = r8d_result.max_val;
@@ -1432,7 +1441,7 @@ namespace lfs::io {
                     break;
 
                 default: {
-                    auto r8d_result = encode_r8_delta(data, dims, count);
+                    auto r8d_result = encode_r8_delta(data, dims, count, forced_min, forced_max);
                     result.data = std::move(r8d_result.data);
                     result.min_val = r8d_result.min_val;
                     result.max_val = r8d_result.max_val;
@@ -1445,7 +1454,10 @@ namespace lfs::io {
                 return result;
             }
 
-            static EncodedProperty encode_scales(const float* data, size_t dims, size_t count, RadScalesEncoding encoding) {
+            static EncodedProperty encode_scales(const float* data, size_t dims, size_t count,
+                                                 RadScalesEncoding encoding,
+                                                 std::optional<float> forced_min = std::nullopt,
+                                                 std::optional<float> forced_max = std::nullopt) {
                 EncodedProperty result;
 
                 switch (encoding) {
@@ -1456,7 +1468,7 @@ namespace lfs::io {
                     break;
 
                 case RadScalesEncoding::Ln0R8: {
-                    auto ln_result = encode_ln_0r8(data, dims, count);
+                    auto ln_result = encode_ln_0r8(data, dims, count, forced_min, forced_max);
                     result.data = std::move(ln_result.data);
                     result.min_val = ln_result.min_val;
                     result.max_val = ln_result.max_val;
@@ -1472,8 +1484,16 @@ namespace lfs::io {
                     break;
 
                 default:
-                    result.data = encode_ln_f16(data, dims, count);
-                    result.encoding = "ln_f16";
+                    if (forced_min.has_value() && forced_max.has_value()) {
+                        auto ln_result = encode_ln_0r8(data, dims, count, forced_min, forced_max);
+                        result.data = std::move(ln_result.data);
+                        result.min_val = ln_result.min_val;
+                        result.max_val = ln_result.max_val;
+                        result.encoding = "ln_0r8";
+                    } else {
+                        result.data = encode_ln_f16(data, dims, count);
+                        result.encoding = "ln_f16";
+                    }
                     result.compression = "none";
                     break;
                 }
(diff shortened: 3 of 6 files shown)


src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+            });
+        }
+
+        SparkAabb spark_splat_aabb(const float* center, const float* scale, const float* rotation) {
+            double sum = 0.0;
+            for (int i = 0; i < 4; ++i) {
+                sum += rotation[i] * rotation[i];
+            }
+            const float inv = 1.0f / std::max(std::sqrt(static_cast<float>(sum)), 1.0e-12f);
+            const float x = rotation[0] * inv;
```
MrNeRF, line 2482: **BUG**: `rotation_raw()` is wxyz but this reads xyzw: bhatt_lod.cpp:1031 writes `rotation_vec[i*4+0] = qw`, and this same file passes `input_wxyz=true` to `encode_quat_oct88r8` at line 2281 for the same tensor. The AABB is built from a scrambled quaternion, so every split plane and octant center derived from the oriented ellipsoids differs from build-lod. The output is still a valid tree (the AABB only steers batching), but chunk membership silently deviates. Measured on kerstbol (137,502 splats): fixing this alone moves nodes between chunks (per-chunk internals 22782/13906/11980 to 22782/13903/11983).

```cpp
const float w = rotation[0] * inv;
const float x = rotation[1] * inv;
const float y = rotation[2] * inv;
const float z = rotation[3] * inv;
```

src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+                }
+            };
+
+            for (std::size_t batch_index = 0; batch_index < batches.size(); ++batch_index) {
+                std::priority_queue<Candidate> queue;
+                for (const std::uint32_t old : batches[batch_index].nodes) {
+                    if (old >= node_count) {
+                        return spark_lod_error("LoD batch references a node outside the tree");
+                    }
+                    queue.push(Candidate{.size = tree.size_at(old), .node = old});
```
MrNeRF, line 2639: **Parity**: `tree.size_at()` is the renderer's LOD-cut metric (bhatt_lod.cpp:1111-1122: `2*max_scale*(1+0.7*(min(4a-3,5)-1))`), but Spark's `chunk_tree_size` pops by `feature_size() = 2*max_scale*sqrt(1+e*ln a)` (tsplat.rs:27-38), evaluated on raw opacities because build-lod runs `chunk_tree` before `encode_lod_opacity` (main.rs:240 vs :300). Both are monotonic in scale, but the opacity factor differs, so pop order among nodes with different accumulated opacity is not build-lod's.

All inputs are already in the exported SplatData, and `lod_merge_math.hpp:49` has the exact `lod_opacity` helper; computing the Spark key locally here (and at the second push below) is a few lines.

Measured: this fix plus the quaternion fix makes the reorder reproduce Spark's own `chunk_tree` permutation on an identical tree except one near-tie heap pop swap (8 link entries out of 237,833 nodes, parents 6066/6309). Without them, 237,357 of 237,833 positions differ.

src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+                        continue;
+                    }
+                    const std::uint32_t child_start = tree.child_start_at(old);
+                    if (static_cast<std::uint64_t>(child_start) + child_count > node_count) {
+                        return spark_lod_error("LoD child range exceeds node count");
+                    }
+                    if (child_count > kBatchSize) {
+                        return spark_lod_error("LoD node has more children than a Spark chunk can hold");
+                    }
+                    if (order.size() + child_count > end) {
```
MrNeRF, line 2662: **BUG, confirmed hang**: infinite loop with unbounded memory growth on real input. A batch whose top node has `child_count > end - order.size()` goes to `remaining`, `split_remaining` on a single node re-yields a batch containing exactly that node (the AABB center is its own center, so everything lands in one octant), and the next iteration recomputes the same start/end. `end - start` can be as small as kMinBatchSize = 8192 and the guard at 2659 only errors above kBatchSize = 65536, so any node with 8193..65536 children spins forever once `start` lands close enough to a 65536 boundary; every spin appends another Batch, so memory grows without bound.

Reproduced: `convert splat_30000.rad out.rad -f rad --stream` on a July-era LichtFeld RAD (1,011,868 nodes, one node with 32,768 children) never leaves "Preparing data...", RSS reaches 10 GiB in 45 s, stack sits in `reorder_lod_for_spark_chunks`. Spark's `chunk_tree_size` has the same latent no-progress retry (chunk_tree.rs:424-429, and no child-count guard at all), but it pops finished batches so it hangs without the memory blowup; an exporter hang is worse than an error either way.

Minimal fix: detect a no-progress batch (`order.size() == start && remaining.size() == batch.nodes.size()`) and return `spark_lod_error`. A real solution would let an over-wide node overflow its chunk.

src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+                                static_cast<std::size_t>(sh_coeffs) * 3u * sizeof(float));
+                }
+                out_tree->child_count[new_index] = remapped_child_count[old];
+                out_tree->child_start[new_index] = remapped_child_start[old];
+                out_tree->lod_level[new_index] = tree.level_at(old);
+                out_tree->centers[new_index] = tree.center_at(old);
+                out_tree->sizes[new_index] = tree.size_at(old);
+            }
+
+            SplatData output(sh_degree,
```
MrNeRF, line 2734: **Perf**: this constructs a full SplatData (the Canonical ctor allocates swizzled shN on CUDA and uploads, splat_data.cpp:664-669) and `pack_splat_data` immediately downloads and un-swizzles it again (`shN_canonical_cpu`). The reorder also fills `out_tree->lod_level/centers/sizes`, which the writer never reads; only `child_count`/`child_start` reach `pack_splat_data`.

Cheaper: pack first, then apply the permutation to `PackedSplatData` (host arrays already) plus the two link arrays. That removes six tensor allocations, one SplatData construction, one CUDA upload and download of the largest tensor, and the window where `lod_splat_data` and `spark_lod_splat_data` are both alive. At minimum, `lod_splat_data.reset()` after the move.

src/io/formats/rad.cpp
```diff
@@ -2627,6 +2982,76 @@ namespace lfs::io {
...
+                    }
+                }
+                const float low_value = percentile(values, 0.05f);
+                const float high_value = percentile(values, 0.95f);
+                return std::max({std::abs(low_value), std::abs(high_value), 1.0f});
+            }
+
+            static nlohmann::json make_lod_splat_encoding(const PackedSplatData& packed,
+                                                          const bool spark_compatible) {
+                nlohmann::json encoding{{"lodOpacity", true}};
```
MrNeRF, line 3019: **Parity**: Spark's build-lod remaps interior opacities into [1,2] before writing (`encode_lod_opacity`, tsplat.rs:83-92: `clamp(0.25*(sqrt(1+e*ln a)-1)+1, 1, 2)`); this path writes the raw accumulated values. On kerstbol the stream RAD has alpha max 49.72 at the root, 2,207 nodes > 1 and 231 nodes > 2, where Spark's own file tops out at 1.60. Every consumer assumes the [1,2] convention: Spark's shader stretches with `min(4a-3, 5)` (splatVertex.glsl:108), the JS u8 unpack is `2a` (utils.ts:766), and our own loader applies `4a-3` when rebuilding tree sizes (rad.cpp:3423). With raw values, all 231 nodes above 2 saturate the expansion cap.

Since this PR is the "stream RAD == Spark build-lod" PR, the remap belongs in the stream path (pack or reorder, ~5 lines). If it stays out, the PR body should not claim build-lod parity.

src/io/formats/rad.cpp
```diff
@@ -2627,6 +2982,76 @@ namespace lfs::io {
...
+                if (!spark_compatible || packed.count == 0) {
+                    return encoding;
+                }
+
+                std::vector<float> rgb(packed.sh0, packed.sh0 + packed.count * 3u);
+                const float rgb_min = std::min(percentile(rgb, 0.01f), 0.0f);
```
MrNeRF, line 3025: **Parity, low**: percentiles and clamps match Spark exactly, but Spark's Auto fallbacks are not mirrored:
- rgb: Spark switches to F16 and leaves rgbMin/rgbMax at defaults when `rgb_min < -1 || rgb_max > 2` (rad.rs:381). This always emits r8_delta; on kerstbol p99 = 2.0107 > 2, so Spark's file has rgb f16 with rgbMin 0 / rgbMax 1 while this writes r8_delta [-0.0622, 2.0107].
- scales: Spark switches to LnF16 when `ln_max - ln_min > 25` (rad.rs:414); this always forces ln_0r8.
- `encode_ln_0r8` zero threshold: this uses `ln_min - 1` (so -13 with the clamped -12); Spark passes `zero = -30` (rad.rs:684) and maps ln in (-30, -12] to code 1 = e^-12. Scales in (e^-30, e^-13] become 0 here vs e^-12 in Spark. Cosmetic, but easy to align.

src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+                    }
+                    std::ranges::sort(next_batches, [](const Batch& a, const Batch& b) {
+                        return a.nodes.size() > b.nodes.size();
+                    });
+                } else {
+                    std::array<std::vector<std::uint32_t>, 8> octants;
+                    for (const std::uint32_t old : remaining) {
+                        const std::size_t base = static_cast<std::size_t>(old) * 3u;
+                        const int octant =
+                            (means[base + 0u] > center.x ? 1 : 0) |
```
MrNeRF, line 2615: **Parity, low**: two exact-tie divergences from Spark, both trivial to align:
- Octant bit: `means > center ? 1 : 0` gives bit 0 on exact equality; Spark's `center < split ? 0 : 1` (chunk_tree.rs:471) gives 1.
- The two-element sort above uses `std::ranges::sort`; Spark's `sort_by_key` is stable (a before b on equal size). libstdc++ happens not to reorder equal two-element ranges, but `std::stable_sort` makes it portable.

src/io/formats/rad.cpp
```diff
@@ -2408,6 +2450,300 @@ namespace lfs::io {
...
+            constexpr std::uint32_t kBatchSize = SPARK_CHUNK_SIZE;
+            constexpr std::uint32_t kMinBatchSize = 8u * 1024u;
+            constexpr std::array<int, 8> kOctantOrder = {0, 1, 3, 2, 6, 7, 5, 4};
+            constexpr std::uint32_t kInvalid = std::numeric_limits<std::uint32_t>::max();
+
+            std::vector<std::uint32_t> order;
+            order.reserve(node_count);
+            order.push_back(0);
+
+            std::vector<std::uint32_t> old_to_new(node_count, kInvalid);
```
MrNeRF, line 2563: Minor: `old_to_new` is only ever read as a visited flag (`!= kInvalid` at 2673); a `std::vector<uint8_t>` does the job with 4x less memory. And `Batch` wraps a single vector; `std::vector<std::vector<uint32_t>>` reads the same.

src/io/formats/rad.cpp
```diff
@@ -819,7 +821,9 @@ namespace lfs::io {
             float max_val;
         };
 
-        Ln0R8Result encode_ln_0r8(const float* data, size_t dims, size_t count) {
+        Ln0R8Result encode_ln_0r8(const float* data, size_t dims, size_t count,
+                                  std::optional<float> forced_min = std::nullopt,
+                                  std::optional<float> forced_max = std::nullopt) {
             // First pass: compute ln of all positive scales and find range
             std::vector<float> log_values;
```
MrNeRF, line 828: Minor: `log_values` is filled and never read (pre-existing; the encode loop below re-logs from `data`), and with both bounds forced the min/max scan result is discarded too. Delete the vector and skip the range scan when bounds are forced; this is a per-conversion pass over count*dims elements.

src/io/formats/rad.cpp
```diff
@@ -2451,13 +2789,27 @@ namespace lfs::io {
                     auto lod_progress = [&](float p, const std::string& stage) -> bool {
                         return report_progress(p * 0.1f, stage);
                     };
-                    auto lod_result = lfs::core::build_bhatt_lod(*export_source, 1.25f, lod_progress);
+                    auto lod_result = lfs::core::build_bhatt_lod(*export_source,
+                                                                 spark_compatible ? 1.75f : 1.25f,
```
MrNeRF, line 2793: **Behavior**: converting an existing RAD with `--stream` keeps whatever tree the input had: convertFile skips rechunkRadFile for stream, falls through to a full load, and this branch only builds when the source has no tree. Verified: a 1.25-base LichtFeld RAD converts to 237,833 nodes in Spark order with Spark encoding, not a 1.75 rebuild. Either rebuild from leaves for stream mode or document that stream conversion of an existing LoD RAD is a re-encode, not a rebuild.

src/app/converter.cpp
```diff
@@ -419,15 +419,17 @@ namespace lfs::app {
...
                 }
             }
 
-            // An explicit non-default builder always takes the bucketed
-            // converter; the monolithic in-memory path is bhatt-only.
+            // The bucketed out-of-core converter is LichtFeld-native and is
+            // kept behind non-stream RAD. Stream RAD follows Spark build-lod:
+            // one global Bhatt tree, Spark chunk ordering, 65,536-node chunks.
             if (params.format == param::OutputFormat::RAD && isPlyExtension(input) &&
+                params.rad_export_mode == param::RadExportMode::NonStream &&
```
MrNeRF, line 432: **Design, needs a decision**: this drops the out-of-core path for every stream RAD conversion. On master, any PLY over `target_bucket_splats` = 2,000,000 (or an in-core estimate above 70% of MemAvailable) went through the bucketed converter; stream is the default export mode for CLI, GUI, and Python, so the default RAD export of a 50M-splat PLY now means full PLY load + `build_bhatt_lod` workset + CUDA upload + reorder copy + pack, all in-core. Spark's build-lod is in-core too, but matching Spark should not cost the ability to convert scenes Spark cannot.

Options: (a) keep the bucketed path as fallback when `shouldStreamLodConvert` fires and log that the order is bucket-local; (b) teach the bucketed converter the Spark encoding block + 1.75 base and accept non-global order; (c) accept the regression and state it in the PR body and CLI help. The body currently says only "bypasses the bucketed PLY converter" without the size implication.

src/core/argument_parser.cpp
```diff
@@ -1528,6 +1528,15 @@ namespace {
         }
         params.rad_export_mode = rad_none_stream ? param::RadExportMode::NonStream
                                                  : param::RadExportMode::Stream;
+        if (params.format == param::OutputFormat::RAD &&
```
MrNeRF, line 1531: **Docs**: the rejections are right. One sentence is still owed here and in the PR body: with stream as the default, the default RAD export changes fidelity. 1.75 LoD base (186,170 vs 237,833 nodes on kerstbol), scales from per-chunk ln_f16 to 8-bit ln_0r8 over a global range of at least [-12, 9] (about 4% max relative scale error), SH from per-chunk max-abs to a global 5%/95% clamp. Right for Spark parity, but it belongs in --help, e.g. "--stream ... Spark build-lod compatible: 1.75 LoD base, 8-bit global scales".

tests/test_argument_parser.cpp
```diff
@@ -482,6 +482,91 @@ TEST(ArgumentParserTest, ConvertNoProvenanceDisablesStamp) {
...
+        "rad",
+        "--lod-builder",
+        "octree"};
+
+    auto parsed = lfs::core::args::parse_args(static_cast<int>(std::size(argv)), argv);
+    ASSERT_FALSE(parsed.has_value());
+    EXPECT_NE(parsed.error().find("--lod-builder octree requires --none-stream"), std::string::npos);
+}
+
+TEST(ArgumentParserTest, ConvertRejectsRemovedRadProfileFlag) {
```
MrNeRF, line 551: This asserts that a flag which never existed on master is rejected; every unknown flag fails parse, so the test is a tautology. Delete.

tests/test_python_io.cpp
```diff
@@ -1523,6 +1523,102 @@ TEST_F(PythonIOTest, RadSaveLoadRoundtripPreservesCompactSplat) {
     EXPECT_EQ(leaf_positions, expected_positions);
 }
 
+TEST_F(PythonIOTest, RadStreamSaveRealPlyUsesSparkSizedForwardLodChunks) {
+    const fs::path source_path = data_dir / "kerstbol-isolated-rotated_137502.ply";
+    if (!fs::exists(source_path)) {
```
MrNeRF, line 1528: CI skips this (the PLY is untracked), so nothing in CI asserts Spark ordering; the 64-splat roundtrip above does run through `reorder_lod_for_spark_chunks` but checks none of it. Worth adding chunk-order assertions on the synthetic so the reorder has one CI-run correctness check.

tests/test_ply_to_rad_lod.cpp
```diff
@@ -331,87 +331,89 @@ TEST(PlyToRadLod, OutOfCoreLoadKeepsTreeAndStreamsChunks) {
...
-        EXPECT_EQ(partial->lod_tree->child_count_at(i), full->lod_tree->child_count_at(i));
-        EXPECT_EQ(partial->lod_tree->level_at(i), full->lod_tree->level_at(i));
-        if (HasFailure()) {
-            FAIL() << "tree mismatch at node " << i;
-        }
-    }
-
-    // Parent links in the sidecar must agree with a reference forward scatter
-    // over the full in-RAM tree.
     {
```
MrNeRF, line 334: The scoped-lifetime fix is correct, but the 154-line re-indent that comes with it is unrelated to the PR subject; better as its own commit.

MrNeRF (changes requested):
Reviewed against the Spark reference (rust/spark-lib `chunk_tree.rs` / `rad.rs` / `tsplat.rs`, build-lod). Built the branch, ran the touched gtests (18 pass, 2 pre-existing skips), converted a 137,502-splat PLY with `--stream` and `--none-stream`, ran Spark's `build-lod --quality` on the same PLY, and ran Spark's own `chunk_tree` over the identical LichtFeld tree, then diffed the resulting trees node by node.

Direction is right and the shape is what was asked for: the `chunk_tree_size` port is structurally faithful (chunk boundary math, heap tie order, 3x binary-vs-octant rule, octant visit order, 1.5-sigma oriented AABB all match), and the metadata percentiles and clamps mirror Spark exactly. But two defects make the emitted order not Spark's, the LoD opacity is written in a form no Spark consumer expects, one design decision drops the out-of-core path for every stream conversion, and a real LichtFeld RAD sends the reorder into an infinite loop. Details inline; the headline numbers, reordering an identical 237,833-node tree:

| variant | positions differing from Spark `chunk_tree` |
|---|---|
| PR as-is | 237,357 |
| + quaternion fix | 237,289 |
| + quaternion fix + Spark `feature_size` key | 8 link entries (one near-tie heap swap) |

So the two inline fixes take this from unrelated order to bit-equal with Spark up to f16 key precision.

Also verified: infinite loop reproduced on a July-era LichtFeld RAD (RSS 10 GiB in 45 s, never leaves "Preparing data..."); stream alpha max 49.72 with 231 nodes > 2 where Spark's own file tops out at 1.60; on this scene Spark's Auto picks rgb f16 (p99 = 2.0107 > 2) where this PR forces r8_delta.

One caveat for the replication steps in the body: LichtFeld's `build_bhatt_lod` at 1.75 is not bit-identical to Spark's `bhatt_lod` (level_min -26 vs -19, 186,170 vs 181,676 nodes, root fan-out 5 vs 6). That is a pre-existing bhatt_lod.cpp property, not this PR's fault, but "compare LoD node count" cannot pass from a PLY today and the body should not imply it does.

kmeirlaen:
will go back to the original implementation/PR

PR #1646, opened 2026-08-17
Reset point cloud mode when loading standalone assets

francescofugazzi:
## Summary

- Reset point cloud rendering mode after successfully loading a standalone asset
- Prevent the dataset-specific point rendering mode from carrying over to subsequently loaded splats or meshes
- Apply the reset only after a successful load
- Add regression coverage for both splat and mesh loading paths, including the exact mode event sequence and voxel size

## Testing

- Manually verified by loading a dataset followed by a Gaussian PLY
- Confirmed that the PLY switches back to splat rendering instead of remaining in point cloud mode

### DIFF
--- a/src/visualizer/scene/scene_manager.cpp
+++ b/src/visualizer/scene/scene_manager.cpp
@@ -891,14 +891,6 @@ namespace lfs::vis {
                     updateCropBoxToFitScene(true);
                 }
 
-                // Dataset loading enables point-cloud rendering by default. A later
-                // standalone splat load replaces that dataset, so do not carry the
-                // dataset-specific render mode into the new scene.
-                ui::PointCloudModeChanged{
-                    .enabled = false,
-                    .voxel_size = DEFAULT_VOXEL_SIZE}
-                    .emit();
-
                 selectNode(node_id);
 
                 // Check for companion PPISP file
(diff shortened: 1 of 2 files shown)


src/visualizer/scene/scene_manager.cpp
```diff
@@ -891,6 +891,14 @@ namespace lfs::vis {
                     updateCropBoxToFitScene(true);
                 }
 
+                // Dataset loading enables point-cloud rendering by default. A later
+                // standalone splat load replaces that dataset, so do not carry the
+                // dataset-specific render mode into the new scene.
+                ui::PointCloudModeChanged{
+                    .enabled = false,
+                    .voxel_size = DEFAULT_VOXEL_SIZE}
+                    .emit();
```
MrNeRF: The mesh branch of this same function does the same clear-and-replace but skips this reset. So dataset → import mesh → add PLY still shows the splat as points. Worth the same emit there — or doing the reset in `handleSceneCleared` so every scene clear is covered. Fine as a follow-up.

Small note: this also resets `voxel_size` to 0.01. Same as the checkpoint path, so pre-existing behavior.

tests/test_scene_validity.cpp
```diff
@@ -1233,6 +1233,36 @@ namespace lfs::python {
...
+            [&](const auto& event) { mode_events.push_back(event); });
+        core::events::ui::PointCloudModeChanged{
+            .enabled = true,
+            .voxel_size = 0.01f}
+            .emit();
+
+        lfs::vis::SceneManager scene_manager;
+        scene_manager.loadSplatFile(splat_path);
+
+        ASSERT_GE(mode_events.size(), 2u);
```
MrNeRF: Could be pinned tighter: `ASSERT_EQ(mode_events.size(), 2u)` and check both events (first enabled, second disabled, plus voxel_size). `GE` + `back()` would still pass if extra events sneak in.

MrNeRF (commented):
Reviewed and tested locally: cherry-picked onto master, built, and ran the tests. The new test passes with the fix and fails without it, so the coverage is real. Full SceneValidityTest suite is green (50/50). Project restore (#1635) is not affected — it never calls `loadSplatFile`.

Fix looks correct and matches the existing checkpoint-load pattern. Two small points inline.

francescofugazzi:
Followed up on both review notes:

- Moved the point cloud mode reset to the common successful-load path, covering both splats and meshes without changing state after a failed load
- Tightened the splat regression test to verify the exact event sequence and voxel sizes
- Added regression coverage for the mesh replacement path
- Kept the reset voxel size at the existing `0.01` default, consistent with checkpoint loading

PR #1658, opened 2026-08-17
Add persistent MCP server management and privacy-safe diagnostics

francescofugazzi:
## Summary

This PR adds persistent, live MCP server management to LichtFeld Studio. MCP configuration can now be inspected and changed from Preferences or the status bar without restarting the application.

It also introduces privacy-bounded request diagnostics, accurate runtime lifecycle and error reporting, local-network endpoint discovery, remappable MCP shortcuts, safe-mode integration, and comprehensive contract coverage.

The server remains enabled by default on the loopback interface to preserve the existing local MCP workflow. Network exposure and request logging remain explicit opt-in settings.

## Persistent MCP preferences

MCP configuration is stored in the user `preferences.json` file:

- server enabled state;
- loopback-only or local-network binding;
- listening port;
- request logging state.

Preference values are validated before being applied. Invalid ports fall back to the default port.

Safe mode remains process-local:

- it does not load saved MCP preferences;
- it does not modify them;
- it keeps the MCP listener and request logging disabled for that process;
- it exposes the effective safe-mode state to the UI;
- it keeps the MCP Preferences section available for diagnosis while disabling all live configuration controls.

Resetting the MCP section or all preferences restores the documented defaults.

## Runtime server management

The MCP HTTP server can apply configuration changes while the application is running.

Runtime changes are handled by one long-lived background worker. Rapid UI changes are staged and deliberately coalesced so obsolete intermediate configurations are not restarted.

The runtime management layer provides:

- serialized start, stop, and reconfiguration;
- staged configuration while a listener is being restarted;
- deterministic cancellation and listener shutdown when the active server is detached;
- loopback binding through `127.0.0.1`;
- optional local-network binding through `0.0.0.0`;
- single-process ownership of the configured address and port;
- truthful enabled, running, bind-scope, port, logging, endpoint, and error state;
- explicit Disabled, Starting, Running, Stopping, and Failed lifecycle phases;
- request success and error counters;
- preservation of bind and listener failures until successful recovery;
- retrying a previously failed configuration without requiring a different port;
- prompt runtime-state propagation to the UI.

The listener lifecycle is guarded against startup and shutdown races, exceptions during listener initialization, accept-loop failures, stale socket descriptors, and listener death between readiness and status publication.

An occupied MCP address and port now produce a real bind failure instead of allowing multiple processes to report Listening through socket reuse. Windows uses exclusive address ownership, while POSIX preserves clean restart behavior without enabling `SO_REUSEPORT`.

The configuration worker has an exception barrier so configuration, allocation, endpoint discovery, or logging failures report a Failed state without terminating the process or silently killing the worker.

Status polling and configuration staging do not hold the global MCP server mutex while performing adapter discovery, filesystem access, log appends, or native status queries. Python MCP preference and status bindings also release the GIL while performing native work.

Local-network endpoint discovery is implemented for Windows and Linux. Loopback, unspecified, and link-local addresses are excluded from the client endpoint list. Active VPN and virtual adapters may appear because they are valid interfaces through which an MCP client can connect.

## Preferences interface

A dedicated MCP section has been added to Preferences.

It provides:

- an MCP server enable switch;
- a local-network exposure option;
- a warning that network clients can invoke MCP tools without authentication;
- an editable port with explicit Enter or OK confirmation;
- an opt-in request logging toggle;
- current runtime lifecycle status;
- active client endpoints;
- structured bind and listener failures;
- the current request log path;
- an action for opening the log directory;
- a safe-mode notice when MCP controls are unavailable.

Editing the port does not restart the listener on every keystroke. The value remains a draft until it is confirmed with Enter, the adjacent confirmation button, or the Preferences footer OK button.

Closing Preferences through the title-bar close button discards an unconfirmed port draft. The close button is connected to the specialized Preferences close handler rather than only hiding the panel.

External runtime changes, such as status-bar controls or MCP shortcuts, update the applied runtime state without overwriting a port draft that is still being edited.

Confirming a port that is already applied retries the listener only when its lifecycle phase is Failed. A listener that is still Starting is not unnecessarily restarted.

Bind errors identify the affected address and port, remain visible until the problem is resolved, and do not advertise an inactive endpoint. Successfully retrying the listener clears the stale error.

The Preferences content area is vertically scrollable. Reducing the window height keeps the footer fixed while preventing MCP status, endpoint, and error content from overlapping its controls.

The Preferences panel can also be opened directly on the MCP section from the status bar.

## Status bar integration

The MCP status chip reflects the effective runtime state:

- **Off**
- **Starting**
- **Local**
- **Network**
- **Stopping**
- **Error**

When the server is disabled, the chip still communicates whether its pending bind scope is loopback-only or local-network, so changing the scope does not silently hide the configuration that will be used on the next enable.

The expanded view includes:

- listener lifecycle status;
- active endpoints;
- detailed bind address and port failures;
- request success and error counters;
- a direct server power control;
- an action to open the MCP Preferences section.

The popup measures its content dynamically, including localized strings, multiple endpoints, and detailed errors.

Its measured rectangle is registered with the shared RmlUI overlay hit-test system using window-client coordinates. Only the actual popup area reserves input, and pointer interaction cannot pass through the popup to rotate or manipulate the viewport underneath.

The same coordinate contract is used by live and cached rendering paths and is covered with a non-zero window-origin regression test.

Localized runtime status formatting is guarded against malformed translation format strings. A translation error falls back to the untranslated localized string rather than throwing from the render loop.

Safe mode disables the status-bar power control and blocks MCP shortcut mutations while continuing to expose the effective disabled state.

## Runtime shortcuts and keymap integration

Two remappable MCP actions have been added:

- `Ctrl+Shift+M` — toggle the MCP server;
- `Ctrl+Shift+N` — switch between loopback and local-network binding.

Changing the bind scope does not implicitly enable a disabled server. The pending scope remains visible through the status-bar state.

The actions are exposed through the visualizer runtime boundary, Python keymap API, generated stubs, Input Settings, and localized action labels.

The keymap profile format is migrated to version 22. Existing profiles receive the new default bindings without losing user customization.

Default binding descriptions use the canonical `getActionName()` values so profile migration remains stable if action enum values are changed in the future.

MCP runtime shortcuts remain available during Python keyboard capture, while unrelated application shortcuts retain their existing capture behavior.

Safe mode blocks both MCP runtime shortcuts without changing the persisted configuration.

## Privacy-bounded request logging

MCP request logging is disabled by default and must be explicitly enabled.

When enabled, LichtFeld Studio creates a per-session JSONL log under the MCP user log directory. Log files are created lazily, appended under synchronization without rewriting earlier records, and confined to that directory.

Log state and serialized disk-write synchronization are kept separate. Runtime status queries use the cached active log path and do not wait for filesystem resolution or an in-progress append.

Each record contains operational metadata such as:

- event and JSON-RPC method;
- request identifier;
- outcome and duration;
- source IP address and port;
- destination IP address and port;
- structured error classification.

The log deliberately excludes:

- request parameters;
- tool arguments;
- tool names;
- response payloads;
- free-form error messages;
- internal diagnostic details.

Network addresses are included to support connection diagnostics and may identify devices and interfaces on the local network. This is documented alongside the explicit opt-in behavior.

Safe mode forces effective request logging off without modifying the persisted logging preference.

## Error classification and accounting

Request diagnostics distinguish between transport, JSON-RPC, and MCP tool-execution outcomes.

### JSON-RPC errors

Protocol-level failures are recorded with:

- error type `json_rpc`;
- parse, validation, dispatch, execution, or protocol stage;
- a stable reason;
- the numeric JSON-RPC error code.

### MCP tool execution errors

A `tools/call` response may be a valid JSON-RPC result envelope while still reporting `isError: true`.

These records are classified separately with:

- error type `tool_execution`;
- stage `tool_call`;
- typed application error code and domain when available;
- retryability;
- operation identifier.

Tool failures using `result.isError` are counted as errors rather than successful requests.

### Listener and configuration errors

Runtime listener failures use structured lifecycle information rather than matching English error strings.

The status boundary exposes:

- lifecycle phase;
- stable error kind;
- attempted bind address;
- attempted port;
- effective running state.

This allows Preferences and the status bar to render accurate failures without depending on the wording of diagnostic messages.

## Additional fixes

While implementing and validating MCP management, the following related issues were also corrected:

- close the keymap profile stream before atomic migration replacement, avoiding Windows file-sharing failures;
- preserve detailed listener errors and allow a clean retry using the same configuration;
- clear stale listener errors after successful recovery;
- prevent a listener failure from being overwritten by a stale Running publication;
- prevent readiness waits from spinning indefinitely after a listener exception;
- avoid closing a stale socket descriptor after an accept-loop failure;
- keep the long-lived configuration worker alive after recoverable exceptions;
- make detach explicitly stop the outgoing listener after pending work is cancelled;
- prevent MCP status popups from passing pointer input through to the viewport;
- correct status-overlay coordinate handling for non-zero desktop window positions;
- export the measured status-overlay helpers required by the test target;
- add the required Windows IP Helper and socket libraries for network endpoint discovery;
- correct request counters for tool failures returned through successful JSON-RPC envelopes;
- reject concurrent ownership of an occupied MCP port instead of silently sharing it through socket reuse;
- preserve unconfirmed port drafts during external runtime changes;
- discard drafts through the actual Preferences title-bar close path;
- keep safe-mode UI state synchronized with the effective disabled listener;
- prevent short Preferences windows from overlapping MCP content with the fixed footer.

## Localization

All new user-facing MCP controls, lifecycle labels, safe-mode notices, warnings, errors, endpoint descriptions, shortcuts, and logging information are localized in:

- German;
- English;
- Spanish;
- French;
- Italian;
- Japanese;
- Korean;
- Dutch;
- Polish;
- Chinese.

## Documentation

The development documentation covers:

- persistent MCP preferences;
- default loopback behavior;
- optional local-network exposure;
- the lack of MCP network authentication;
- runtime lifecycle and controls;
- single-process listener ownership;
- bind troubleshooting and recovery;
- endpoint discovery;
- request log location and append-only lifecycle;
- logged metadata and privacy exclusions;
- status-bar integration and shortcuts;
- port draft confirmation and close behavior;
- safe-mode behavior and disabled controls.

## Tests

The changes add or extend contract coverage for:

- MCP preference persistence and validation;
- safe-mode isolation and disabled runtime controls;
- preference resets;
- runtime server configuration and truthful lifecycle status;
- rapid enable/disable coalescing;
- deterministic detach and listener shutdown;
- listener exception and readiness handling;
- bind failure and same-configuration retry behavior;
- concurrent listener ownership rejection;
- ephemeral test-port allocation;
- endpoint discovery and wildcard-address exclusion;
- status-bar content sizing and hit testing;
- non-zero window-origin overlay coordinate handling;
- responsive Preferences scrolling and fixed-footer separation;
- title-bar port-draft discard through the real event listener;
- draft preservation during external runtime toggles;
- failed-only listener retry behavior;
- runtime actions and visualizer boundary dispatch;
- keymap version 22 migration;
- canonical default binding descriptions;
- remappable shortcuts and Python capture behavior;
- Windows-safe atomic keymap migration;
- Python GIL release around native MCP operations;
- lazy, append-only, and confined MCP log creation;
- source and destination socket metadata;
- JSON-RPC error classification;
- MCP tool-execution error classification;
- request counter accounting;
- payload and parameter privacy;
- localization and Python UI contracts.


src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +492,96 @@ namespace lfs::mcp {
                                                                lfs::core::ReportChannel::OwnerLog);
                     }
                 });
+            running_.store(false, std::memory_order_release);
         });
 
         return true;
     }
 
     void McpHttpServer::stop() {
```
MrNeRF, line 675: This can freeze the whole app. `stop()` holds the lock while waiting for the listener thread, and the status bar polls `status()` with the same lock from the render thread. If a tool call is still running, it waits for the render thread — and now everyone waits for everyone. One toggle while an agent is mid-request is enough. Please make `status()` read a snapshot with its own small lock, and don't join threads while holding this one.

src/mcp/mcp_http_server.cpp
```diff
@@ -83,21 +328,153 @@ namespace lfs::mcp {
...
+            {"event", "state"},
+            {"state", "started"},
+            {"address", bind_address},
+            {"port", config.port},
+        });
+        const auto listener_url = std::format("http://{}:{}/mcp", bind_address, config.port);
+        const bool announce_listener = listener_url != last_announced_listener_url_;
+        if (announce_listener)
+            last_announced_listener_url_ = listener_url;
+        listener_thread_ = std::jthread([this, listener_url, announce_listener](
```
MrNeRF: There's a race: httplib's `stop()` does nothing if the listener hasn't fully started yet. So a quick enable→disable can leave the socket open and the join hangs forever. Call `wait_until_ready()` before stopping.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +492,96 @@ namespace lfs::mcp {
...
+        g_active_server = server;
+    }
+
+    McpHttpStatus activeMcpHttpStatus() {
+        std::lock_guard lock(g_active_server_mutex);
+        return g_active_server ? g_active_server->status()
+                               : McpHttpStatus{.enabled = false};
+    }
+
+    bool applyActiveMcpHttpConfig(const McpHttpConfig& config) {
```
MrNeRF, line 873: This joins the previous config thread on the UI thread. Same freeze risk as above. Better: one long-lived worker thread with a queue, so the UI never waits.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +492,96 @@ namespace lfs::mcp {
...
         }
-        if (listener_thread_.joinable()) {
+        if (http_server_)
+            http_server_->stop();
+        if (listener_thread_.joinable())
             listener_thread_.join();
+        running_.store(false, std::memory_order_release);
+    }
+
+    bool McpHttpServer::applyConfig(const McpHttpConfig& config) {
```
MrNeRF, line 704: If the listener dies but the config didn't change, the fast path returns early and "retry" does nothing. Also note: after a failed bind, httplib refuses new binds until `stop()` is called — a plain `start()` retry can never work.

src/mcp/mcp_http_server.cpp
```diff
@@ -83,21 +328,153 @@ namespace lfs::mcp {
...
+                log_event["request_id"] = *id;
+            appendTransportMetadata(log_event, req);
+            if (request_failed)
+                appendResponseErrorMetadata(log_event, *rpc_req, rpc_resp);
+            appendSessionLog(log_event);
             res.set_content(serialize_response(rpc_resp), "application/json");
         });
+    }
+
+    void McpHttpServer::appendSessionLog(const nlohmann::json& event) {
```
MrNeRF, line 421: This keeps the whole log in memory and rewrites the entire file on every request. Long session with logging on = gigabytes of disk writes. Just append to the file instead.

src/visualizer/input/input_bindings.cpp
```diff
@@ -1021,6 +1024,8 @@ namespace lfs::vis::input {
             {KeyTrigger{KEY_F11, MODIFIER_NONE}, Action::TOGGLE_FULLSCREEN, "Fullscreen"},
             {KeyTrigger{KEY_F10, MODIFIER_NONE}, Action::TOGGLE_PERFORMANCE_HUD, "Performance HUD"},
             {KeyTrigger{KEY_COMMA, MODIFIER_CTRL}, Action::OPEN_PREFERENCES, "Preferences"},
+            {KeyTrigger{KEY_M, MODIFIER_CTRL | MODIFIER_SHIFT}, Action::TOGGLE_MCP_SERVER, "Toggle MCP server"},
```
MrNeRF: The new test `DefaultProfileMigrationClosesSourceBeforeAtomicPersistence` fails: this table says "Toggle MCP server" / "Toggle MCP binding" but the test expects the display names. Please make these match `getActionName()` — that also matters because profile migration matches bindings by description, and "Toggle MCP binding" would never match on a future enum reshuffle.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -5620,12 +5641,15 @@ namespace lfs::vis::gui {
             const float status_bar_x = screen.work_pos.x;
             const float status_bar_y = screen.work_pos.y + screen.work_size.y;
             const float status_bar_w = screen.work_size.x;
+            const float status_local_x = panel_input.mouse_x - status_bar_x;
+            const float status_local_y = panel_input.mouse_y - status_bar_y;
+            const bool inside_status_overlay =
```
MrNeRF, line 5646: Input goes *into* the popup fine, but the viewport doesn't know the popup is there — clicking the power switch also rotates the camera underneath. The popup needs to register itself in the hit test (like the other overlays do via `trackContextFrame`).

src/visualizer/gui/rml_status_bar.cpp
```diff
@@ -932,6 +994,54 @@ namespace lfs::vis::gui {
...
+            std::string summary;
+            std::string details;
+            std::string tooltip;
+            std::string color;
+            if (!status.enabled) {
+                summary = LOC("status_bar.mcp_off");
+                tooltip = LOC("status_bar.mcp_disabled_detail");
+                color = colorToRml(p.text_dim);
+            } else if (!status.running) {
+                summary = LOC("status_bar.mcp_error");
```
MrNeRF, line 1038: Every restart briefly shows "Error" (enabled but not running yet). Maybe add a "starting…" state. Also the popup shows a generic error text — the useful "Unable to bind X:Y" only shows in Preferences.

src/app/application.cpp
```diff
@@ -1190,15 +1209,60 @@ namespace lfs::app {
...
+                    if (!mcp::applyActiveMcpHttpConfig(config))
+                        return false;
+                    vis::saveMcpPreferences({
+                        .enabled = config.enabled,
+                        .expose_network = config.expose_network,
+                        .port = config.port,
+                        .request_logging = config.request_logging,
+                    });
+                    return true; },
+                .toggle_mcp_binding = [] {
```
MrNeRF: Ctrl+Shift+N with the server off silently saves `expose_network=true` and the UI shows nothing. Next time someone turns the server on it binds to the network without warning. Please add some feedback (toast) or show the pending scope in the Off chip.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -249,15 +283,192 @@ def _set_remember_view_snap(self, enabled):
...
+    def _on_mcp_port_change(self, _handle, event, args):
+        if args:
+            self._set_mcp_port(args[0])
+        if event.get_bool_parameter("linebreak", False):
+            self._commit_mcp_port()
+
+    def _on_confirm_mcp_port(self, _handle, _event, _args):
+        self._commit_mcp_port()
+
+    def _commit_mcp_port(self):
```
MrNeRF, line 340: Closing the panel with the title-bar ✕ keeps the typed port around; a later OK applies it. Please clear the draft on close. Also, the bind error is detected by matching the English string "Unable to bind " from C++ — that breaks the moment the message changes. A structured field would be safer.

MrNeRF (changes requested):
Nice feature overall — prefs, safe mode, logging privacy, migration, and locales are solid and well tested. But I can't merge yet: there's a deadlock risk in the start/stop code and one of the new tests fails on this branch (CI doesn't run the C++ tests, so it went unnoticed). Details inline.

francescofugazzi:
Fixed all requested MCP review items.

- Separated lifecycle coordination from status snapshot locking. Status reads now use their own small lock, and listener joins no longer occur while holding the lifecycle mutex.
- Replaced per-change configuration threads with one long-lived background worker. UI changes are staged and return immediately; rapid pending changes are deliberately coalesced so obsolete intermediate configurations are not restarted.
- Made listener shutdown readiness-safe by waiting for `cpp-httplib` to enter its accept loop or report failure before calling `stop()`.
- Closed the enable/disable startup race and added coverage ensuring a rapid queued disable cannot leave a listener running.
- Made unchanged-configuration retries check listener health instead of returning early. Failed binds now reset `cpp-httplib` before retrying, and the same address and port can recover without requiring a different value.
- Split structured lifecycle state into Disabled, Starting, Running, Stopping, and Failed phases, with typed invalid-port, bind, and listener failures.
- Updated the status-bar chip and popup to show startup, shutdown, pending disabled bind scope, and the concrete address and port involved in bind failures.
- Registered the measured MCP popup rectangle as an active overlay so controls no longer pass pointer input through to the viewport.
- Made the Preferences title-bar close action discard an unconfirmed port draft, while the footer OK action explicitly validates and applies it.
- Removed English error-string matching from Preferences and exposed structured error kind, address, and port fields through the Python UI boundary.
- Changed MCP JSONL persistence from whole-file atomic rewrites to synchronized append-only writes, avoiding growing memory use and repeated rewriting of long session logs.
- Updated the default keymap table to use the canonical `getActionName()` descriptions required by migration and future enum remapping.
- Added deterministic bind-failure coverage. The test-only port holder uses exclusive binding on Windows and non-reusable binding on POSIX; production socket policy is unchanged.
- Added and extended lifecycle, retry, logging, status-overlay, Preferences, keymap migration, localization, and runtime-action contracts.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +540,228 @@ namespace lfs::mcp {
...
+        }
+        return result;
+    }
+
+    void setActiveMcpHttpServer(McpHttpServer* const server) {
+        std::jthread worker_to_join;
+        {
+            std::lock_guard lock(g_active_server_mutex);
+            g_active_server = server;
+            g_pending_config.reset();
```
MrNeRF: Blocker: this reset drops a queued config on detach. Worker pops `enable` and is inside `start()`; the UI stages `disable` into `g_pending_config`; this discards the disable and the join waits for the in-flight enable — listener left running. Production survives because both call sites in application.cpp follow with `mcp_http.stop()`, but `RapidQueuedDisableCannotLeaveAListenerRunning` asserts the stronger guarantee and fails intermittently (repro ~1/200 with `--gtest_repeat=1000`: `phase == Running`). Either drain the pending config into the outgoing server before joining, or document detach-as-cancel and make the test `stop()` before asserting.

tests/test_mcp_protocol.cpp
```diff
@@ -711,10 +745,236 @@ namespace lfs::mcp {
...
 
-    TEST(McpHttpServerTest, ToolHandlerThrowRespondsWithInternalErrorAndEchoedIdNoLeak) {
+    TEST(McpHttpServerTest, RapidQueuedDisableCannotLeaveAListenerRunning) {
+        McpHttpServer server;
+        ASSERT_TRUE(server.start(McpHttpConfig{.enabled = false, .port = 47698}));
+        setActiveMcpHttpServer(&server);
+
+        ASSERT_TRUE(applyActiveMcpHttpConfig({.enabled = true, .port = 47698}));
+        ASSERT_TRUE(applyActiveMcpHttpConfig({.enabled = false, .port = 47698}));
+        setActiveMcpHttpServer(nullptr);
```
MrNeRF, line 905: This is the racy step: if the worker already popped the enable and is inside `start()`, the queued disable sits in `g_pending_config` and `setActiveMcpHttpServer(nullptr)` discards it — the assertions below then see Running. Fails ~0.5% of runs here under `--gtest_repeat=1000`. While in here: 47697/47698 are fixed ports; an unlucky occupant fails the suite, ephemeral ports would be sturdier.

src/visualizer/gui/rml_status_bar.cpp
```diff
@@ -1347,7 +1551,15 @@ namespace lfs::vis::gui {
             return;
         }
 
-        queueCachedVulkanContext(x, y, w_px, h_px, screen_w, screen_h,
+        int window_x = 0;
+        int window_y = 0;
+        if (auto* const window = rml_manager_->getWindow())
+            SDL_GetWindowPosition(window, &window_x, &window_y);
```
MrNeRF: `x`/`y` here are already window-client (`screen.work_pos = {0, menu_h}`, gui_manager.cpp:4949), and `trackContextFrame` stores a window-client origin — `processInput` passes `bar_x - input.screen_x` with `screen_x == 0`. Subtracting the desktop window position shifts the overlay rect by minus the window position, and since render/renderCached run after processInput each frame, this wrong origin wins. On Wayland SDL reports (0,0) so it looks fixed; on X11/Windows the popup is still click-through (the round-1 bug). Same in `renderCached` below. Drop the `SDL_GetWindowPosition` subtraction and pass `(x, y - overlay_height)` like processInput does. The new unit test drives `trackContextFrame` directly, so it can't catch this — worth a case going through the render math with a non-zero window origin.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -249,15 +284,219 @@ def _set_remember_view_snap(self, enabled):
...
+            "mcp_log_file",
+            "mcp_has_log_file",
+        ):
+            self._handle.dirty(name)
+        if self._document:
+            endpoint_list = self._document.get_element_by_id("mcp-endpoints")
+            if endpoint_list:
+                endpoint_list.set_attribute("rows", str(self._mcp_endpoint_rows()))
+
     def _on_close(self, _handle, _event, _args):
```
MrNeRF, line 508: This handler is unreachable from the title bar: `#close-btn` is wired by `Panel.on_mount` (_lfs_panel_contract.py:46-51) to `set_panel_enabled(False)` only, and nothing fires the "close" model event. Disable doesn't unmount, so the draft survives: type a port, close via the title bar, reopen, footer OK — the stale draft commits and restarts the listener. `test_title_bar_close_discards_unconfirmed_mcp_port` calls `_on_close` directly, so it passes against the broken wiring. Bind `#close-btn` to `_on_close` in `PreferencesPanel.on_mount` and drive the test through that listener. connecting-clients.md:113 documents this behavior, so right now the docs are ahead of the code.

src/mcp/mcp_http_server.cpp
```diff
@@ -83,30 +358,180 @@ namespace lfs::mcp {
...
+    McpHttpServer::~McpHttpServer() {
+        stop();
+    }
+
+    void McpHttpServer::stopListenerAndJoin() {
+        if (listener_thread_.joinable()) {
+            // bind_to_port() and listen_after_bind() are separate in cpp-httplib.
+            // Wait until the listener has entered its accept loop (or failed)
+            // before stop(): cpp-httplib ignores stop() while startup is pending.
+            http_server_->wait_until_ready();
```
MrNeRF, line 478: Two httplib edges left here. (1) If `listen_internal` exits via exception (e.g. ThreadPool construction failing), neither `is_running_` nor `is_decommissioned` gets set — `run_guarded` swallows the throw — and this wait spins forever with the lifecycle transition held, wedging every later start/stop and shutdown. Having the listener thread call `http_server_->decommission()` when `!is_running()` after `run_guarded` returns closes it. (2) httplib's accept-error path closes `svr_sock_` without invalidating it, and `wait_until_ready` guarantees we can enter `stop()` while `is_running_` is still true — `stop()` then shuts down a stale, possibly reused fd. Skipping the stop-side close when `!is_running()` after the wait avoids the double-close.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +540,228 @@ namespace lfs::mcp {
...
+        applied_config_ = {
+            .enabled = true,
+            .expose_network = config.expose_network,
+            .port = config.port,
+            .request_logging = request_logging,
+        };
+        has_applied_config_ = true;
+        {
+            std::lock_guard status_lock(status_mutex_);
+            status_.running = true;
```
MrNeRF: If the accept loop dies between `wait_until_ready()` and here, the exit handler has already written Failed (generation still matches, phase was Starting) and this block overwrites it with Running plus a cleared error — permanently, since status polling won't re-check and only a manual re-apply restarts. Re-check `is_running()` under `status_mutex_` before publishing, or make the exit handler's Failed sticky within the same generation.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +540,228 @@ namespace lfs::mcp {
...
+                            if (!g_config_cv.wait(lock, stop_token, [] {
+                                    return g_pending_config.has_value();
+                                }))
+                                break;
+                            active_server = g_active_server;
+                            config = *g_pending_config;
+                            g_pending_config.reset();
+                        }
+                        if (active_server)
+                            active_server->applyConfig(config);
```
MrNeRF, line 827: No exception barrier: a throw out of `applyConfig` (`appendSessionLog` does json + filesystem, `networkEndpoints` allocates) propagates out of the jthread callable and terminates the process — from a logging path, while the request handler itself is carefully wrapped in `try_or_log`. And if the worker ever dies without terminating, `applyActiveMcpHttpConfig` keeps staging and notifying nobody, so the UI sticks at Starting forever. A try/catch here that logs and marks status Failed keeps the loop alive.

src/mcp/mcp_http_server.cpp
```diff
@@ -115,18 +540,228 @@ namespace lfs::mcp {
...
+                                     .phase = McpHttpPhase::Disabled,
+                                 };
+    }
+
+    bool applyActiveMcpHttpConfig(const McpHttpConfig& config) {
+        {
+            std::lock_guard lock(g_active_server_mutex);
+            if (!g_active_server || !g_config_worker.joinable())
+                return false;
+            g_active_server->stageConfig(config);
```
MrNeRF: `stageConfig` runs `getifaddrs` under `g_active_server_mutex` on the caller's (UI) thread, and `activeMcpHttpStatus` holds the same mutex across `status()`, which waits on `log_mutex_` — held across disk appends when request logging is on. The status bar polls this on the render thread every 100-200 ms and the panel every 50 ms, so a slow adapter walk or log write becomes a frame hitch. Snapshot `g_active_server` under the mutex, release, then stage/query.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -249,15 +284,219 @@ def _set_remember_view_snap(self, enabled):
...
+        self._commit_mcp_port()
+
+    def _commit_mcp_port(self):
+        port = self._validated_mcp_port()
+        if port is None:
+            self._dirty_mcp()
+            return False
+        if port == self._mcp_applied_port:
+            status = lf.ui.get_mcp_status()
+            if status.get("enabled") and (
```
MrNeRF: `enabled and not running` is also true while `phase == "starting"`, so OK on the already-applied port re-stages the config and can restart a bind that was about to succeed. Gate the retry on `phase == "failed"` (or `error_kind` in bind_failed / listener_failed / invalid_port). `test_failed_bind_can_retry_the_same_confirmed_port` only sets `running=False` + `error`, so it doesn't pin this down.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -249,15 +284,219 @@ def _set_remember_view_snap(self, enabled):
...
+            bool(status.get("expose_network")),
+            int(status.get("port", 45677)),
+            bool(status.get("request_logging")),
+        )
+
+    def _sync_mcp_runtime(self):
+        signature = self._mcp_runtime_config_signature()
+        if signature == self._last_mcp_runtime_config:
+            return
+        self._load_mcp_preferences()
```
MrNeRF, line 420: This reload also overwrites `_mcp_port` while the user is typing: any external change (status-bar power switch, Ctrl+Shift+M/N) changes the signature and replaces the draft mid-edit with no confirm. If `_mcp_port != str(self._mcp_applied_port)`, keep the draft and refresh only the other fields.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -249,15 +284,219 @@ def _set_remember_view_snap(self, enabled):
...
+        self._mcp_applied_port = int(self._mcp_port)
+        self._mcp_request_logging = bool(preferences.get("request_logging", False))
+        self._last_mcp_runtime_config = self._mcp_runtime_config_signature()
+
+    def _load_mcp_preferences(self):
+        self._read_mcp_preferences()
+        self._dirty_mcp()
+
+    def _set_mcp_enabled(self, enabled):
+        self._mcp_enabled = bool(enabled)
```
MrNeRF, line 313: Under LFS_SAFE_MODE this switch turns on and stays on while the server stays off (persist and start are both no-ops), and `_last_mcp_runtime_config` is taken after the apply so `_sync_mcp_runtime` never corrects it. Expose safe mode through `get_mcp_status`/`get_mcp_preferences` and disable the MCP controls with a one-line notice instead.

src/visualizer/gui/rml_status_bar.cpp
```diff
@@ -932,6 +994,80 @@ namespace lfs::vis::gui {
...
+                                  ? "status_bar.mcp_disabled_network_detail"
+                                  : "status_bar.mcp_disabled_detail");
+                color = colorToRml(p.text_dim);
+            } else if (status.phase == RuntimeServicePhase::Starting) {
+                summary = LOC("status_bar.mcp_starting");
+                const auto endpoint = std::format("{}:{}",
+                                                  status.network_exposed ? "0.0.0.0"
+                                                                         : "127.0.0.1",
+                                                  status.port);
+                details = std::vformat(LOC("status_bar.mcp_starting_detail"),
```
MrNeRF: Note: `std::vformat` on a locale-supplied string throws `std::format_error` on a malformed brace, and this runs in the render loop. All 10 shipped translations are currently well-formed ({} present everywhere), so just future-proofing: a try/catch falling back to the raw string (here and in the bind-failed branch) would keep a bad translation from crashing the frame.

MrNeRF (changes requested):
The lifecycle rework is solid. I traced all three round-1 deadlock chains against the vendored httplib and they are genuinely closed: status snapshots are off the lifecycle mutex, the coalescing worker keeps listener joins off the UI thread, and `wait_until_ready()` before `stop()` closes the enable-then-disable race. The keymap migration test is fixed, and all 134 area tests plus the python panel tests pass here. Nice work.

One blocker: `setActiveMcpHttpServer(nullptr)` discards `g_pending_config` before joining the worker, so a queued disable is dropped whenever the worker is still applying the earlier enable — and the new `RapidQueuedDisableCannotLeaveAListenerRunning` test asserts exactly the guarantee this breaks. It fails intermittently here (~1 in 200 runs under `--gtest_repeat=1000`, `phase == Running`). The app survives only because both call sites follow the detach with `mcp_http.stop()`. Either drain the pending config on detach, or define detach as cancel and make the test `stop()` before asserting.

Two of the round-1 fixes don't actually fire in production:
- The popup overlay registration in `render`/`renderCached` subtracts `SDL_GetWindowPosition` from coordinates that are already window-client, and it runs after `processInput`, so the wrong rect wins every frame. Invisible on Wayland (position reports 0,0), still click-through on X11/Windows.
- Title-bar close still only calls `set_panel_enabled(False)`; `_on_close` is bound to a "close" model event nothing fires, and the new test calls `_on_close` directly, so it stays green while the port draft survives.

Smaller things, all inline: `wait_until_ready()` can spin forever if `listen_internal` exits via exception; `start()` can publish Running over a listener-death Failed; the config worker has no exception barrier; `g_active_server_mutex` is held across `getifaddrs` and the log append lock on the render-thread poll path; python tools still hold the GIL across their body (last remnant of round-1 chain c); the panel re-applies during `starting`, clobbers an in-progress port draft on external toggles, and desyncs the enable switch under safe mode.

Details inline.

francescofugazzi:
Fixed the remaining MCP review issues.

- Defined detach as a deterministic cancellation boundary: pending work is discarded, any in-flight transition is joined, and the outgoing listener is explicitly stopped before detach completes.
- Hardened the cpp-httplib lifecycle against listener exceptions, readiness waits that could otherwise spin indefinitely, accept-loop double-close hazards, and a failed listener being overwritten with Running.
- Added an exception barrier around the long-lived configuration worker so configuration or logging failures report a Failed state without terminating the process.
- Reduced render/UI-thread contention by releasing the global server lock before status queries and configuration staging, separating log-state and append synchronization, caching the active log path, and avoiding adapter discovery during UI staging.
- Released the Python GIL around native MCP preference, status, configuration, and log-directory operations.
- Wired the actual Preferences title-bar close button to discard unconfirmed port drafts. Footer OK still validates and applies the draft explicitly.
- Limited same-configuration retries to Failed listeners, preserved unconfirmed port drafts across external runtime toggles, and exposed safe mode explicitly with disabled MCP controls and a localized diagnostic notice.
- Corrected the status-bar popup hit region to use window-client coordinates and added coverage through the production coordinate helper with a non-zero window origin.
- Made localized status formatting resilient to malformed translated format strings.
- Disabled MCP status-bar and shortcut mutations while safe mode is active.
- Replaced fixed test ports with ephemeral ports and extended lifecycle, detach, retry, safe-mode, draft, GIL, overlay, and listener-failure coverage.
- Added single-process listener ownership. An occupied address and port now produce a real BindFailed state instead of allowing two listeners to report Listening through socket reuse.
- Restored scrolling in vertically constrained Preferences windows so MCP content cannot overlap the fixed footer.

MrNeRF:
Thank you!

Issue #1654, opened 2026-08-17
Bug - opening a project file that is no longer present throws an error


kmeirlaen:
opening a project file that is no longer present should inform the user. Right now, the application throws an error in the console and the user is not informed/aware that the project was not properly loaded / present:

```
[14:19:17.592] [error] py_ui_menus.cpp:540  Menu 'lfs_plugins.file_menu.FileMenu' callback execution error: Traceback (most recent call last):
  File "C:\Users\Gebruiker\Downloads\LichtFeld-Studio-windows-v0.5.3.dev20260817+4c84ad27\bin\lfs_plugins\file_menu.py", line 310, in <lambda>
  File "C:\Users\Gebruiker\Downloads\LichtFeld-Studio-windows-v0.5.3.dev20260817+4c84ad27\bin\lfs_plugins\file_menu.py", line 25, in _confirm_discard_then
  File "C:\Users\Gebruiker\Downloads\LichtFeld-Studio-windows-v0.5.3.dev20260817+4c84ad27\bin\lfs_plugins\file_menu.py", line 312, in <lambda>
NotFoundError: The project could not be opened.
```

* open LFS
* load a dataset folder and import the dataset
* start training
* pause training
* save the project as "project.licht"
* close LFS
* remove or rename project.licht
* open LFS
* confirm the "open recent" contains the "project.licht" reference in the correct path
* open the "project.licht" from the path using the "open recent" menu
* confirm a dialog is presented to the user: "remove from recent list" or "cancel opening"
--> no dialog displayed, instead, an error appears in the log

Alternative: only a dialog that the file could not be found and automatically remove the file from the recent list.

PR #1672, opened 2026-08-18
Add persistent spatial scene reconstruction for the viewport

francescofugazzi:
## Summary

Adds the first self-contained scene reconstruction backend for the viewport.

The initial backend is a Vulkan-only spatial reconstruction path, designed as the stable foundation for later temporal and optional vendor backends without changing training, scene data, or splat storage.

## What changed

- Added a registry with stable backend IDs, backend requirements, presets, and explicit requested/effective/fallback state.
- Added `Off` and `Spatial` scene-reconstruction choices under Appearance → Scene Rendering.
- Added backend-specific Spatial presets:
  - Quality — 75% input resolution
  - Balanced — 67%
  - Performance — 50%
- Persisted the selected backend and per-backend preset in user preferences.
- Kept safe mode isolated: it uses the native path and neither reads nor writes the reconstruction preference.
- Added Python, IPC, visualizer, and generated-stub support for the new settings.
- Added a lazy Vulkan spatial filter pipeline; no CUDA or CPU staging path is introduced.
- Composed reconstruction input scale with the existing viewport render scale while keeping native rendering unchanged.
- Fixed split-view invalidation: moving the divider now re-renders independent dual panels instead of incorrectly reusing a cached composition; compatible comparison modes retain the efficient cached divider update.
- Added runtime telemetry for requested/effective backend, fallback state, and composed render scale.
- Added locale entries for all shipped languages.
- Documented the backend contract, persistence, scale composition, and future temporal/vendor extension points.

## Scope and guarantees

- This PR affects viewport presentation only.
- It does not alter training, checkpoints, `.licht` persistence, model parameters, PLY storage, or live splat representation.
- The Spatial backend has no temporal inputs or history dependency.
- The split-view fix applies independently of the selected scene-reconstruction backend.
- Temporal, DLSS, and FSR backends remain intentionally out of scope for follow-up PRs.

## Validation

- Focused C++ tests: 18 passed.
- Python localization and preferences tests: 42 passed.
- Python stubs regenerated successfully.
- Manual verification completed with a loaded 2M-Gaussian PLY, switching repeatedly between Off and Spatial.


src/visualizer/rendering/rendering_manager_vulkan.cpp
```diff
@@ -1588,7 +1590,12 @@ namespace lfs::vis {
             markDirty(resize_result.dirty);
         }
         const bool resize_deferring = frame_lifecycle_service_.isResizeDeferring();
-        float scale = std::clamp(frame_settings.render_scale, 0.25f, 1.0f);
+        const auto requested_upscaler = sceneUpscalerBackendFromId(frame_settings.scene_upscaler)
+                                            .value_or(SceneUpscalerBackend::Native);
+        float scale = effectiveSceneRenderScale(
```
MrNeRF, line 1659: The scale here follows what was requested, but the decision to fall back happens later in the viewport pass. If the spatial pipeline fails to build, we keep rendering at the small size and present it without the sharpen: a blurry viewport that never recovers until the user switches the feature off. Please base the scale on the effective backend, or force scale 1.0 while the fallback is active. The docs promise exactly that ("Native presentation ignores the multiplier").

src/visualizer/rendering/passes/vulkan_viewport_pass.cpp
```diff
@@ -1604,6 +1610,29 @@ namespace lfs::vis {
                                   frustum_descriptor_layout);
         }
 
+        [[nodiscard]] bool ensureSpatialScenePipeline() {
+            if (scene_spatial_pipeline != VK_NULL_HANDLE)
+                return true;
+            if (scene_spatial_pipeline_attempted)
+                return false;
+
+            scene_spatial_pipeline_attempted = true;
```
MrNeRF: If the build fails once, this flag blocks any retry for the rest of the session, and the layout created just before the failure is never destroyed. Please clean up the layout on failure and allow a retry, for example when the user toggles the setting.

src/visualizer/rendering/passes/vulkan_viewport_pass.cpp
```diff
@@ -2184,6 +2239,20 @@ namespace lfs::vis {
                 // content_rect arrives panel-local; lift it into framebuffer
                 // coords so the shader's letterbox check matches gl_FragCoord.
                 VulkanSplitViewParams adjusted = params.split_view;
+                adjusted.spatial_filter =
```
MrNeRF: This turns the sharpen on for both panels. In GT compare that also sharpens the ground truth photo, and the rendered panel is not drawn at the reduced scale either, so the comparison is no longer apples to apples. Please only filter panels that were really rendered at the reduced size.

src/visualizer/rendering/passes/vulkan_viewport_pass.cpp
```diff
@@ -2184,6 +2239,20 @@ namespace lfs::vis {
...
+                    scene_upscaler_selection.effective == SceneUpscalerBackend::Spatial;
+                if (adjusted.coordinate_extent.x > 0 && adjusted.coordinate_extent.y > 0) {
+                    const float scale_x = static_cast<float>(rect.width) /
+                                          static_cast<float>(adjusted.coordinate_extent.x);
+                    const float scale_y = static_cast<float>(rect.height) /
+                                          static_cast<float>(adjusted.coordinate_extent.y);
+                    adjusted.content_rect = {
```
MrNeRF: Nit: rounding x, y, width and height separately can make an even letterbox uneven by one pixel. Rounding the edges instead (x and x+w) keeps it exact.

src/visualizer/rendering/scene_upscaler_registry.hpp
```diff
@@ -0,0 +1,83 @@
...
+        Native = 0,
+        Spatial,
+    };
+
+    enum class SceneUpscalerFallback : std::uint8_t {
+        None = 0,
+        RuntimeUnavailable,
+    };
+
+    struct SceneUpscalerRequirements {
```
MrNeRF: Nothing reads these flags yet, and `built_in` below is always true and never read either. Please drop both for now and add them together with the first backend that actually needs them. Same for the test that only exists to poke them.

src/python/lfs/py_rendering.cpp
```diff
@@ -789,6 +789,23 @@ namespace lfs::python {
...
+                   "scene_upscaler",
+                   "Scene Reconstruction",
+                   "Stable scene reconstruction backend identifier",
+                   "native");
+        add_string(&Proxy::scene_upscaler_preset,
+                   "scene_upscaler_preset",
+                   "Scene Reconstruction Preset",
+                   "Backend-specific scene reconstruction quality preset",
+                   "native");
+        add_float(&Proxy::scene_upscaler_scale,
```
MrNeRF: This value is always overwritten from the preset in `updateSettings`, so setting it from Python silently does nothing. Please make it read only or drop it from the proxy.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -33,6 +33,17 @@ class PreferencesPanel(Panel):
         (2.0, "200%"),
     )
 
+    SCENE_UPSCALER_OPTIONS = (
```
MrNeRF: These tables duplicate the C++ registry, and `_sync_scene_reconstruction_catalog` always succeeds against the real binding. Please drop the copies and seed from the registry only, like themes already do. Otherwise a third backend can show up in the app but stay missing here.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -610,6 +765,10 @@ def _reset_section(self, section=None):
         elif section == "appearance":
             lf.ui.set_theme("dark")
             lf.ui.set_ui_scale(0.0)
+            settings = lf.get_render_settings()
+            if settings is not None:
+                settings.scene_upscaler = "native"
+                settings.scene_upscaler_preset = "native"
```
MrNeRF: After this reset the preset record list is not rebuilt, and `_refresh_selection` stamps the state so `on_update` will not repair it later. Please call `_sync_scene_upscaler_preset_records()` here.

src/python/stubs/lichtfeld/__init__.pyi
```diff
@@ -287,6 +287,9 @@ def project_recent_files() -> list[str]:
 def project_clear_recent_files() -> None:
     """Clear the most-recently-used .licht project list"""
 
+def project_remove_recent_file(path: str) -> None:
```
MrNeRF, line 290: This one is from another change, the stub regen picked it up. Please drop it from this PR.

tests/python/test_preferences_panel.py
```diff
@@ -401,6 +474,17 @@ def test_preferences_content_scrolls_without_overlapping_the_fixed_footer():
     assert "padding-right: 6dp;" in content_rule
 
 
+def test_scene_reconstruction_catalog_comes_from_the_native_registry():
```
MrNeRF: This only checks that three strings appear somewhere in the file, it would still pass if they were in a comment. The C++ registry test already owns this contract, please drop it.

MrNeRF (changes requested):
Thanks Francesco, nice feature. Off by default, the settings and persistence work is clean, all languages are covered, and the split view divider fix is a welcome bonus. A few things need fixing before merge, details inline. Two more that have no good inline spot:

- Split view screenshots: `composeSplitViewCpu` was not updated, so with Spatial on a capture does not get the sharpen the screen shows. The comment above that function promises they match, please keep that true.
- MCP: `render.settings.get/set` in `mcp_gui_tools.cpp` does not know the new fields, so scripts can neither see nor change the setting. Please add them.

All of it in this PR please, no follow-ups.

francescofugazzi:
**Scale follows the effective backend, not the request.** The viewport pass
now reports its actual runtime selection back to the rendering manager, and the
render scale is only reduced once the spatial pipeline is confirmed ready
(effective == Spatial, no fallback). While the pipeline is unavailable or has
fallen back, native resolution is kept, so a failed build can never leave a
blurry, un-sharpened viewport that only recovers on a manual toggle.

**Retry + cleanup on failure.** The one-shot `attempted` guard is gone. If the
spatial pipeline build fails, the partially created Vulkan pipeline and layout
are destroyed and the next toggle retries cleanly.

**Per-panel sharpen, never on ground truth.** The sharpen filter moved from a
global flag to a per-panel flag. Ground-truth comparisons are deliberately left
unfiltered — the GT photo stays raw, and the rendered panel is not filtered
either since it is not rendered at the reduced scale, keeping the comparison
apples-to-apples. CPU split-view captures now apply the same sharpen as the
screen presentation, so screenshots match the display.

**Letterbox rounding.** The content rect is now rounded by its edges rather
than by x/y/width/height separately, so an even letterbox stays pixel-exact.

**MCP.** `render.settings.get/set` now expose `scene_upscaler` and
`scene_upscaler_preset`, both in the JSON payload and the schema.

**Registry cleanup.** Dropped the unused `SceneUpscalerRequirements` struct,
the always-true `built_in` field, and the probe test that only poked them.
They'll return alongside the first backend that actually needs them.

**Python proxy.** Removed the dead writable `scene_upscaler_scale` binding (it
was always overwritten from the preset, so writing it silently did nothing).

**Preferences catalog.** Removed the duplicated backend/preset tables; the panel
now seeds exclusively from the native registry, like themes already do, and the
preset list is rebuilt right after an appearance reset.

**Stub + test scope.** Dropped the unrelated `project_remove_recent_file` stub
that the stub regeneration picked up, and the weak catalog string test (the C++
registry test already owns that contract).

**Additional fix found during validation.** While validating with ground-truth
comparison active, toggling scene reconstruction (native/spatial/presets) could
momentarily flash a different camera's reference image. The GT preview was being
sized from the reconstruction-reduced render size, so changing the setting
changed the GT cache key, forcing a reload that showed a stale cache entry
(possibly a neighbor camera) for a frame. The GT comparison is now sized solely
from the viewport and the base render scale, so it stays stable regardless of
the reconstruction setting, and the stale fallback now prefers the same camera's
previous image instead of an arbitrary most-recently-used entry.

tests/python/test_preferences_panel.py
```diff
@@ -104,6 +147,36 @@ def test_language_selection_does_not_reload_active_language(preferences_panel_mo
     assert state.set_language_calls == []
 
 
+def test_scene_reconstruction_uses_backend_specific_presets(preferences_panel_module):
+    module, state = preferences_panel_module
+    panel = module.PreferencesPanel()
+    panel._refresh_selection = lambda: None
+
+    panel._set_scene_upscaler_index("1")
```
MrNeRF, line 285: Blocker: this test fails deterministically. Since the catalog moved to registry seeding, a freshly constructed panel has an empty `_scene_upscaler_catalog`, so this call returns without doing anything and the first assert fails. The sibling test below got the `_sync_scene_reconstruction_catalog()` call, this one did not. Our CI never runs this file, so the green checks say nothing here; please re-run the suite locally after the fix.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -241,6 +272,112 @@ def _set_scale_index(self, value):
...
+        try:
+            remembered_preset = str(lf.ui.get_scene_reconstruction_preset_preference(backend))
+        except AttributeError:
+            remembered_preset = default_preset
+        preset = (
+            remembered_preset
+            if any(item[0] == remembered_preset for item in presets)
+            else default_preset
+        )
+        settings.scene_upscaler = backend
```
MrNeRF: Blocker: the remembered preset does not survive this in the real app. These two assignments are two full snapshot commits; the first still carries the old backend's preset, `updateSettings` replaces it with the default, and that default is persisted over the remembered slot. And after the record list rebuild the preset select writes index 0 back, so the second assignment does not stick either. Live result: Performance → Off → Spatial ends on Quality and preferences.json now remembers quality. I would fix it in `updateSettings`: when the backend changes and the incoming preset is not valid for it, restore the remembered per-backend preset instead of the default. That repairs this path, MCP and Python in one place.

src/app/mcp_gui_tools.cpp
```diff
@@ -981,6 +983,8 @@ namespace lfs::app {
             set_bool("mip_filter", settings.mip_filter);
             set_int("sh_degree", settings.sh_degree);
             set_float("render_scale", settings.render_scale);
+            set_string("scene_upscaler", settings.scene_upscaler);
```
MrNeRF: Note: unknown ids are silently coerced and the coercion is persisted. A set of `"dlss"` reports success, becomes native, and overwrites the user's saved spatial selection. Please reject unknown backend and preset ids the way `raster_backend` does, and list the valid values in the schema.

src/visualizer/rendering/passes/vulkan_viewport_pass.cpp
```diff
@@ -1973,6 +2008,8 @@ namespace lfs::vis {
         }
 
         void prepare(const VulkanViewportPassParams& params) {
+            if (params.scene_upscaler == SceneUpscalerBackend::Spatial)
+                static_cast<void>(ensureSpatialScenePipeline());
```
MrNeRF, line 2018: Note: this retries the build on every frame while Spatial is requested. On a permanently failing device that is create-modules, create-layout, fail, destroy at frame rate, with a LOG_ERROR each time. I asked for a retry on toggle; please latch on the off→on edge (or a cooldown).

src/visualizer/rendering/passes/vulkan_viewport_pass.cpp
```diff
@@ -2007,6 +2044,30 @@ namespace lfs::vis {
             environment_pass.prepare(params.environment, params.frame_slot);
             depth_blit_pass.prepare(params.depth_blit, params.frame_slot);
             split_view_pass.prepare(params.split_view, params.frame_slot);
+            const bool runtime_available = params.split_view.enabled
+                                               ? split_view_pass.ready(params.frame_slot)
```
MrNeRF: Note: `ready()` is “this slot has panels bound”, not “the filter is available”. A transient bind failure reports fallback, the next frame drops to native, and you get a resolution pop plus one blurry unfiltered frame on the way back. For split view the filter lives in split_view.frag, so availability is really just “the split pipeline exists”.

src/visualizer/rendering/rendering_manager_vulkan.cpp
```diff
@@ -750,13 +752,31 @@ namespace lfs::vis {
...
+                            sample(std::get<0>(data), std::get<1>(data), std::get<2>(data),
+                                   sample_u, sample_v, 0),
+                            sample(std::get<0>(data), std::get<1>(data), std::get<2>(data),
+                                   sample_u, sample_v, 1),
+                            sample(std::get<0>(data), std::get<1>(data), std::get<2>(data),
+                                   sample_u, sample_v, 2)};
+                    };
+                    glm::vec3 color = sample_color(panel_u, panel_v);
+                    if (panel.spatial_filter) {
+                        constexpr float kSpatialSharpenStrength = 0.18f;
```
MrNeRF, line 780: Note: the kernel matches the shader, the sampling does not: nearest here vs bilinear on screen, the padded-texture clamp is ignored, and the compose target is render_size — so a GT compare capture with Spatial on comes out at the reduced resolution while the screen shows the panels native. The comment above this function still promises captures match the screen.

src/visualizer/rendering/rendering_manager_vulkan.cpp
```diff
@@ -1804,6 +1881,7 @@ namespace lfs::vis {
 
         const DirtyMask split_deferred_dirty = frame_dirty & ~DirtyFlag::SPLIT_POSITION;
         if ((frame_dirty & DirtyFlag::SPLIT_POSITION) != 0 &&
+            !split_position_requires_panel_rerender &&
```
MrNeRF, line 1901: Note: the resize-defer block below still calls `update_cached_split_position` without this guard, so a divider drag during resize deferral stretches the stale panel textures in IndependentDual — the same artifact this fix removes elsewhere.

src/visualizer/gui/rmlui/resources/preferences.rml
```diff
@@ -50,6 +50,28 @@
...
 
+      <div class="preferences-section" data-if="show_appearance">
+        <div class="section-header text-accent" data-class-is-expanded="scene_rendering_expanded" data-event-click="toggle_section('scene_rendering')">
+          <span class="section-arrow text-accent" data-class-is-expanded="scene_rendering_expanded">&#x25B6;</span>
+          <span class="text-accent">@tr:preferences.scene_rendering</span>
+        </div>
+        <div class="section-content" data-if="scene_rendering_expanded">
+          <div class="preferences-row">
+            <span class="preferences-label">@tr:preferences.scene_reconstruction</span>
+            <select class="preferences-select" data-value="scene_upscaler_idx">
```
MrNeRF: Note: safe mode does not disable these selects, so a safe-mode session can still switch Spatial on live. The startup and preference isolation are correct (I verified the file stays untouched), but scene-reconstruction.md promises native for the process. Either copy the MCP disable pattern or soften the doc.

src/python/lfs_plugins/preferences_panel.py
```diff
@@ -610,6 +747,11 @@ def _reset_section(self, section=None):
         elif section == "appearance":
             lf.ui.set_theme("dark")
             lf.ui.set_ui_scale(0.0)
+            settings = lf.get_render_settings()
+            if settings is not None:
+                settings.scene_upscaler = "native"
```
MrNeRF: Note: this resets the live setting but leaves the other backends' remembered slots in `scene_upscaler_presets`, while the reset-all confirmation copy says scene reconstruction is restored to defaults. Minor, but either clear the slots or adjust the copy.

MrNeRF (changes requested):
Thanks Francesco, good round. The scale now genuinely follows the effective backend, the pipeline cleanup and the per-panel filter are exactly what I asked for, and I could not get the fallback path to strand a blurry viewport anymore. I built the branch and drove it live for a while. Three things still block the merge, details inline.

One of them has no good inline spot, and it is my fault: dropping the `project_remove_recent_file` stub was a bad call on my side. The binding shipped with #1654 without its stub, so the regen keeps re-adding it and a default local build now fails at the stub check (CI hides this with `-DBUILD_PYTHON_STUBS=OFF`). Please put the stub entry back, with my apologies for the detour.

The other two blockers are inline: the new Python test that cannot pass, and the per-backend preset memory, which in the running app loses the remembered preset and overwrites it in preferences.json. I reproduced that with the panel alone: Performance, switch to Off, switch back to Spatial, and you are on Quality with the file rewritten.

Smaller things, one line each, also inline: MCP set silently coerces unknown ids and persists the coercion over the saved choice; a failed spatial pipeline retries every frame and logs an error each time; CPU split captures still do not match the screen under Spatial; split view reports availability from the panel-bind state; the resize-defer path moves the cached divider without the new IndependentDual guard; safe mode leaves the new panel controls active.

Nothing structural. The registry, the persistence shape and the feedback loop are right. With the three blockers fixed this merges.

francescofugazzi:
Addressed the final review points:

- fixed backend-specific scene reconstruction presets across `Performance → Off → Spatial`, including the RML option rebuild path;
- added regression coverage for catalog initialization, preset restoration, and active preset selection;
- validated MCP backend/preset identifiers and exposed the supported values in the tool schema;
- latched failed spatial pipeline creation until an explicit Native → Spatial retry;
- based split availability on pipeline availability rather than per-frame panel binding;
- preserved IndependentDual divider behavior during resize deferral;
- disabled scene reconstruction controls in safe mode;
- rebuilt preset records after Appearance reset;
- aligned CPU split captures with the displayed output, including sampling and padded-texture bounds;
- reset all remembered scene-reconstruction preset slots with Preferences reset;
- restored the required Python stub entry.

MrNeRF:
Just the merge conflicts and I will take later another look. Thank you :)

MrNeRF:
Ok, thank you!

PR #1680, opened 2026-08-19
Align Input Settings rebind rows

francescofugazzi:
Gives the Input Settings binding rows a clean three-column layout: the action name, the key combination and the Rebind button are now consistently spaced, and the button stays pinned to the right column across all rows instead of sitting flush against each description.


MrNeRF:
thx!

Issue #1693, opened 2026-08-19
Clicking outside of a focused input field does not always unfocus it


thomas-koch-core11:
Clicking outside of a focused input field does not unfocus it except when clicking within the rendering or training panel

PR #1694, opened 2026-08-19
(#1693) Fix focused input fields not becoming unfocused when clicking outside…

thomas-koch-core11:
Fix focused input fields not becoming unfocused when clicking outside the rendering or training tab (#1693)


MrNeRF:
Best bug reports are the ones that get also fixed by the author immediately. Thank you!

PR #1725, opened 2026-08-20
Add built-in temporal scene reconstruction to the Vulkan viewport

francescofugazzi:
## Summary

This PR adds built-in temporal scene reconstruction to the Vulkan viewport alongside the existing Native and Spatial presentation paths.

Temporal reconstruction renders the scene at a backend-specific internal resolution and reconstructs the final viewport using current color, motion, depth, projection jitter, and per-view color/depth history. It is strictly a viewport presentation feature: it does not alter training tensors, Gaussian precision, `.licht` project data, or exported splats.

Native remains the default and unchanged baseline.

## Backend and preset model

The scene-reconstruction registry exposes stable runtime identifiers:

| Backend | Presets |
| --- | --- |
| `native` | `native` |
| `spatial` | `quality`, `balanced`, `performance` |
| `temporal` | `quality`, `balanced`, `performance` |

Backend-specific presets remain independent from the renderer's base `render_scale`. Switching reconstruction backend does not overwrite the global renderer scale.

The requested backend, effective backend, runtime readiness, and fallback state remain separate. Invalid identifiers use registry defaults, while runtime failures remain observable and fall back safely to Native.

## Temporal rendering path

The Vulkan temporal path adds:

- deterministic, resolution-aware projection jitter;
- current-to-previous camera reprojection;
- canonical motion-vector encoding;
- current and previous depth history;
- disocclusion and invalid-history rejection;
- full-resolution color history;
- bounded temporal accumulation and sharpening;
- explicit first-frame and warm-up handling;
- independent history for the main, training, and supported split-view panels;
- transactional history publication only after a successful temporal resolve.

Motion, depth, jitter, history, and output extents are validated through explicit contracts before temporal work is recorded.

A failed or incomplete resolve never publishes invalid history for the next frame.

## History invalidation and ownership

Temporal history is reset when required by:

- explicit camera cuts;
- reset, focus, and view-snap operations;
- keyframe and sequencer discontinuities;
- scene or backend changes;
- projection or render-extent changes;
- presentation-relevant background changes;
- invalid temporal inputs;
- failed temporal completion.

When leaving Temporal, submitted viewport work is retired before per-view color and depth history allocations are released. Immutable pipeline state remains available for later re-entry, avoiding persistent history VRAM without recreating every pipeline after each backend switch.

## Viewport coverage

Temporal reconstruction supports:

- the regular viewport;
- the training viewport;
- perspective projection;
- orthographic projection;
- Independent Dual split view;
- PLY comparison.

Split output is presented only when both panel resolves succeed; otherwise both panels fall back together.

Ground-truth comparison images remain unchanged.

Equirectangular projection and appearance-corrected readback remain Native because their projection or ownership contracts are not equivalent to the Vulkan temporal path.

## Persistence and safe mode

The selected reconstruction backend and the last valid preset for each backend are stored as user-global preferences rather than `.licht` project state.

Safe mode starts with Native presentation, disables the reconstruction controls in Preferences, and does not read or overwrite the persisted reconstruction selection. Runtime Python or plugin changes remain process-local.

## Production integration benchmark

`tools/benchmark_scene_reconstruction.py` compares the registered reconstruction backends through the running production viewport.

The benchmark:

- discovers backend and preset identifiers from the application;
- controls the viewport through MCP;
- runs deterministic camera orbits;
- warms every backend independently;
- separates performance measurement from quality capture;
- rotates and reverses backend ordering between rounds;
- waits for temporal convergence before every quality capture;
- captures the composited, presented viewport;
- compares Spatial and Temporal output against matching Native frames using PSNR and SSIM;
- records effective backend and fallback state for every phase;
- restores the original camera, backend, and preset on completion or failure;
- writes schema-v4 reports atomically, preserving partial results after later failures.

Viewport timing excludes readback, PNG encoding, Base64 conversion, and HTTP transfer. PNG latency is reported separately. The timing scope covers the application camera mutation through post-render command recording and is not presented as an isolated GPU timestamp.

Before running the benchmark, start LichtFeld Studio, load a representative Gaussian scene, stop animation or training activity, and enable the MCP server.

Short smoke run:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --frames 16 --quality-frames 4 --warmup-frames 4 --performance-rounds 2
```

Longer comparison:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --frames 128 --quality-frames 8 --warmup-frames 8 --performance-rounds 4
```

Explicit backend selection:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --backends native,spatial,temporal
```

Custom MCP endpoint and report:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --url http://127.0.0.1:45677/mcp --output build/reconstruction.json
```

Complete command reference:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --help
```

## Tests and CI

The temporal contracts have a standalone CPU-only target:

```sh
cmake -S . -B build -DBUILD_TEMPORAL_CPU_TESTS=ON
cmake --build build --target lichtfeld_temporal_cpu_tests
ctest --test-dir build --output-on-failure --no-tests=error -R '^lichtfeld\.temporal\.cpu$'
```

The target compiles the production temporal coordinator, planning, motion, resolve, history, quality-regression, and pipeline-validation code without executing CUDA or Vulkan GPU work.

Ubuntu Release CI now runs:

- 65 standalone temporal C++ tests;
- the registered `lichtfeld.temporal.cpu` CTest with zero-test detection;
- all 16 Python benchmark contract tests.

The Python benchmark tests cover backend discovery, status polling, convergence ordering, presented-viewport capture, phase isolation, state restoration, report generation, and failure handling.

The MCP-driven production benchmark remains documented as an additional manual integration test because it intentionally requires a running application, a loaded scene, and a real rendered viewport.

## Validation

Validated with:

- 65/65 standalone temporal C++ tests;
- 1/1 temporal CTest registration;
- 16/16 benchmark Python tests;
- repeated production MCP benchmark runs without backend fallback;
- Error Debt gate;
- locale completeness;
- localization contract tests;
- Visual Studio Community clang-format dry-run;
- `git diff --check`.


MrNeRF (changes requested):
I measured this and can't merge it as is.

Setup: a 3M splat scene on a 4090, camera orbiting, vsync off so the GPU is the limit.

Speed, viewport 3480 x 2316:

| | fps | ms/frame | vs Off |
|---|---|---|---|
| Off | 131 | 7.6 | 1.00x |
| Spatial / Quality | 180 | 5.6 | 1.37x |
| Temporal / Quality | 168 | 6.0 | 1.28x |
| Spatial / Balanced | 191 | 5.2 | 1.46x |
| Temporal / Balanced | 185 | 5.4 | 1.41x |
| Spatial / Performance | 217 | 4.6 | 1.66x |
| Temporal / Performance | 211 | 4.7 | 1.61x |

Speed, viewport 1560 x 1116:

| | fps | ms/frame | vs Off |
|---|---|---|---|
| Off | 244 | 4.1 | 1.00x |
| Spatial / Quality | 262 | 3.8 | 1.08x |
| Temporal / Quality | 259 | 3.9 | 1.06x |
| Spatial / Balanced | 268 | 3.7 | 1.10x |
| Temporal / Balanced | 265 | 3.8 | 1.09x |
| Spatial / Performance | 288 | 3.5 | 1.18x |
| Temporal / Performance | 286 | 3.5 | 1.17x |

Picture quality. Still frame with the camera at rest after the 8-frame settle, compared pixel by pixel against the Off render at the same camera. Higher is closer to the full-resolution picture.

| | PSNR at 3480 x 2316 | PSNR at 1560 x 1116 |
|---|---|---|
| Spatial / Quality | 33.4 dB | 31.1 dB |
| Temporal / Quality | 27.7 dB | 27.6 dB |
| Spatial / Balanced | 31.1 dB | 28.8 dB |
| Temporal / Balanced | 27.5 dB | 26.9 dB |
| Spatial / Performance | 26.8 dB | 25.2 dB |
| Temporal / Performance | 25.9 dB | 25.0 dB |

So Temporal is slower than Spatial at every preset and blurrier at every preset, at both sizes. Temporal / Quality renders at 75% and is about as blurry as Temporal / Performance at 50%, which tells me the blur doesn't come from the lower resolution. It comes from the resolve.

Each frame is rendered with a small sub-pixel offset (the jitter). When `scene_temporal_resolve.comp` reads the current frame it doesn't take that offset back out, and it never places the low-res samples onto the full-res grid by where the jitter put them. It averages eight slightly shifted copies of the picture. That's a blur, not a reconstruction. Quality keeps more history (0.95), so it blurs more. That's why "Quality" is the worst setting.

The bigger question is whether we want this at all. The only reason to have a temporal upscaler next to a spatial one is that it gives a sharper picture for the same speed. Right now it gives a softer picture for less speed. Until that flips, the user gains nothing, and we take on 5,800 lines, three extra GPU passes, full-res history buffers per view, and a new dependency between renderer and compositor that we have to carry from now on. Most of the diff isn't even the feature: plans, coordinators, tickets, contracts for vendor backends that don't exist, and tests that only check the contracts agree with each other. Nothing tests whether the output picture is good, and the PR comes with no numbers.

What it would take: fix the resolve so Temporal beats Spatial on PSNR/SSIM against Off at the same preset, static and moving, without ghosting at edges. Put your own before/after numbers in the PR. Cut it down to what the feature needs: motion pass, depth history, resolve, history tracking, one test against a high-res reference.

If that's too much, closing this and keeping Spatial is fine. One upscaler that works beats two where one is worse.

Worth keeping: bad input never shows a broken frame, the 8-frame cycle stops on its own so the GPU doesn't spin when idle, safe mode and preferences are handled right, Off and Spatial are untouched.

francescofugazzi:
Addressed the review findings and added measured quality and performance evidence:

- corrected the temporal resolve coordinate model so current color and motion are sampled on the current jittered render grid;
- removed current and previous projection jitter when addressing stable color history;
- addressed previous depth on its matching jittered grid;
- kept padded color-allocation coordinates separate from unpadded motion and depth coordinates;
- changed the eight-frame warm-up to uniform accumulation bounded by the selected preset weight;
- synchronized frame and history resets after camera cuts, backend changes, scene changes, invalid inputs, and explicit resets;
- removed the unused generic `scene_motion_contract` and `scene_upscaler_inputs` abstractions and their contract-only tests;
- added a deterministic high-resolution synthetic regression where the corrected Temporal resolve beats both Spatial and the previous Temporal resolve;
- added a moving reprojection regression that detects unreprojected ghosting;
- added a production-path benchmark that discovers registered backends and presets through MCP, measures the live viewport without readback, and compares sparse captures against matching Native frames;
- isolated all performance measurements from PNG capture and distributed cases over rotated/reversed rounds to reduce backend-order and thermal bias;
- added preflight checks, explicit fallback detection, atomic partial reports, raw samples, per-round statistics, PSNR/SSIM, and guaranteed camera/backend restoration;
- documented all benchmark options and portable `uv` commands.

Synthetic quality regression:

| Metric | Spatial | Corrected Temporal | Previous Temporal |
| --- | ---: | ---: | ---: |
| PSNR | 45.347 dB | 45.901 dB | 29.609 dB |
| SSIM | 0.999523 | 0.999581 | 0.981795 |

Moving reprojection regression:

| Path | PSNR |
| --- | ---: |
| Reprojected history | 41.499 dB |
| Unreprojected history | 29.522 dB |

Production-path run on a 2M-splat scene at 2200×1475:

| Backend / preset | Median latency | Median-derived FPS | PSNR vs Native | SSIM |
| --- | ---: | ---: | ---: | ---: |
| Native | 21.32 ms | 46.90 | reference | 1.000000 |
| Spatial / Quality | 18.43 ms | 54.26 | 29.362 dB | 0.989546 |
| Temporal / Quality | 17.50 ms | 57.13 | 29.033 dB | 0.988729 |
| Spatial / Balanced | 15.91 ms | 62.86 | 27.629 dB | 0.984440 |
| Temporal / Balanced | 17.15 ms | 58.29 | 27.282 dB | 0.983151 |
| Spatial / Performance | 15.12 ms | 66.13 | 25.904 dB | 0.977042 |
| Temporal / Performance | 15.19 ms | 65.84 | 25.280 dB | 0.973482 |

The severe Temporal Quality collapse is gone. The live production captures are now close to Spatial, but Temporal is still approximately 0.33–0.62 dB below Spatial at the corresponding presets on this scene. I am reporting that remaining gap explicitly rather than claiming that Temporal already beats Spatial in every production case.

The production timing covers camera mutation through post-render completion without viewport readback. It includes GUI scheduling and command recording and is not an isolated GPU timestamp.

Full benchmark:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --frames 128 --quality-frames 8 --warmup-frames 8 --performance-rounds 4
```

Focused Native/Spatial verification:

```sh
uv run --no-project --with numpy --with pillow python tools/benchmark_scene_reconstruction.py --backends native,spatial --frames 64 --quality-frames 8 --warmup-frames 8 --performance-rounds 2 --output build/scene-reconstruction-benchmark-spatial.json
```

Validation:

- 52 focused C++ temporal, motion, resolve, pipeline, coordinator, and quality-regression tests passed;
- 16 Python benchmark tests passed;
- both production benchmark reports completed without fallback and restored the original camera/backend state;
- error-debt baseline gate passed;
- `git diff --check` passed.

MrNeRF (changes requested):
I measured the rework and read the new code. The round-1 blur collapse is genuinely fixed — credit where due: the resolve now takes the jitter back out, and at rest the picture converges to Spatial-level quality. But the numbers in the PR are produced by a capture path that never sees the reconstructed image, and the motion/jitter contracts underneath have bugs that matter far beyond this backend — because this branch is the substrate DLSS will plug into. That changes how I judge it, in both directions. Details below.

Setup: 4080, 3.5M splat scene, ~1745x1094 viewport. My nested test display caps at 30 fps, so for speed I report GPU utilization at a fixed frame rate instead of fps — same ranking, different unit.

Picture quality, on screen, camera at rest after the 8-frame settle, against Off at the same camera:

| | Spatial | Temporal |
|---|---|---|
| Quality | 39.76 dB | 39.64 dB |
| Balanced | 37.66 dB | 37.58 dB |
| Performance | 34.62 dB | 34.52 dB |

GPU cost while orbiting, all cases locked at the same frame rate (utilization is proportional to per-frame cost):

| | GPU busy |
|---|---|
| Off | 35% |
| Spatial / Quality | 25% |
| Temporal / Quality | 34% |
| Spatial / Performance | 23% |
| Temporal / Performance | 30% |

So on screen Temporal now ties Spatial (~0.1 dB below at every preset) instead of losing 5+ dB — but it costs 30–40% more GPU than Spatial for that tie, and Temporal/Quality costs almost as much as Off. As a standalone upscaler the round-1 verdict still holds: the user gains nothing over Spatial yet.

**Why your benchmark disagrees with the screen.** `render.capture` returns the internal raster before reconstruction. Bare captures come back at 1120x809 / 747x540 on my 1493x1079 viewport — that is the reduced-resolution input, nearest-upscaled to the requested size, never the resolve output (`rendering_manager_vulkan.cpp` lazy capture reads `readOutputImage`, and `rendering_manager_viewport.cpp` has no temporal awareness at all). Your quality phase therefore compares Native's full-res raster against a nearest-upscaled pre-reconstruction raster — for Temporal a still-jittered one. I reproduced your 0.3–0.6 dB "Temporal below Spatial" gap through that exact path; on the actual screen the gap is ~0.1 dB. Your own numbers under-sell your own fix, and neither backend's reconstruction is ever measured. The synthetic 45.9-vs-45.3 test has the same problem one level down: it runs neither the compute shader nor `resolveSceneTemporalSample` but a third, simpler lerp loop with no neighborhood clamp, on a 192x128 smooth sinusoid, against a "Spatial" that is plain bilinear rather than the shipped sharpen. Those dB figures don't cover any code that ships.

**The DLSS angle.** Temporal reconstruction is the prerequisite for DLSS, and this branch builds exactly the inputs DLSS consumes: motion vectors, jitter reporting, depth, per-view history and extents. Judged as that foundation, the architecture is right — and the bugs below are disqualifying, because DLSS trusts these inputs directly and has none of the rejection heuristics that currently hide them.

What to fix, in priority order:

1. **Motion vectors are wrong in Y.** `scene_motion_reproject.comp` maps pixel y=0 to NDC y=-1 and unprojects through `glm::perspective`, which is GL-convention: NDC +1 is the top. VkSplat's image is top-left (top of view at pixel 0, `coordinate_conventions.hpp`). So top-of-image pixels are unprojected along the bottom-of-frustum ray with the wrong depth pairing; `flip_y` only negates the 2-D output afterwards and can't repair it. Identity cameras produce zero motion either way, which is why the contract test passes. Fix: flip before unprojecting — `current_ndc.y = 1.0 - 2.0 * (pixel.y + 0.5) / extent.y` (and mirror `previous_ndc` back the same way), or build the VP with a Vulkan-style Y flip and keep the shader as is; then add one non-identity test that renders two analytically known camera poses and checks the sign of the Y motion. Today the depth reject and neighborhood clamp mostly swallow the bad history, which is exactly why moving-camera Temporal gives no benefit — and DLSS would show this as ghosting instead of hiding it.

2. **Declared jitter diverges from applied jitter after convergence.** The raster uses `temporal_convergence_.jitter()`, which is zero once the 8 samples are done; the resolve and motion passes get `temporalJitterNdc(sequence_)`, which is stuck at a non-zero Halton(9) offset (`rendering_manager_vulkan.cpp:2147` vs `:2502`/`:3790`). Any re-render outside `temporal_source_dirty` — selection overlays, PPISP toggles — then resolves an unjittered image as if it were jittered: the round-1 bug in miniature, on the steady-state path. Fix: publish the jitter that was actually applied — one source of truth, `temporal_convergence_.jitter()` converted to NDC — never recompute from the sequence index. DLSS requires exact jitter reporting, so this seam has to be airtight anyway.

3. **Jitter policy during motion.** Every camera-dirty frame calls `prepare(…, restart=true)`, so while the camera moves each frame renders Halton sample 0 — constant jitter, no sample diversity; the 8-sample walk only happens after the camera stops. For the built-in resolve that means Temporal cannot gain anything while moving. For DLSS it's a blocker: vendor temporal upscalers expect a continuous jitter sequence every frame. Advance the sequence on every rendered frame; restart only the accumulation state on actual cuts, not the jitter index on every dirty.

4. **Output extent is off by one.** The output extent is `round(render_size / scale)` (`gui_manager.cpp:4196`), not the viewport size: 1920 at 0.67 becomes render 1286, output 1919. History and presentation then run 1919-wide against a 1920 viewport — a persistent sample-grid error, not just a stretch. Publish `current_size` as the output extent.

5. **Training fights itself.** Every training preview refresh (4 Hz default) sets SPLATS, which restarts the 8-frame burst and wakes seven follow-up frames whose history is invalidated by the next preview anyway. That's ~8 jittered renders plus temporal compute per preview that can never converge, contending with the trainer for the GPU. Keep the history reset, drop the burst restart and follow-up wakes when the only dirty source is the training refresh (same applies to the LOD-streaming CAMERA follow-ups).

6. **Per-backend preset memory is clobbered again.** Spatial and Temporal share preset IDs, so a backend-only settings change (MCP/Python `{"scene_upscaler": "temporal"}`) keeps the current preset because the string is valid for the new backend, then persists it over the remembered slot — I verified `scene_upscaler_presets.spatial` being overwritten on disk. The GUI path is correct because the panel explicitly loads the remembered preset. Fix in `updateSettings`: on backend change, restore the remembered preset whenever the request didn't explicitly carry one — not only when the carried ID is invalid.

7. **One resolve kernel, tested.** CPU model and shader already disagree: zero `motion_rejection_pixels` means "always blend" on CPU and "never blend" in GLSL; the shader has an extra render-space history reject, gathers its own 3x3 clamp, bilinear-samples motion with the color sampler (nearest would avoid edge ghosting), and treats far-plane depth as disocclusion — the CPU model has none of that. Make `resolveSceneTemporalSample` byte-for-byte the shader's algorithm, drive the synthetic quality test through it (with the clamp — that's the term that limits real-world gain), compare against the shipped Spatial sharpen rather than plain bilinear, and delete the "legacy resolve" strawman arm: there is no previously shipped temporal resolve to regress against.

8. **Make the benchmark measure the screen.** Capture the resolve output (or crop the swapchain, as `render.capture_window` already does) for reconstruction backends; expose the convergence counter in `render.reconstruction.status` and have the quality phase settle at the capture pose until it hits zero instead of capturing 1–8 frames into convergence. And wire the new C++ temporal tests plus the 16 benchmark unit tests into CI — today nothing runs them.

9. Smaller: free the history images when leaving Temporal (`reset` paths call `invalidateView`, which keeps ~35–40 MB of VRAM per view resident until shutdown — `resetView` is right there); `camera_cut` is plumbed through the tracker but nothing ever sets it, so view snaps rely entirely on motion rejection; background-color changes restart jitter but don't invalidate history, so the old clear color survives in history for a few frames.

What I verified working, live: requested/effective/fallback reporting is honest everywhere, including the preserved request when a mode falls back; the split view is transactional (both panels Temporal or a clean Native fallback with a reason); camera cuts show no ghosting — the first post-cut frame is clean; the 8-frame cycle stops on its own and idle GPU is ~0% on all backends; safe mode neither reads nor writes the preference; the Preferences flow, preset row, and all ten locales are right; 63 C++ and 16 Python tests pass locally.

Bottom line: with DLSS as the destination I'm no longer asking whether we want a temporal backend — this is the scaffolding for it, and the scaffolding is worth landing. But DLSS consumes motion vectors and jitter verbatim, so items 1–4 are hard blockers, 5–6 are user-visible regressions, and I won't merge quality claims produced by a pipeline that measures the pre-reconstruction raster. Fix the contracts, make the numbers come from the screen, and the built-in resolve at Spatial parity is a perfectly good no-vendor fallback tier.

francescofugazzi:
Addressed the remaining review points:

- changed quality capture to use the composited, presented viewport rather than the pre-reconstruction raster input;
- exposed reconstruction convergence through MCP and wait for complete settling before every quality capture;
- separated performance measurement from PNG capture and quality comparison phases;
- retained the documented MCP-driven integration benchmark, including smoke, extended, backend-filtered, and custom-endpoint examples;
- added deterministic camera ordering, per-case warm-up, rotated performance rounds, atomic partial reports, state restoration, and explicit fallback diagnostics;
- added schema-v4 benchmark reports containing presented-capture provenance, convergence policy, phase-specific backend state, raw timing samples, and aggregate statistics;
- propagated explicit camera cuts from reset, focus, view-snap, keyframe, and sequencer paths;
- invalidated history when presentation-relevant background state changes;
- release per-view temporal color and depth history after submitted work retires when leaving Temporal, while retaining immutable pipeline state for fast re-entry;
- extracted temporal pipeline request validation into a production CPU-only contract source;
- added a standalone `lichtfeld_temporal_cpu_tests` target covering the production temporal coordinator, planning, motion, resolve, history, quality, and pipeline-validation paths without GPU execution;
- registered the standalone 65-test C++ suite and all 16 benchmark unit tests in Ubuntu Release CI;
- made CI fail explicitly if the temporal CTest registration is missing instead of silently accepting a zero-test run.

Validated locally with:

- 65/65 standalone temporal C++ tests;
- 1/1 temporal CTest registration;
- 16/16 Python benchmark tests;
- repeated production MCP benchmark runs without backend fallback;
- Error Debt, locale completeness, localization contracts, Visual Studio clang-format dry-run, and `git diff --check`.

MrNeRF:
Thank you! Lgtm!

MrNeRF:
Merged into dev as e8a45c108 (squash, co-authored). Round-3 measurements on a 3M-splat scene (4090, 1560x1116, presented image at rest vs Native): Temporal/Quality 48.99 dB vs Spatial 48.48, Balanced 46.90 vs 46.46, Performance 43.47 vs 43.35; GPU busy while orbiting at a fixed frame rate is within about one point of Spatial at every preset.

Changes folded in before the merge:
- motion vectors are now jitter-free (the pipeline builds the motion matrices from unjittered projections; the resolve addresses history from the output grid and adds the previous jitter only for the history depth lookup), which is what DLSS consumes;
- a requested Temporal that fell back to Native no longer presents a jittered raster;
- render.capture keeps the internal raster by default and takes presented=true for the composited crop the benchmark needs;
- axis and gizmo view snaps mark a camera cut;
- the convergence counter is cancelled when no temporal frame was published;
- error-debt baseline regenerated after merging dev.

Thanks Francesco, good work on the rework.

PR #1727, opened 2026-08-20
Keep the viewport stable when toggling the UI

francescofugazzi:
## Summary

Fix viewport stretching and transient deformation when toggling the UI with Hide UI.

Previously, hiding or restoring the sidebar changed the viewport layout before a frame matching the new extent was available. The previously rendered image could therefore be displayed inside a differently sized viewport, briefly stretching or shifting the scene.

## Changes

- stage UI visibility transitions until a frame matching the target viewport extent is ready;
- keep the previous layout and correctly sized frame visible while preparing the new viewport output;
- commit the target layout and matching presentation together;
- distinguish the logical viewport extent from the internal reconstruction resolution used by scene reconstruction backends;
- prevent cached frames from being treated as current when their logical extent does not match the requested viewport;
- render discrete UI layout changes immediately at full resolution;
- retain the existing interactive preview behavior for continuous window and split-divider resizing;
- keep training running while Hide UI transitions are processed;
- preserve the existing pause-and-resume policy for fullscreen transitions;
- safely retain the previous layout if a matching frame is not produced before the transition guard expires;
- add lifecycle regression coverage for interactive and discrete viewport resizing.

## Runtime behavior

Hide UI no longer stretches the existing viewport image while the sidebar is hidden or restored.

During active training, the viewport may briefly retain its previous frame while obtaining a safe model read and preparing the correctly sized output. Training itself remains active and no Hide UI-triggered pause/resume cycle is introduced.

The behavior applies to both single and independent dual-view layouts and remains compatible with reconstruction backends whose internal rendering resolution differs from the logical viewport extent.


src/visualizer/rendering/rendering_manager_vulkan.cpp
```diff
@@ -1735,7 +1741,13 @@ namespace lfs::vis {
         // (exclusive) never stalls the UI frame. On contention, retain the last
         // splat image and retry on the next cadence tick. First frame / no cache
         // falls back to a blocking acquire below.
-        const bool training_try_lock = is_training;
+        // Passive training previews never wait for the step boundary. A discrete
+        // layout transition is different: it must replace the old-size output
+        // before presentation, so wait for one safe read epoch rather than
+        // stretching a cached frame or pausing the trainer.
+        const bool training_try_lock =
```
MrNeRF: This makes the render wait for a safe read of the model instead of trying the lock. On big scenes one step takes several hundred ms, so Hide UI now freezes the UI for that long. Is that acceptable, or should the old frame just stay one more frame and we commit as soon as the next regular render lands?

src/visualizer/gui/gui_manager.cpp
```diff
@@ -4545,14 +4597,26 @@ namespace lfs::vis::gui {
...
         if (now < interactive_transition_guard_until_) {
             return;
         }
 
+        if (ui_visibility_resize_active_) {
+            if (ui_visibility_target_ready_ && !ui_visibility_layout_committed_) {
+                // A failed or unavailable renderer must not leave the UI transition
+                // half-applied. Retain the old, correctly matched layout and allow an
+                // explicit retry instead of stretching its cached image.
+                LOG_WARN("UI visibility transition timed out before a fresh viewport frame; retaining the previous layout");
```
MrNeRF: If no fresh frame arrives within the 1.2 s guard, the toggle is dropped and the user sees nothing happen - only this log line. I'd rather commit the layout anyway at that point (one stretched frame is better than a toggle that silently does nothing), or at least make the button visibly retry.

MrNeRF (commented):
Nice - staging the layout until a frame of the right size exists is the correct fix, and tagging frames with matches_viewport_extent is cleaner than guessing from sizes. Merges cleanly with master. Two things inline about what happens on slow scenes.

francescofugazzi:
Addressed the review points:

- made UI visibility transitions non-blocking during training;
- retained the previous matched viewport frame while the training model is unavailable;
- committed the requested layout after the guard timeout instead of dropping the toggle;
- kept the pending scene render active for the next safe training frame;
- added regression coverage for matching-frame and timeout transitions in both visibility directions.

MrNeRF:
Tested. Works great! Thank you!

PR #1728, opened 2026-08-20
Group File imports and surface rejected formats

francescofugazzi:
## Summary

Reorganize the File menu by grouping all import actions under a dedicated Import submenu and provide clear feedback when an import is rejected or fails before reaching the asynchronous loading backend.

Previously, selecting an unsupported dataset folder or an invalid checkpoint could fail silently because the retained import panel rejected the path without presenting any user-visible explanation.

## File menu organization

The following actions are now grouped under `File > Import`:

- Dataset
- PLY, SOG, SPZ, RAD, or USD splat
- Mesh
- Checkpoint
- Configuration

The Import submenu is positioned directly alongside the existing Export actions, keeping import and export operations together as one logical section of the File menu.

## Import validation and feedback

- validate dataset selections before opening the dataset import panel;
- validate checkpoint headers and parameters before opening the checkpoint import panel;
- show a localized error dialog when the selected folder is not a supported dataset;
- show a localized error dialog when the selected file is not a valid checkpoint;
- record rejected imports as warnings in the application log, including the selected path and rejection reason;
- catch immediate dispatch failures for splat, mesh, and configuration imports and present the failure instead of relying only on operator logging;
- preserve the existing asynchronous error handling for dataset, splat, mesh, and configuration backend failures.

## Checkpoint failure handling

Checkpoint loading could still fail after the initial file validation, for example while validating its associated dataset or restoring training state.

These backend failures are now routed through the existing file-load failure event so they reach the standard GUI error surface instead of remaining visible only in the CLI log.

## Localization

Added localized labels and import error messages for all shipped languages:

- German
- English
- Spanish
- French
- Italian
- Japanese
- Korean
- Dutch
- Polish
- Chinese

## Tests

Added focused coverage for:

- File menu Import submenu structure and ordering;
- grouping Import directly alongside Export;
- rejected dataset selections;
- rejected checkpoint selections;
- warning diagnostics for unrecognized formats;
- visible handling of immediate import dispatch failures.

## Validation

- verified the new File > Import organization in the application;
- verified invalid dataset feedback;
- verified invalid checkpoint feedback;
- verified warning diagnostics in the CLI;
- all standalone localization contracts passed;
- Python and locale JSON syntax validation passed;


src/python/lfs_plugins/file_menu.py
```diff
@@ -19,6 +20,68 @@
...
+        raise _ImportRejected(
+            "dataset format was not recognized",
+            "menu.file.dataset_not_recognized",
+        )
+    if not open_dataset_import_panel(path):
+        raise RuntimeError("dataset import dialog is unavailable")
+
+
+def _open_checkpoint_import_checked(path: str) -> None:
+    if not lf.read_checkpoint_header(path) or not lf.read_checkpoint_params(path):
```
MrNeRF: This opens and parses the checkpoint twice just to answer "is this a checkpoint". On a multi-GB .resume that is a visible stall before the panel even shows. The header alone is enough here - the panel reads the params anyway.

src/python/lfs_plugins/file_menu.py
```diff
@@ -19,6 +20,68 @@
...
+        super().__init__(reason)
+        self.message_key = message_key
+
+
+def _warn_import_failure(path: str, reason: str) -> None:
+    message = f"Import rejected: path='{path}', reason='{reason}'"
+    try:
+        lf.log.warn(message)
+    except Exception:
+        _logger.warning(message)
```
MrNeRF: This fallback never runs inside the app (lf.log.warn is always there), so it only adds a second logging framework to the plugin layer. I'd call lf.log.warn directly and drop the logging import.

MrNeRF (commented):
Looks good - the split between "rejected before the backend" and "failed in the backend" is the right model, and the new dialogs fix a real silent-failure case. Checked that all the bindings and helpers this relies on exist and that it merges cleanly with current master. Two small things inline; the first one I'd fix before merge.

francescofugazzi:
Addressed both review points:

- checkpoint preflight now reads only the lightweight header, leaving parameter parsing to the retained import panel;
- removed the Python logging fallback and routed warnings directly through `lf.log.warn`;
- added regression coverage preventing parameter reads during menu preflight.

PR #1730, opened 2026-08-20
Polish Training panel and Scene Graph selection workflows

francescofugazzi:
## Summary

Polishes the Training panel and Scene Graph to improve space usage, hierarchy readability, filtering, node actions, keyboard navigation, viewport selection, and multi-selection workflows.

The Training panel keeps its primary controls, search field, and active telemetry visible while only the parameter list scrolls. The Scene Graph gains consistent search behavior, optional hierarchy-selection markers, keyboard navigation, compact bulk controls, clearer row actions, camera training-state feedback, safe destructive operations, and explicit duplicate-name rename feedback.

## Training panel

- Keeps Start Training, Clear, search, and context-sensitive controls fixed.
- Scrolls only the training parameter list.
- Keeps active training telemetry visible near the top instead of below the parameter list.
- Shows telemetry from the beginning of dataset preparation and throughout running, paused, stopping, stopped, and completed states.
- Hides telemetry again after Clear or Reset.
- Uses a narrow 4 dp scrollbar with sufficient bottom clearance.
- Fills the available docked-panel height without leaving unnecessary empty space.
- Preserves the color picker and remaining parameter controls inside the scrollable region.
- Removes unsupported RmlUi styling that produced runtime parser warnings.

## Search controls

- Aligns the Training and Scene Graph search fields around the same compact visual structure.
- Uses a single bordered search container with an embedded search icon.
- Adds a compact clear action to both search fields.
- Fixes Scene Graph clearing when the nested X icon receives the click.
- Reads the live RmlUi form-control value during input, clearing, synchronization, and session restoration.
- Prevents the Scene Graph filter from lagging behind the latest entered character.
- Keeps the search controls compact without relying on unsupported placeholder or tooltip workarounds.

## Scene Graph layout

- Aligns the Scene Graph boundary and search field with the tabs and summary chips.
- Removes the redundant internal Models tree header while retaining the model and node summary chips.
- Removes the redundant selected-items summary chip because the contextual action bar reports the active selection count.
- Keeps hierarchy expansion and node type information together on the left.
- Places visibility, training participation, and deletion in stable right-aligned action slots.
- Keeps row and bulk-action columns aligned with or without the vertical scrollbar.
- Removes unnecessary vertical separators and duplicated controls.
- Uses distinct semantic colors for visibility, training state, and deletion.
- Keeps long node names and inline rename fields shrinkable without displacing the action rail.
- Reclaims the marker gutter when optional selection markers are disabled.
- Uses the same compact 4 dp vertical scrollbar as the Training parameter region.
- Removes obsolete zero-height header constants and their dead layout arithmetic.

## Camera training state

- Exposes camera training participation directly in Scene Graph rows.
- Allows individual cameras and camera groups to be enabled or disabled for training.
- Shows partially enabled camera groups with a mixed state.
- Uses a compact `enabled / total` count only when a group is partially enabled.
- Keeps fully enabled groups compact by showing only the total camera count.

## Pointer selection workflow

- Provides optional compact selection markers aligned consistently along the left edge of the tree.
- Keeps marker placement independent from hierarchy indentation.
- Persists marker visibility as a user preference in a dedicated Scene Graph section.
- Supports selection from a single marker without requiring modifier keys.
- Recursively selects or deselects descendants from a parent marker.
- Represents partially selected sections with a mixed state.
- Removes the full-section state when any descendant is deselected.
- Applies selected-row appearance consistently to pointer, marker, hierarchy, and keyboard selection.
- Supports reversible hierarchy selection through Ctrl+Shift+click.
- Clears marker and Scene Manager selection together.
- Prevents rapid clicks on visibility, training, and delete controls from propagating as row double-click activation.

## Viewport selection

- Makes viewport selection modifiers consistent with Scene Graph multi-selection.
- Uses Ctrl+click to add or remove an individual camera or point-picked node.
- Preserves the current selection when Ctrl or Shift is held while clicking empty viewport space.
- Preserves Ctrl/Shift rectangle-selection behavior.
- Latches press-time modifiers through release for point and rectangle gestures.
- Prevents modified viewport clicks from triggering double-click camera navigation.

## Keyboard workflow

The Scene Graph supports the following default bindings. The registered actions can be changed through the existing keymap preferences.

- `Up` / `Down`: move the independent keyboard cursor through visible nodes.
- `Shift+Up` / `Shift+Down`: extend the selection from the fixed anchor.
- `Ctrl+Up` / `Ctrl+Down`: toggle the source node and advance the keyboard cursor.
- `Ctrl+Shift+click`: select or deselect a complete node hierarchy.
- `Ctrl+Shift+A`: select every Scene Graph node.
- `Ctrl+Shift+H`: show or hide the active selection.
- `Ctrl+Shift+T`: enable or disable training participation for compatible selected cameras.
- `Enter`: activate the current node or open its camera image preview.
- `Delete`: request deletion of the active selection through the confirmation flow.
- `Escape`: cancel rename or clear the active selection.

The removed transient `Ctrl+Space` action is not retained.

## Keymap compatibility

- Registers Scene Graph selection actions through the native input system.
- Exposes the actions through the Python keymap bindings and generated stubs.
- Keeps the actions available in the remappable UI keymap section.
- Removes obsolete Scene Graph bindings before installing the finalized defaults.
- Prevents legacy integer action identifiers from being interpreted as different current actions.
- Routes the global Delete action through the Scene Graph whenever a Scene Graph selection exists.

## Contextual bulk actions

A compact action bar appears below the tree whenever one or more nodes are selected.

It provides:

- one clear-selection action beside the selected-node count;
- one state-aware visibility action;
- one state-aware training-participation action;
- one delete action.

The controls adapt to the aggregate selection state:

- the clear action removes the complete active selection;
- the visibility action shows or hides the complete selection;
- the training action enables or disables all compatible selected cameras;
- incompatible training selections disable the training action;
- deletion is disabled unless every selected node can be removed safely;
- mixed removable and non-removable selections cannot be partially deleted;
- icons, tooltips, and colors reflect the action that will be applied.

Each action uses an exact 22 dp border-box hit target with a centered 14 dp icon. The clear action remains beside the selection count, while visibility, training, and deletion remain grouped on the right.

## Safe deletion

- Routes inline, context-menu, keyboard, and bulk deletion through the same confirmation flow.
- Shows a localized confirmation modal before modifying the scene.
- Captures stable node UUIDs instead of relying on names or transient row state.
- Resolves UUIDs back to live node IDs when confirmation is accepted.
- Revalidates that every node still exists and remains removable before deletion.
- Aborts the complete operation if any selected node becomes unavailable or non-removable.
- Uses an ID-based Scene Manager batch operation.
- Validates the complete batch before modifying any node.
- Prevents duplicate names from resolving to or deleting the wrong scene node.
- Preserves undo support for successful batch deletion.

## Rename conflicts

- Detects when an inline rename would duplicate another live scene-node name.
- Shows a localized warning modal instead of failing silently.
- Offers Rename to return to the inline editor with the attempted text preserved.
- Offers Cancel, Escape, or dismissal to exit rename mode while retaining the original node name.
- Prevents focus-loss and change events from queuing duplicate conflict dialogs.
- Records rejected duplicate-name attempts as warnings in the application log.

## Camera image preview

- Preserves the first camera-preview request while the lazily created preview panel is initialized.
- Prevents the initial Enter activation from opening an uninitialized or stale preview.
- Keeps an open camera preview synchronized while navigating between camera nodes.
- Stops synchronization after the preview is closed instead of reopening it automatically.
- Prevents inline Scene Graph action icons from accidentally opening the image preview through rapid double clicks.

## Preferences and persistence

- Adds a dedicated Scene Graph subsection under Interface preferences.
- Persists whether compact Scene Graph selection markers are shown.
- Updates the live Scene Graph when the preference changes.
- Restores the default marker state through the existing preference-reset path.
- Prevents the setting from being mutated in safe mode.

## Localization

Adds localized strings for every supported language covering:

- Scene Graph preferences;
- selection clearing;
- select-all and selection-toggle keymap actions;
- single-node deletion confirmation;
- multi-node deletion confirmation;
- selected-node counts;
- show-selected and hide-selected actions;
- training enable and disable actions;
- duplicate-name rename conflicts and retry behavior.

## Regression coverage

Extends focused native and Python contracts for:

- Training panel fixed and scrollable regions;
- telemetry lifetime;
- Training and Scene Graph search controls;
- live filter-value synchronization;
- Scene Graph alignment and stable action slots;
- optional selection markers and preference persistence;
- recursive and mixed hierarchy selection;
- keyboard action registration and version 24-to-25 keymap migration;
- contextual bulk-action ordering and geometry;
- Delete-key confirmation routing;
- stable UUID and node-ID deletion;
- atomic batch validation and mixed non-removable selections;
- duplicate scene-node names;
- inline-action double-click isolation;
- modifier-aware viewport selection;
- duplicate-name rename feedback;
- compact Training and Scene Graph scrollbars;
- localization completeness;
- first-use camera image-preview initialization.


MrNeRF:
Thank you for the pr. I added a screenshot because the changes cause some issues or rather are looking unfinished:
- It is not clear that the field is a search field
- There is a gap at the bottom and the parameters don't scroll cleanly, i.e. they can be cut half off
- Not sure if the graph should just pop up or simply be there as it takes quite a bit especially if parsing a huge dataset.
 
image

francescofugazzi:
Addressed the review feedback:

- added a compact magnifying-glass indicator so the parameter filter is visually identifiable as a search field;
- removed the ineffective placeholder binding: the project currently uses RmlUi 6.2, while native text-input placeholders were introduced in RmlUi 6.3 ([RmlUi changelog — Text input placeholder](https://github.com/mikke89/RmlUi/blob/master/changelog.md#text-input-placeholder));
- replaced the hidden parameter scrollbar with a minimal 4 dp scrollbar and added bottom spacing so the final parameter row can scroll fully into view;
- made the telemetry region visible as soon as training enters the running state, including dataset preparation at iteration zero;
- kept the telemetry region stable throughout running, paused, stopping, stopped, and completed states, preventing it from appearing late and abruptly resizing the parameter list;
- preserved the existing behavior that hides telemetry again after Clear or Reset;
- expanded the regression coverage for the search indicator, scroll-region layout, and telemetry lifetime.

MrNeRF:
rmlui 6.3 available? Lemme check if we can pull that it already.

MrNeRF:
Unfortunately the last changes even trashed the layout more. Please see the images
Screenshot from 2026-08-21 21-48-48
Screenshot from 2026-08-21 21-48-31

The search bar is also different than in the scene graph at the top.

francescofugazzi:
Addressed the remaining Training panel and Scene Graph usability issues:

- made the Training panel fill its available height while keeping controls, search, and active telemetry fixed;
- limited scrolling to the parameter area and retained a narrow scrollbar with bottom clearance;
- unified the Training and Scene Graph search presentation and fixed both clear actions;
- aligned the Scene Graph view with its tabs and summary row;
- removed the redundant internal Models header while preserving the model and node summary chips;
- removed the redundant selected-items summary chip;
- reorganized row controls into stable visibility, training, and delete slots without vertical separators;
- exposed individual and mixed camera training states with compact enabled/total feedback;
- added an always-visible, left-aligned checkbox gutter independent from hierarchy indentation;
- added recursive section selection and tri-state parent feedback;
- added a compact contextual bar for single and multiple selections with one state-aware visibility action, one training action, and deletion;
- routed inline, context-menu, keyboard, and bulk deletion through localized confirmation using stable UUIDs and live eligibility revalidation;
- added the required strings to all supported locales;
- extended the focused resource, localization, selection, layout, and deletion regression coverage.

francescofugazzi:
Extended and finalized the Scene Graph workflow:

- made compact hierarchy-selection markers optional through a persisted Scene Graph preference;
- moved the marker option into a dedicated Scene Graph section in Preferences;
- added an independent keyboard cursor that can move without replacing the active selection;
- added Shift+arrow for range selection;
- added Ctrl+arrow to toggle the source node before advancing the keyboard cursor;
- added reversible hierarchy selection through Ctrl+Shift+click;
- added Ctrl+Shift+A to select all Scene Graph nodes;
- added Ctrl+Shift+H to toggle visibility for the active selection;
- added Ctrl+Shift+T to toggle training participation for compatible selected cameras;
- kept Enter as the activation key for camera previews and other supported nodes;
- removed the superseded Ctrl+Space action;
- registered the new actions in the remappable keymap and exposed them through the Python bindings and stubs;
- added a clear-selection action beside the selected-node count;
- corrected the bulk-action geometry so every icon is centered inside an exact 22 dp hit target;
- kept visibility, training, and deletion actions grouped on the right;
- reduced the Scene Graph scrollbar to the same compact 4 dp width used by the Training parameter region;
- synchronized an open camera preview with Scene Graph cursor navigation without reopening a closed preview;
- preserved the first camera-preview request during lazy preview-panel creation;
- added localized labels for the new preference, actions, and selection controls in every supported language;
- extended native and Python regression coverage for keymap migration, selection controls, action-bar geometry, compact scrolling, localization, preferences, and first-use image-preview initialization.

src/visualizer/gui/rmlui/elements/scene_graph_element.cpp
```diff
@@ -2041,29 +2244,167 @@ namespace lfs::vis::gui {
...
+                if (!live_node)
+                    continue;
+                if (const auto can_remove = live_manager->canRemoveNode(live_id); !can_remove)
+                    continue;
+                live_names.push_back(live_node->name);
+            }
+            if (live_names.empty())
+                return;
+            if (const auto delete_result =
+                    live_manager->removeNodesWithResult(live_names, /*keep_children=*/false);
```
MrNeRF: You go to the trouble of capturing UUIDs and resolving them again, then delete by **name**: `removeNodesWithResult(names)` resolves through `getNodeIdByName`, first match. Two nodes with the same name (easy with imported PLYs) and the wrong one goes. `removeNodeWithResult(NodeId)` exists; use the ids you just resolved (an id-based batch overload if you want to keep the planned-removal behaviour of the name variant).

src/visualizer/gui/rmlui/elements/scene_graph_element.cpp
```diff
@@ -2041,29 +2244,167 @@ namespace lfs::vis::gui {
...
+                return;
+            auto& live_scene = live_manager->getScene();
+            std::vector<std::string> live_names;
+            live_names.reserve(node_uuids.size());
+            for (const core::Uuid& uuid : node_uuids) {
+                const core::NodeId live_id = live_scene.getNodeIdByUuid(uuid);
+                const auto* live_node = live_scene.getNodeById(live_id);
+                if (!live_node)
+                    continue;
+                if (const auto can_remove = live_manager->canRemoveNode(live_id); !can_remove)
```
MrNeRF: A node that stopped being removable between the request and the confirmation is skipped here and the rest is deleted, which is the partial bulk deletion the description says is avoided. Either abort the whole batch when any node fails this check, or adjust the text.

src/visualizer/input/input_controller.cpp
```diff
@@ -1712,6 +1712,21 @@ namespace lfs::vis {
                     gui->openPreferences();
                 return;
 
+            case input::Action::SELECT_ALL_SCENE_NODES:
```
MrNeRF, line 1744: These three are gated on focus, good. `DELETE_NODE` further down (~1875) is not, and it still removes the selected nodes directly, which is why the Delete key skips the new confirmation (see the review summary).

src/visualizer/gui/rmlui/elements/scene_graph_element.hpp
```diff
@@ -237,17 +280,14 @@ namespace lfs::vis::gui {
...
-        std::string last_header_text_;
         std::vector<std::string> row_top_dp_cache_;
-        bool last_header_visible_ = false;
-        bool last_header_expanded_ = true;
         float panel_screen_x_ = 0.0f;
         float panel_screen_y_ = 0.0f;
 
         static constexpr int kRowHeightDpInt = 20;
-        static constexpr int kHeaderHeightDpInt = 24;
+        static constexpr int kHeaderHeightDpInt = 0;
```
MrNeRF: Both header-height constants are now always 0. Drop them and the arithmetic that uses them rather than carrying a dead layout knob.

MrNeRF (changes requested):
Thanks Francesco, this is a big quality-of-life step. I built the branch, ran the C++ and Python suites (117 / 585 / 109 green) and drove it for a while with garden and a training run. The sticky Training panel with the telemetry above the parameter list, the action rail, the mixed training state on camera groups (I got "Training (182 / 185)" from a Ctrl+Shift+T on three cameras), the bulk bar and the delete confirmation from the trash icon and the bulk bar all behave exactly as described. Cancel keeps, Delete removes, Undo brings it back. Nice.

One blocker, and it is the one thing the description promises most explicitly: **the Delete key does not go through the confirmation.** With a Scene Graph row focused, Delete removed the selection instantly, no modal. I reproduced it twice, once on three training cameras during a paused run (whose inline trash icon this PR correctly disables) and once on a plain PLY node. The old global `Action::DELETE_NODE` binding (`input_controller.cpp` ~1875, Delete in GLOBAL/transform modes) still emits `cmd::RemovePLY` for every selected node name and wins over the new `KI_DELETE` handler. So the key bypasses both the confirmation and the "cannot remove safely" guard. Route that action through the Scene Graph's request path when the tree has focus, the way the new `*IfFocused` actions do, or take the direct removal out of it.

Two smaller things in the deletion path, inline. Everything else I looked at (keymap migration, focus gating of the new shortcuts, preferences and safe mode, locales) is fine.

One observation I could not pin down: typing "07962" into the Scene Graph search once left the chip at `Filter: "0796"` with the list unfiltered until the next keystroke. Subsequent edits tracked the field exactly. Worth a look at whether the filter reads the value before or after the last text event.

francescofugazzi:
Addressed the remaining review findings and related interaction issues:

- routed the global Delete action through the Scene Graph confirmation path whenever a Scene Graph selection exists;
- made bulk deletion resolve stable UUIDs back to live node IDs instead of deleting by name;
- made deletion validation atomic before confirmation, after confirmation, and inside the Scene Manager batch operation;
- prevented mixed removable and non-removable selections from being partially deleted;
- removed the obsolete zero-height Scene Graph header constants and associated layout arithmetic;
- made Scene Graph filtering consistently read the live RmlUi form-control value;
- prevented rapid clicks on visibility, training, and delete icons from triggering row double-click activation;
- kept row and bulk-action columns stable with or without the compact scrollbar;
- added modifier-aware viewport selection: Ctrl+click toggles camera or point-picked nodes;
- preserved Ctrl/Shift rectangle selection and excluded modified clicks from double-click navigation;
- added explicit duplicate-name rename feedback with a localized warning modal, retry/cancel behavior, preserved edit text, and a console warning;
- extended native and Python regression coverage for stable-ID deletion, atomic validation, filtering, inline-action isolation, viewport modifiers, and rename conflicts.

MrNeRF:
Thank you for the improvements! All good!

PR #1731, opened 2026-08-20
Preserve calibrated camera frustums during training

francescofugazzi:
## Summary

Prevent camera frustums from shrinking or narrowing when training starts with downsampled images.

## Problem

Camera frustum geometry was derived from the mutable training image dimensions while continuing to use the original camera focal length.

When `resize_factor` or `max_width` reduced the operational image resolution, the resulting field of view became artificially narrower even though the underlying camera calibration had not changed.

This affected the displayed frustum as soon as training initialized the resized image dimensions.

## Changes

- derive camera frustum geometry from the calibrated camera width, height, and vertical field of view;
- keep frustum geometry independent of training image downscaling;
- preserve rectified calibration produced by camera undistortion;
- apply the same calibration contract to:
  - viewport frustum rendering;
  - software frustum picking;
  - scene selection bounds;
- retain the existing equirectangular camera representation.

## Regression coverage

Added regression checks ensuring that frustum rendering, picking, and selection:

- use calibrated camera dimensions and `FoVy`;
- do not depend on mutable training image dimensions;
- do not reconstruct the field of view from an unscaled focal length.

## Validation

Manually verified that camera frustums retain their original shape and size when training starts with scaled images.

### DIFF
--- a/src/rendering/raster_rendering_engine.cpp
+++ b/src/rendering/raster_rendering_engine.cpp
@@ -308,9 +308,13 @@ namespace lfs::rendering {
             const glm::mat4& visualizer_camera_to_world,
             const float scale) {
             std::vector<glm::vec3> points;
-            const int image_width = camera.image_width() > 0 ? camera.image_width() : camera.camera_width();
-            const int image_height = camera.image_height() > 0 ? camera.image_height() : camera.camera_height();
-            if (image_width <= 0 || image_height <= 0 || scale <= 0.0f) {
+            // Picking must follow the calibrated frustum, not the resolution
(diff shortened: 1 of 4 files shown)


MrNeRF (approved):
Good fix - the frustum should follow the calibrated camera, not the training image size, and reading FoVy directly is cleaner than recomputing it. I checked the one thing that could bite: FoVy is computed in the Camera constructor from focal_y and the calibrated height and refreshed on undistortion, so it is valid for every loader exactly when the old focal check passed. No gap. Approving.

PR #1735, opened 2026-08-20
Fix Python UI callback teardown ordering

francescofugazzi:
## Summary

Fix application shutdown ordering so pending Python UI callbacks are drained while the embedded Python runtime and GIL are still available.

## Changes

- destroy the interactive viewer and its GUI resources before finalizing Python;
- allow `GuiManager` to drain scheduled Python UI callbacks during teardown;
- preserve the existing MCP shutdown, GPU teardown, and clean-exit ordering;
- prevent intermittent `Unable to run scheduled Python UI callback: Python GIL is unavailable` errors when closing the application, particularly during or shortly after training.

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -1310,9 +1310,12 @@ namespace lfs::app {
             mcp::setActiveMcpHttpServer(nullptr);
             mcp_http.stop();
 
-            python::finalize();
-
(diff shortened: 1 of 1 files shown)


MrNeRF:
I think we had the reordering a couple of times. It comes back over and over again. Something more structural is into play here.

Issue #1516, opened 2026-07-31
Drag-and-drop architecture for RmlUi panels (asset browser → viewport)


MrNeRF:
Part of #1505. Blocks the asset browser feeling native; should ride the same milestone as #1508.

Dragging an asset into the viewport is the defining gesture of every asset browser users know — and it is structurally impossible today: each RmlUi panel owns a separate `Rml::Context`, so cross-panel DOM drag cannot work; the native drag-source Python APIs are stubs hard-coded to return `false`; and a file dropped *onto the asset panel* is routed by extension straight into the scene, bypassing the catalog.

**Scope**
- [ ] Architecture decision: shared drag overlay layer vs cross-context payload bus (C++).
- [ ] Drag source: asset card → viewport drop = load at drop location; → scene tree = add under node.
- [ ] Drop target: files dropped onto the panel register in the catalog (and route through the existing project-switch dirty preflight where applicable).
- [ ] Expose to Python so plugin panels can participate.

Issue #1511, opened 2026-07-31
Asset browser: global search, recents, keyboard navigation, selection basics


MrNeRF:
Part of #1505.

Search is fuzzy-name-only and scoped to the selected folder; there is no "All Assets" or "Recent" view, so "where is that bicycle ply" has no answer across 20 folders. `AssetIndex.search_assets` already matches name+path+tags — it is dead code. There is zero keyboard interaction (no arrows, Enter, Delete, type-ahead), no shift-range select, and middle-click opens the context menu.

**Scope**
- [ ] Global "All Assets" + "Recent" scopes; search across name/path/type/tags (wire the existing dead function).
- [ ] Keyboard: arrows, Enter=load, Delete, type-ahead.
- [ ] Shift-range multi-select (Ctrl-toggle exists).
- [ ] Fix middle-click-opens-context-menu.
- [ ] Persist UI state across restarts (view/sort/filters/selected folder/sidebar geometry); un-collapse the folder tree by default.
- [ ] Hoist per-row context menus to one shared menu (each rendered row currently instantiates a full folder-list dropdown — hundreds to thousands of hidden DOM nodes).

Issue #1505, opened 2026-07-31
Asset manager overhaul — tracking issue


MrNeRF:
The asset manager needs a ground-up identity fix, not cosmetic polish. Four-way analysis (our implementation vs Blender vs Godot/Unreal/Unity/USD vs domain needs) concluded:

**Today**: ~11.9k lines of Python asset manager, registered **disabled by default**. Good virtualized UI and atomic persistence — but it is a catalog of absolute-path bookmarks: no content hashing, no drag-and-drop (structurally blocked in C++), O(N²) catalog operations, and zero integration with the project format. Meanwhile the .licht project format (in development) already implements the correct asset model — stable UUIDs, content fingerprints with typed relink dispositions, commit lineage, embedded previews with a foreign-reader contract — and nothing outside the project file reads it.

**Strategy**: keep the panel UI, replace its data model with project-format identity, and start stamping exports immediately (every unstamped export is permanently unattributable — the one deadline item).

**Recorded decisions**
- No Blender-style "editable linked assets"/library overrides. Embed-or-read-only.
- Caches accelerate, never decide: no cache resolves a reference; the container is the only authority.
- Import modes: Reference / Embed / Embed-with-identity, named honestly.

## Children

**Foundations**
- [ ] #1506 identity foundations (fingerprint/relink bindings + catalog adoption)
- [ ] #1507 export provenance stamps ← **do first, cannot be retrofitted**

**Core rework**
- [ ] #1508 one asset model (catalog rebased onto project-format identity)
- [ ] #1509 relink flow + durable project index
- [ ] #1516 drag-and-drop architecture (C++)

**Panel quality**
- [ ] #1510 O(N²) catalog performance
- [ ] #1511 search/keyboard/selection
- [ ] #1512 thumbnail eviction + encoding bug
- [ ] #1513 error surfacing + confirmations
- [ ] #1514 feature debt (tags, watching, type drift)
- [ ] #1515 storage panel (cache visibility)

**Platform/infra**
- [ ] #1517 OS-level project thumbnails
- [ ] #1518 browsable metadata at known offsets
- [ ] #1519 cache discipline + resolver

**Game changers**
- [ ] #1520 lineage graph
- [ ] #1521 machine-wide content index
- [ ] #1522 splat-native live previews
- [ ] #1523 capture-session organization
- [ ] #1524 publish round-trip

[...]

Issue #1524, opened 2026-07-31
Publish round-trip with durable identity


MrNeRF:
Part of #1505. Depends on #1507; sequence after the other game-changers (external-service dependency).

Publishing should be an edge in the lineage graph, not a fire-and-forget upload. Today the prototype publish flow identifies uploads by sanitized filename — re-uploading silently overwrites and nothing links back to the producing project.

**Scope**
- [ ] Published entries carry asset UUID + content fingerprint + provenance stamp.
- [ ] Local library shows publish state and staleness vs the current commit.
- [ ] Re-publish updates the same entry (identity by UUID, verified by hash) instead of duplicating.
- [ ] "Open in studio" from a shared link resolves to the exact project commit, relinking the dataset by content (#1521) when present locally.
- [ ] Stamps strippable before upload (privacy: dataset names are often client names).

PR #1770, opened 2026-08-21
Make the Asset Manager native to .licht projects

### LINKED ISSUE #1506: Asset identity foundations: expose content-fingerprint/relink API to Python and use it in the catalog
Part of #1505.

The asset catalog (`src/python/lfs_plugins/asset_index.py`) has no content identity: integrity is `exists: bool` + `file_size_bytes`, duplicate detection is a linear path scan, and a moved file is simply "missing". Meanwhile the C++ side of the .licht project format (in development) already implements content fingerprints (xxh3 head+tail windows with an mtime fast path and typed verification dispositions) and a relink API — none of it is bound to Python.

**Scope**
- [ ] Bind `fingerprint_path`, `check_fingerprint`, `ReferencesChapter::verify_and_refresh`, `relink` into the `lichtfeld` Python module (additive API + stub updates).
- [ ] Bind `lfs::core::user_config_dir()` and delete the two duplicated storage-resolution implementations in the plugin (`asset_index.py` and `asset_manager_panel.py` both re-implement platform dirs).
- [ ] Catalog uses fingerprints as identity: verify-on-access with the typed dispositions instead of `exists`.
- [ ] Duplicate detection on import: same content ⇒ same asset ("this PLY is already catalogued as X"), replacing the O(N) `find_asset_by_path` scan.

**Why first**: nearly every other asset-manager issue depends on content identity. Engines precedent: Godot and Unity both retrofitted content hashes after shipping path/mtime identity and paying for it.

### LINKED ISSUE #1512: Thumbnail pipeline: eviction, path-encoding bug, failure surfacing
Part of #1505.

Three defects in the thumbnail path:

1. **The cache never evicts.** `cleanup_orphans`/`clear_all` exist with zero production callers; `~/.lichtfeld/asset_manager/thumbnails/` grows monotonically, including thumbnails of deleted assets.
2. **Bug: percent-encoding mismatch blanks thumbnails.** The panel percent-encodes absolute POSIX paths, but the native RmlUi path decoder returns early for absolute paths *without* decoding — a home directory with a space or non-ASCII character (e.g. `/home/josé/…`) silently renders blank cards. Windows additionally diverges (`file:///` vs encoded `C:/`).
3. Failures are memoised silently; no user-visible signal.

**Scope**
- [ ] Wire orphan cleanup + size cap/LRU for the thumbnail directory.
- [ ] Fix the encoding contract between the panel and the native decoder (round-trip test with a space + non-ASCII path, both platforms).
- [ ] Surface generation failures via the existing toast system.

shadygm:
Closes #1506.
Closes #1512.

Part of #1505. Includes the scoped Asset Manager work from #1511 and #1516.

## What changed

- Made the Asset Manager `.licht`-only and keyed projects by their persistent project UUID; paths are now locators instead of identity.
- Added native Python APIs for project inspection, fingerprints, references/relinking, and the application-owned Asset Library directory.
- Reduced `library.json` to folder/project organization and location data, with migration, atomic backup, UUID deduplication, and stale-entry cleanup.
- Restored recursive watched-directory discovery with overlap handling and `.licht` validation.
- Switched gallery thumbnails to the embedded `.licht` preview and removed the legacy thumbnail cache/render queue.
- Removed obsolete multi-format import paths, cached metadata, type badges, tags, collections, notes, Recent assets, and other dead catalog/UI code.
- Added a separate **All Assets** scope with global name/path search, default-expanded folders, and clearer folder navigation.
- Added keyboard navigation: arrow movement, Enter to load, Delete, and type-ahead focus for search.
- Consolidated row/card/folder actions into the shared native context menu and kept project loading on the standard dirty/training confirmation path.
- Added cross-context drag payload support and Python bindings so Asset Manager projects can be dropped onto the viewport to open them.
- Kept the Asset Manager open through viewport-drop loading, including dirty-project and active-training confirmations.
- Streamlined the UI with icon-only gallery/list controls, one refresh-and-clean action, consistent button sizing, aligned list columns, and less redundant metadata.


shadygm:
## #1505 child-issue audit

### Completed or obsolete

- **Completed — #1506:** this PR exposes the native fingerprint, reference verification/relinking, inspection, and application-owned storage APIs. The catalog uses the stronger persistent `.licht` project UUID as identity, with paths as locators and UUID-based deduplication.
- **Completed — #1507:** the export provenance work was completed independently and the issue is already closed.
- **Completed — #1516:** this PR implements the cross-context C++ drag-payload bus, exposes it to Python panels, and supports dragging Asset Manager projects onto the viewport to load them through the normal project-open lifecycle.
- **Obsolete / resolved by removal — #1512:** this PR deletes the standalone thumbnail generator and unbounded disk cache. Gallery previews come from the embedded `.licht` THMB, percent-encoded paths are decoded natively, and the old generation-failure state no longer exists.

### Keep open

- **Partially addressed:** #1508, #1509, #1510, #1511, #1513, #1514, #1515, and #1519.
- **Not implemented:** #1517, #1518, #1520, #1521, #1522, #1523, #1524, and #1771.
- **Tracker remains open:** #1505.

**Reviewer note:** please independently re-verify every assessment against the implementation and each issue's current acceptance criteria before closing issues or updating the tracker.

src/python/lfs_plugins/asset_index.py
```diff
@@ -9,1355 +11,749 @@
...
+            is_v2 = data.get("schema_version") == SCHEMA_VERSION
+            if is_v2 and not migrating_legacy_location:
+                if self._load_v2(data) and not self.save():
+                    return False
+                self._cleanup_obsolete_storage()
+            else:
+                self._migrate_legacy(data)
+                if not self.save():
+                    return False
+                self._refresh_backup()
```
MrNeRF: `save()` two lines up already copied the legacy `library.json` to `.bak`; this then copies the freshly written v2 file over it, so the only pre-migration copy is destroyed at the exact moment the lossy one-shot migration ran (and `_cleanup_obsolete_storage()` removes the legacy directory right after). Master's migration test pinned `.bak` still holding the original bytes — deleted here rather than kept green. Drop this call in the migration path, or back the legacy file up under a distinct name before `save()` (see the review summary).

src/python/lfs_plugins/asset_index.py
```diff
@@ -9,1355 +11,749 @@
...
+                )
+        self._ensure_default_folder()
 
-            # Load scenes
-            self._scenes = {
-                sid: Scene.from_dict(s) for sid, s in data.get("scenes", {}).items()
-            }
+        legacy_projects = data.get("projects")
+        if not isinstance(legacy_projects, dict):
+            legacy_projects = data.get("assets", {})
```
MrNeRF: A pre-#1265 `library.json` ships `projects` as the *folder* map with the real entries under `assets` keyed by `project_id` — master migrates that shape explicitly, shape-based, because those files also claim version "1.0.0". Here a dict-valued `projects` wins, the folder records have no path so every candidate is skipped, and `assets` is never read: the migration produces an empty catalog (see the review summary). Needs the folders-absent → projects-are-folders mapping and `project_id` → `folder_id`.

src/python/lfs_plugins/asset_index.py
```diff
@@ -9,1355 +11,749 @@
...
-
-            self._version = data.get("version", LIBRARY_VERSION)
-            self._created_at = data.get("created_at", datetime.now().isoformat())
-            self._modified_at = data.get("modified_at", datetime.now().isoformat())
-
-            # Load folders
-            self._folders = {
-                fid: Folder.from_dict(f) for fid, f in data.get("folders", {}).items()
-            }
+        self._folders = folders
```
MrNeRF: `_load_v2` assigns folders here and inserts projects one at a time, then any bad row — duplicate path, non-canonical UUID key, non-.licht path — raises out of the whole load. `load()` returns False but the half-loaded state stays, `on_mount` ignores the False, and the next user action `save()`s the truncated catalog, rotating `.bak` on the way. Build replacement dicts and swap at the end (master's approach), or tolerate bad rows per-record instead of failing the file.

src/python/lfs_plugins/asset_index.py
```diff
@@ -9,1355 +11,749 @@
...
-                a.folder_id == asset_folder_id for a in self._assets.values()
-            )
-            if not folder_has_scenes and not folder_has_assets:
-                del self._folders[asset_folder_id]
+            self._projects[project_uuid] = project
+            self._project_by_path[path_key] = project_uuid
+            self._apply_inspection(project, inspection)
+        else:
+            stored_path_missing = not Path(project.path).is_file()
+            use_observed_path = adopt_existing or stored_path_missing or self._path_key(project.path) == path_key
```
MrNeRF: `stored_path_missing` makes `adopt_existing=False` adopt anyway: the catalog points at `/mnt/nas/proj.licht`, the NAS is briefly down, a watch root holds an old copy with the same UUID — the startup scan rewrites the locator to the backup copy, and remounting does not undo it. A scan that was told not to adopt should not move a locator because the stored path failed one `is_file()` probe. `test_watched_scan_does_not_replace_a_live_explicit_locator` never registers an explicit locator first, so it does not catch this.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -118,5724 +81,1495 @@ class AssetManagerPanel(Panel):
...
+            if not asset.get("exists", False)
+        ]
+        removed = self._asset_index.delete_assets(missing) if missing else 0
+        if removed:
+            self._selected_asset_ids.difference_update(missing)
+            if self._selection_cursor_id in missing:
+                self._selection_cursor_id = None
+            self.refresh_catalog(scan_watched=False)
+
+    def refresh_and_clean(self, _handle=None, _ev=None, _args=None):
```
MrNeRF: This is the only refresh the toolbar offers, and the clean half deletes every entry whose file did not answer `is_file()` at that moment. Unplugged drive plus one reflexive refresh click and those entries are gone from `library.json` with no prompt — reproduced, see the review summary. Split refresh from clean, or have clean show what it is about to drop.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -118,5724 +81,1495 @@ class AssetManagerPanel(Panel):
...
+        asset_id = element.get_attribute("data-asset-id", "")
+        if not asset_id or not self._asset_index:
+            return
+        verify_asset = getattr(self._asset_index, "verify_asset", None)
+        project = verify_asset(asset_id) if callable(verify_asset) else None
+        if callable(verify_asset) and project is None:
+            self.refresh_catalog(scan_watched=False)
+            return
+        asset = (
+
```
MrNeRF, line 1410: `set_search_query` never clears or clamps the selection, and this takes `_selection_cursor_id` without checking it against `_filtered_assets()` — filter, focus the results, hit Enter, and the *pre-filter* selection loads. The Delete branch below is worse: `_delete_selected_assets` drops the whole hidden multi-selection while the list shows something else. Clamp both to the filtered rows the way `_navigate_selection` already does.

MrNeRF, line 1424: Focusing the input does not retroactively deliver this keydown, so the first character of a type-ahead is always swallowed — typing "garden" filters on "arden". Append the character to `search_query` (or re-dispatch the key) when stealing focus; the test only asserts `focused is True`.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -118,5724 +81,1495 @@ class AssetManagerPanel(Panel):
...
             tr("asset_manager.dialog.rename_folder"),
-            tr("asset_manager.dialog.enter_new_name", name=current_name),
-            current_name,
-            _on_rename_result
+            tr("asset_manager.dialog.enter_new_name", name=current),
+            current,
+            rename,
         )
 
     def on_delete_folder(self, _handle, _ev, args):
```
MrNeRF: No confirmation, and `delete_folder` drops every project entry in the folder along with it — a misclick on a populated folder silently empties that slice of the catalog. Same argument as the Scene Graph delete confirmation in #1730; wants the `confirm_dialog` treatment with the asset count in the message.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -118,5724 +81,1495 @@ class AssetManagerPanel(Panel):
...
 
     @staticmethod
-    def _sort_text(value: Any) -> str:
-        return str(value or "").lower()
+    def _thumbnail_decorator(asset: Dict[str, Any]) -> str:
+        if not asset.get("has_preview") or not asset.get("exists"):
+            return "none"
+        path = quote(str(asset.get("path") or ""), safe=_RML_PATH_SAFE_CHARS)
+        revision = quote(str(asset.get("commit_uuid") or ""), safe="-._~")
+        return f"image(preview://kind=licht&thumb=256&rev={revision}&path={path})"
```
MrNeRF, line 453: `rev=` only changes when the in-memory row's `commit_uuid` changes, and nothing re-verifies rows on project save — so with the panel kept open across a drop-load (the new flow), a re-saved project keeps showing its old THMB until a manual refresh. It is also a no-op for a project with an empty `commit_uuid`, and textures for stale revs are never released. Re-verify at least the current project's row after a save.

src/python/lfs_plugins/asset_watch.py
```diff
@@ -0,0 +1,142 @@
...
+    if not root.is_dir():
+        _log.warning("Watched directory is unavailable: %s", root)
+        return []
+
+    projects: list[str] = []
+
+    def _on_error(exc: OSError) -> None:
+        _log.warning("Could not scan watched directory: %s", exc)
+
+    for current_root, directory_names, filenames in os.walk(
```
MrNeRF, line 57: No directory pruning: master's scanner skipped `sparse`/`dense`/`masks`/`stereo`/`depth`/`__pycache__`, and watch roots are commonly COLMAP dataset parents, so this now walks every `images/` tree on each refresh. Every hit then goes through `register_licht_asset` → `lf.io.inspect_project` even when already cataloged, each call under the process-wide index lock — and the bare `path.resolve()` below lets one `OSError` abort the entire scan. Prune the known dataset dirs, guard `resolve()`, and skip inspection for unchanged cataloged paths.

src/python/lfs_plugins/watch_dirs_panel.py
```diff
@@ -0,0 +1,287 @@
...
+        try:
+            index = int(args[0])
+        except (TypeError, ValueError):
+            return
+        if 0 <= index < len(self._watch_dirs):
+            self._watch_dirs.pop(index)
+            self._reset_scan_state()
+            self._dirty_model()
+
+    def _on_cancel(self, _handle=None, _ev=None, _args=None) -> None:
```
MrNeRF: Cancel only hides the panel now — master's dialog had a cancel event and joined the worker. The daemon thread keeps registering and `save()`ing after the user cancels, and `show()` for another folder while a scan runs starts a second concurrent worker (the generation bumps, nobody joins). `_on_save` also scans *all* folders' roots via `scan_all_watch_directories` instead of the folder being edited.

src/python/lfs_plugins/file_menu.py
```diff
@@ -93,12 +92,35 @@ def _on_result(button):
...
+    if keep_asset_manager_open:
+        return lf.project_open(
+            path, discard_changes, stop_training, True
+        )
     if stop_training:
         return lf.project_open(path, discard_changes, True)
     return lf.project_open(path, discard_changes)
 
 
+def open_project_with_confirmation(path: str) -> None:
```
MrNeRF: Every in-panel load — Enter, double-click, the Load button, context-menu Open — lands here with `keep_asset_manager_open` effectively False, so loading from the catalog closes the catalog; only the viewport drop keeps it open (verified in the running app). If that asymmetry is intended, fine; if the panel is meant to survive browsing, thread the flag through here. The panel tests stub `project_open` with three arguments, so they cannot see it either way.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -118,5724 +81,1495 @@ class AssetManagerPanel(Panel):
...
+        if "available" in asset:
+            return bool(asset.get("available"))
+        return bool(asset.get("exists", True))
+
+    def _project_status_label(self, asset: Dict[str, Any]) -> str:
+        status = str(asset.get("status") or "UNVERIFIED")
+        key = {
+            "AVAILABLE": "asset_manager.status.available",
+            "MISSING": "asset_manager.status.missing",
+            "IDENTITY_MISMATCH": "asset_manager.status.type_changed",
```
MrNeRF: A UUID mismatch is not a type change — this renders as "File type changed". Wants its own status string; "The file at this path belongs to a different project" already exists as the error text in `asset_index`.

MrNeRF (changes requested):
Thanks Shady, this is the direction I wanted the Asset Manager to take — .licht-only with the persistent project UUID as identity and paths demoted to locators — and the interaction work holds up. I built the branch, ran the suites (88 Python tests across the touched modules, 3 new C++ drag-payload tests, all green) and drove it for a while: a recursive scan picked up four projects across three directory levels including a `name with spaces & ünïcode.licht` ("Found 4, added 4"), the gallery rendered the real embedded THMB previews, arrows + Enter loaded through the standard confirmation, Delete removed only the catalog entry with the file untouched, and dragging a card onto the viewport showed the "Drop to open project" overlay, went through the dirty-project modal, loaded the project and kept the Asset Manager open. All exactly as described.

Three blockers, all in catalog persistence, all reproduced by running the code rather than reading it:

1. **Migration destroys its own backup.** In `load()`, `save()` correctly copies the legacy `library.json` to `.bak` — then `_refresh_backup()` copies the freshly written v2 file over it, and `_cleanup_obsolete_storage()` removes the legacy directory too. The migration is intentionally lossy (tags, collections, notes, every non-licht entry), so after first launch the old catalog is unrecoverable. Master's `test_asset_index_migration.py` pinned `.bak` still holding the original bytes; this PR deletes that test instead of keeping it green. I ran it: `.bak` comes out as schema-v2 with the legacy `tags` gone.

2. **Pre-#1265 catalogs migrate to an empty library.** In that still-supported shape `projects` is the *folder* map and the entries live under `assets` keyed by `project_id` — master migrates it shape-based since those files also claim version "1.0.0". `_migrate_legacy` sees a dict-valued `projects`, treats it as the project map (folder records have no path, so everything is skipped) and never reads `assets`. I ran it: folder name lost, zero projects — and with blocker 1 the backup is gone as well.

3. **The refresh button deletes entries for temporarily missing files.** The toolbar's only refresh is `refresh_and_clean`. I moved a cataloged `.licht` away to simulate an unplugged drive and clicked refresh: the entry was permanently purged from `library.json`, no prompt, and the next save rotates the `.bak` away. The v2 load path deliberately keeps `MISSING` entries, so the model supports offline volumes — the button defeats it. The one-shot migration drops offline entries the same way (and `test_legacy_catalog_migration_keeps_only_names_paths_folders_and_watch_roots` pins that as expected, which I think is the wrong pin).

The rest is inline: `_load_v2` leaving a half-loaded index on one bad row, a watch-scan path that can rewrite a live locator, Enter/Delete acting on the pre-filter selection, type-ahead eating the first character, unconfirmed folder delete, thumbnails that stay stale after a resave, the unpruned watch walk, and the scan the dialog can no longer cancel. Two smaller ones with no inline anchor: a failed `save()` leaves dirty in-memory state in `register_licht_asset` / `delete_folder` / `relink_asset` / `update_folder` (`create_folder` ignores the result entirely — only `set_watch_dirs` and `delete_assets` roll back), and the APPDATA / unwritable-home storage fallbacks went away with their tests, so Windows catalogs that only ever lived there will not migrate.

What I checked and found fine: the C++ drag-payload bus (token lifecycle, replacement invalidation, cancel on Escape/modal/focus loss, drop resolved twice per frame so a release cannot slip a frame, overlay excluded from hit testing), preview URL percent-encoding including `&` in filenames, context menus binding UUIDs rather than row indexes, drag-token cancel on unmount and re-begin, every load path going through the dirty/training confirmation, stubs matching the new bindings, and no dangling references anywhere to the deleted scanner/thumbnail/URL-import/SaveAsset machinery.

One design question, not blocking: repair-only and unsupported-newer containers get `available=False`, so a project that needs repair disappears from the Asset Manager's open path entirely. If the catalog is where people go to find their projects, that might deserve a visible "needs repair" state instead.

MrNeRF:
And please don't forget the comments from @kmeirlaen here: https://discord.com/channels/1293599119989805087/1388829600427020318/1541032518340579388 and following

shadygm:
Everything raised in the review and follow-up feedback has now been fixed and pushed in `442a82476` (along with the preceding review-fix commits). This includes the catalog persistence/migration blockers and inline interaction issues, plus the latest Asset Manager folder feedback: folders now map one-to-one to real user-selected directories, the default is configurable under General preferences and defaults to `~/.lichtfeld/assets`, the virtual/watched-folder UI is gone, and the folder info tag has been removed. The Release build passes, the focused regression suite passes (122 tests), and the final info-panel change passes all 31 Asset Manager panel tests.

MrNeRF:
Awesome! Thank you!

PR #1772, opened 2026-08-21
Log when adding a keyframe replaces an existing one

### LINKED ISSUE #1544: Adding a keyframe silently overwrites the one already at that time
**What's wrong**

`SequencerAddKeyframe` replaces any existing keyframe within 0.01 s of the target time
instead of adding one. In the GUI that is the right behaviour and matches Blender — you
can see what happened.

Through the API you cannot. A script that adds N keyframes and gets back fewer has no way
to know which ones were swallowed; nothing is logged and nothing is returned. This is easy
to hit because `seek()` clamps to the clip duration, so several adds in a row can all land
on the *same* clamped playhead and collapse into one keyframe.

Observed while scripting a 12-keyframe camera path: 8 adds produced 7 keyframes, silently.

**Where**

- Replace-instead-of-add: `src/visualizer/gui/sequencer_ui_manager.cpp:357-367`
  (`REPLACE_EPSILON_S`)
- The clamp that makes collisions likely: `src/visualizer/sequencer/sequencer_controller.cpp:65-67`

**How to fix**

Log at INFO when a keyframe is replaced rather than added, and include the replaced id.
The MCP `sequencer.add_keyframe` response already returns the full keyframe list, so a
caller can diff — but it should not have to guess that a replacement is even possible.

OmarFaig:
Adding a keyframe within REPLACE_EPSILON_S (10ms) of an existing one intentionally replaces it, matching Blender/After Effects/Maya. That's fine in the GUI, where you can see it happen, but the same event handler backs the MCP `sequencer.add_keyframe` tool, so a script adding a batch of keyframes can lose some of them with no signal at all. This is easy to hit because SequencerController::seek clamps to the clip duration, so several adds beyond the end of the clip can land on the same clamped playhead and collapse into one keyframe.

This adds a LOG_INFO in the replace branch of cmd::SequencerAddKeyframe::when, naming the replaced keyframe id and the time, so a caller can at least find the swap in the log even though nothing in the MCP response signals it directly.

No existing test reaches this handler: the sequencer MCP tool tests build their own lightweight FakeSequencerBackend that doesn't go through SequencerUIManager, and nothing else instantiates it either. I verified the fix by reading: the replace/add branch itself is unchanged, the log line sits directly on the path already taken when `existing != keyframes.end()`, and the live MCP wiring in mcp_gui_tools.cpp emits the exact same cmd::SequencerAddKeyframe event the GUI does, so one log line covers both paths.

Fixes #1544

### DIFF
--- a/src/visualizer/gui/sequencer_ui_manager.cpp
+++ b/src/visualizer/gui/sequencer_ui_manager.cpp
@@ -378,6 +378,7 @@ namespace lfs::vis::gui {
                                                           std::abs(kf.time - time) < REPLACE_EPSILON_S;
                                                });
             if (existing != keyframes.end()) {
+                LOG_INFO("Replaced keyframe {} at t={:.3f}s instead of adding a new one", existing->id, time);
                 controller_.updateKeyframeById(existing->id, position, rotation, focal_mm);
             } else {
                 lfs::sequencer::Keyframe kf;


MrNeRF:
Thank you!

PR #1773, opened 2026-08-21
Add measurement tool with translate gizmo to HTML viewer export

jacobvanbeets:
## Summary
Adds a self-contained distance-measurement tool to the exported HTML viewer (no external files or web server required):

- Vendors a trimmed PlayCanvas TranslateGizmo (MIT licensed) since the lightweight supersplat-viewer bundle does not include gizmo support (gizmo.js).
- Reimplements a SuperSplat-style measurement tool against the viewer's simpler runtime, reusing its existing depth-picker for point placement (measure-tool.js).
- Adds a toolbar toggle button, SVG line/point overlay, and a length readout/input to the viewer template and stylesheet.
- Freezes the orbit/fly camera controller while a gizmo handle is being dragged, since both are separate listeners on the same canvas and stopPropagation() alone can't arbitrate between them.
- Adds 	ools/generate_html_viewer_resources.py to regenerate the previously hand-maintained src/io/formats/html_viewer_resources.hpp from the editable viewer sources.

## How to use
Click the ruler icon in the exported viewer's bottom toolbar, click two points on the splat surface to measure the distance between them, drag the gizmo handles to fine-tune a point, or edit the length field directly.

## Testing
- Verified the merged viewer JS bundle parses with \
ode --check\.
- Full local rebuild of LichtFeld Studio succeeds with no new warnings/errors.
- Manually exported a scene to HTML and verified the measurement tool (point placement, gizmo dragging, length editing, camera freeze while dragging) in a browser.


src/visualizer/gui/resources/viewer/measure-tool.js
```diff
@@ -0,0 +1,308 @@
...
+// Prevent the viewer's own orbit/fly camera controller from consuming mouse
+// deltas while the measurement gizmo is being dragged. `InputController` only
+// accumulates raw input into `this.frame.deltas` each frame (some other part
+// of the viewer reads and applies those deltas to the camera); starving it of
+// deltas for a frame is enough to keep the camera perfectly still, without
+// having to fight DOM event propagation order against the gizmo's own
+// pointer handlers (both are plain listeners on the same canvas element, so
+// stopPropagation()/preventDefault() alone can't select one over the other).
+let _lfsCameraFrozen = false;
+const _lfsPatchCameraFreeze = () => {
```
MrNeRF: This freeze breaks the camera after a right or middle button drag on a handle.

`read()` also returns the button press/release deltas, and the real `update()` adds them into `_state.mouse[i]`. While frozen the +1 press is thrown away, but the -1 release comes after `transform:end` and gets applied. So every gizmo drag leaves `_state.mouse[button]` at -1. The gizmo takes all three buttons, and `update()` does `pan = this._state.mouse[2] || ...`, so after one right-button drag on a handle every left drag pans with the wrong sign and orbits at double speed, for the rest of the session. I measured it: after one right drag of the x arrow, `_state.mouse` is `[0,0,-1]` and a plain left drag on empty canvas moves the camera position by (0.011, -0.002, -0.212). Before the gizmo drag the same left drag only orbits.

Also related: the flag is cleared inside pointerup, so the mouse deltas from the last frame of the drag are still applied. A 15 px move in the same task as the release gives an 11 degree camera jump.

Two ways out:
1. Keep the freeze but don't drain. Run the normal bookkeeping and only zero the movement terms, and stay frozen for one more `update()` after `transform:end`.
2. Drop the patch. Register the gizmo canvas listeners with `capture: true` and call `stopImmediatePropagation()` on handle hits and while dragging. Capture listeners on the target run before the bubble ones `KeyboardMouseSource` installed, so the camera controller never sees the drag at all and its state stays consistent.

Either way, please limit the gizmo to the left button so right drag keeps meaning pan everywhere.

src/visualizer/gui/resources/viewer/gizmo.js
```diff
@@ -0,0 +1,1240 @@
...
+        return layer;
+    }
+
+    constructor(camera, layer, name = 'gizmo') {
+        super();
+        this._size = 1;
+        this._scale = 1;
+        this._coordSpace = 'world';
+        this._handles = [];
+        this._mouseButtons = [true, true, true];
```
MrNeRF, line 329: All three buttons are on here. With the viewer's right drag = pan, a right drag on a handle is what triggers the broken camera state above. Suggest `gizmo.mouseButtons[1] = gizmo.mouseButtons[2] = false` in the measure tool.

src/visualizer/gui/resources/viewer/index.css
```diff
@@ -196,6 +196,82 @@ button {
 #controlsWrap > #buttonsContainer .controlButton.toggle.right:not(.active) {
   background-position: 200% 0%;
 }
+#controlsWrap > #buttonsContainer .controlButton.active {
+  color: #F60;
+}
+
+/* Measurement tool overlay */
+#measureToolSvg {
```
MrNeRF, line 208: This doesn't reach the children. Line 47 has `#ui * { pointer-events: auto; }`, which matches the circles and the `<use>` lines and turns them clickable again. `document.elementFromPoint()` on a placed point returns `circle#measureLineStart` and the pointerdown never hits the canvas.

Effects: the centre sphere (free xyz move) sits under the marker and can't be grabbed, clicking a point to reselect it only works in the 1 to 2 px ring around the marker, and the 6 px line blocks orbit drags that start on it.

`#measureToolSvg, #measureToolSvg * { pointer-events: none; }` fixes it, I checked the centre handle drags fine with that.

src/visualizer/gui/resources/viewer/index.css
```diff
@@ -196,6 +196,82 @@ button {
 #controlsWrap > #buttonsContainer .controlButton.toggle.right:not(.active) {
   background-position: 200% 0%;
 }
+#controlsWrap > #buttonsContainer .controlButton.active {
```
MrNeRF: This also hits the orbit/fly toggle. Those buttons already use `.active`, and their active state is an orange pill with a white icon. With this rule the icon turns orange on orange and nearly disappears. Please scope it to `#measure.active`.

src/visualizer/gui/resources/viewer/measure-tool.js
```diff
@@ -0,0 +1,308 @@
...
+    // ---- point picking on click (drag = camera navigation, not a pick) --
+    const isPrimary = (e) => (e.pointerType === 'mouse' ? e.button === 0 : e.isPrimary);
+    let clicked = false;
+
+    const onPointerDown = (e) => {
+        if (active && !clicked && isPrimary(e)) {
+            clicked = true;
+        }
+    };
+    const onPointerMove = () => {
```
MrNeRF: Any pointermove clears `clicked`, so a 1 px jitter between down and up cancels the pick. Real mice and trackpads do that a lot, and on touch it almost never places. SuperSplat gets away with it in an editor, but for a viewer I'd use a small deadzone (the viewer's own double tap check uses 8 px).

src/visualizer/gui/resources/viewer/gizmo.js
```diff
@@ -0,0 +1,1240 @@
...
+ * IMPORTANT: this file is NOT a self-contained module at runtime. It is
+ * spliced (via tools/generate_html_viewer_resources.py) into the same script
+ * scope as the bundled PlayCanvas engine classes already present in
+ * index.js (Vec3, Quat, Mat4, Ray, Plane, Color, Layer, Entity, MeshInstance,
+ * Mesh, ShaderMaterial, EventHandler, math, and the various geometry/blend/
+ * cull-mode constants). Do not add real `import` statements here.
+ *
+ * One deliberate behavioral trim: `TransformGizmo._drawSpanLine` (the
+ * full-length axis "guide line" drawn while dragging) is a no-op because the
+ * viewer's engine build does not include `app.drawLine` (immediate-mode
```
MrNeRF: This isn't true, the bundle has `app.drawLine` (line 50028 in index.js, `Scene.drawLine` at 30686). So the axis guide lines while dragging were dropped without need. Either bring them back (about 30 upstream lines) or say the real reason. Leftovers from the trim that are now dead: `color4from3`, `AXES_XYZ`, `_dirFromAxis`. Also the copyright line says 2011-2026, upstream at this revision says 2011-2024.

src/visualizer/gui/resources/viewer/index.js
```diff
@@ -102465,6 +102465,8 @@ const main = (app, camera, settingsJson, config) => {
     initXr(global);
     // Initialize user interface
     initUI(global);
+    // Initialize the measurement tool (translate gizmo + distance readout)
+    __lfsInitMeasureTool(global);
```
MrNeRF: This name only exists after the generator splices the block in, so the source index.js can't run on its own any more (serving the resources folder throws a ReferenceError). If the export appends the two files after the bundle, this can just be `initMeasureTool(global)`. See my comment on the script.

Also: 174 of the 176 changed lines in this file are CRLF to LF changes in the meshoptimizer block. Fine if intended, but it makes the real change hard to find.

tools/generate_html_viewer_resources.py
```diff
@@ -0,0 +1,153 @@
+#!/usr/bin/env python3
```
MrNeRF: I understand why this exists: the header is a 3.4 MB generated blob, master has no generator for it, and you had to rebuild it after touching the JS. But I'd rather not keep that header in git at all. Nothing checks it matches the sources (it already drifted on master), and every JS edit gives a header diff nobody can review.

Simplest fix: read the viewer files at export time. `export_html()` already builds the page by string replacement, and the app already resolves a resource folder at runtime (`lfs::core::getResourceBaseDir()`, used for locales, fonts, themes). Copy `resources/viewer/` there like the locales, read the five files in `generate_html()`, append gizmo.js and measure-tool.js after the bundle (the template only calls `main()` after the module has evaluated, so no marker splice is needed, and there are no name clashes). Then the header, the MSVC chunking and this script all go away, and editing the viewer JS needs no rebuild.

If you'd rather keep it compiled in, generate the header from CMake at build time instead (see `cmake/GenerateGitVersion.cmake` for the pattern) and drop it from git.

src/visualizer/gui/resources/viewer/measure-tool.js
```diff
@@ -0,0 +1,308 @@
...
+            if (!picker) {
+                picker = new Picker(app, camera);
+            }
+            const result = await picker.pick(e.offsetX, e.offsetY);
+            if (result) {
+                points.push(result.clone());
+                selection = points.length - 1;
+                syncGizmo();
+            }
+            e.preventDefault();
```
MrNeRF: These run after the await, so the event is long done. They don't do anything here.

src/visualizer/gui/resources/viewer/measure-tool.js
```diff
@@ -0,0 +1,308 @@
...
+            _dragDir.sub2(points[1], points[0]);
+            _dragStartLength = _dragDir.length();
+            if (_dragStartLength > 1e-6) {
+                _dragDir.normalize();
+            }
+        }
+    });
+    lengthInput.addEventListener('change', () => {
+        if (points.length !== 2 || _dragStartLength <= 1e-6) return;
+        const newLength = Math.max(0.0001, parseFloat(lengthInput.value) || _dragStartLength);
```
MrNeRF: Small one: `0` and an empty field keep the old length, a negative value snaps B onto A at 0.0001, all silently. Maybe reject those and show a unit next to the value.

MrNeRF (changes requested):
Thanks, this is a nice feature and the gizmo trim is clean (I diffed it against PlayCanvas 2.13.6, the revision the bundle pins: picking, drag, scale and pointer code all match). I built the exported page and drove it in headless Chromium, so the findings below are measured, not guessed. Two of them break the tool in normal use, the rest are small.

One more thing not in the diff: THIRD_PARTY_LICENSES.md lists SuperSplat and splat-transform but not the PlayCanvas engine. Since this PR vendors engine source, please add an entry.

And on the Python script, see my comment on the file. Short version: I would rather not have a 3.4 MB generated header in git at all.

jacobvanbeets:
Thanks for the detailed review! I've pushed a new commit addressing all the feedback:

- **Export-time resource loading**: Removed the generated 3.4MB C++ header and the Python generator script. \iewer_template.html\, \index.css\, \index.js\, \gizmo.js\, and \measure-tool.js\ are now read from disk at export time via \getViewerResourcesDir()\, mirroring the existing locale/font/theme resource pattern. Added a dev-time CMake copy step and install rule accordingly.
- **Camera freeze bug**: Replaced the \InputController\ monkeypatch with capture-phase pointer listeners + \stopImmediatePropagation()\ on the gizmo, which also fixes the camera desync that occurred after right/middle-button drags.
- **Gizmo restricted to left mouse button only**.
- **Guide lines restored**: confirmed \pp.drawLine\ does exist in the bundle and restored real guide-line/span-line rendering; removed the \color4from3\ helper, which is now actually dead.
- **SVG pointer-events**: fixed child elements (circles/lines) blocking canvas picks.
- **CSS scoping**: \.active\ color rule scoped to \#measure\ so it no longer leaks onto the orbit/fly toggle buttons.
- **Pointer deadzone**: added an 8px deadzone before canceling point placement, instead of canceling on any movement.
- **Length input validation**: rejects 0/negative/empty values and now shows a unit label.
- Removed no-op \preventDefault\/\stopPropagation\ calls that ran after \wait\ (moved before the \wait\ instead).
- **Copyright/licensing**: corrected the PlayCanvas copyright year (2011-2024) and added PlayCanvas Engine to \THIRD_PARTY_LICENSES.md\.
- **index.js**: restored to be byte-identical with upstream — the measure-tool init hook is now invoked from \iewer_template.html\'s bootstrap script instead of being inserted into \main()\, eliminating the unrelated CRLF/LF diff noise.

I've built and installed locally to confirm the new export-time resource pipeline works end-to-end, and verified a freshly exported HTML file contains the expected measurement tool wiring. Ready for another look whenever you have time — let me know if anything else needs adjusting.

Issue #1534, opened 2026-08-02
Bug: Nonsense error message about using GUT and igs+ together displayed


fasteinke:
This has been around for a while, and latest master still has it: under some circumstance, and in the latest instance occurred after a Reset Training was initiated, and a very small number of steps to train for selected, then attempting to start the training throws up this error message. At no time was igs+, or GUT, selected or played with prior to this point - the message has no relevance. Fiddling with or restarting the app makes it go away, not clear what the precise, required action is.

kmeirlaen:
Can you provide a log file from when it occurs?

fasteinke:
Will look to do so, next time I use LFS ...

MrNeRF:
Not covered by #1779 (GUT SH + speed). The message itself is correct for gut + igs+; what goes wrong is the gut training setting surviving a strategy reset into igs+, which is a GUI state issue. Leaving open.

Issue #796, opened 2026-01-22
GUT transforms to point cloud in editor will not link back to non-GUT viewer models creating offset


grafpez:
When a model is adjusted in the editor (translate, rotate), either with gizmo or numeric, all rendering views (splat, point cloud,..) will have the same identical position.
However, when GUT is selected, and when the model is adjusted for translate or rotate, the GUT splat will become offset to all other non-GUT models (apart from point cloud)
So when further adjustments are made in either, this can create further disparity and creates editing problems.
I think there's some math that's not applied to transforms for models where GUT is selected.
.
To test, load a model, click home, centre the model with  <world space>, <origin> with <translate> and <rotate> to suit.
Check viewer modes (splat rendering, point cloud, gaussian rings, centre markers) are all as intended in the same place.
Select <GUT> and <Point Cloud>.
Move the model distinctively in one direction.
Deselect <GUT>.
Select <Splat Rendering>
You should now notice a shift in the splat models with an offset you created in the process. (However, this shift does not apply to Point Cloud)
took me a while to work this one out as I thought I wasn't using the tools correctly, but I believe this is a bug

MrNeRF:
Status note: the GUT render path in the current viewer applies node transforms in the shader; #1779 does not touch the viewer. Please re-check on master and report if the offset persists.

Issue #742, opened 2026-01-09
Make gut trained plys regular 3dgs viewer compatible


MrNeRF:
supersplat vs LFS first + spirulae second
Detais: https://discord.com/channels/1293599119989805087/1413218366348591244/1434334347594301596

https://github.com/user-attachments/assets/ac0ccfde-1bf3-436f-a527-2a42b953387c

Issue #713, opened 2026-01-06
Transform changes lost when start training with GUT enabled


gradeeterna:
After applying 180 rotation to the model with GUT enabled, the pointcloud rotates back to it's original transform when starting training. It seems to be somehow related to the GUT checkbox in the rendering tab, and training produces incorrect results.

https://github.com/user-attachments/assets/8ab3a9e1-a471-46aa-890e-2199b677cf45

https://github.com/user-attachments/assets/12382aca-0a9e-48f5-bd3b-8bd1c08d930b

Tested in nightly builds from 4-6th Jan.

jacobvanbeets:
I can't reproduce this.
• Load dataset
• click enable GUT
• click rotate gizmo
• Typed 180 in X, rotate 180 degrees
• Clicked training tab
• Clicked start training
Observed the splats and it acts as expected without issue.

gradeeterna:
So strange, I can reproduce every time with any dataset on Windows! 

https://github.com/user-attachments/assets/d1c4c812-ab7e-48b2-a7b7-67696a901f83

jacobvanbeets:
I'm going to attempt create a potential fix but it would need testing. As it works for me it needs to be tested be someone else

jacobvanbeets:
I tried to fix it
1. Get the PointCloud node's world transform
2. Transform all the point positions by that transform
3. Then create the SplatData with the transformed positions

This actually broke it for me.
@MrNeRF Sorry i don't know what to do here
This might be one for you to try and solve.

MrNeRF:
Status note: since the Vulkan viewer rework the GUT alpha-blend shader applies the node's world transform (model_transforms in alphablend_shader.slang). #1779 does not touch the viewer; if this still reproduces on current master please say so and it can be looked at separately.

PR #1779, opened 2026-08-22
Train every SH degree on the GUT path and make it faster

### LINKED ISSUE #834: investigate sh3 in gut training
I think the rendering panel is not properly connected i.e. the degree cannot be adjusted. 
Check the file if that really has sh3 when trained with sh3

### LINKED ISSUE #1703: When using the GUT mode, SH can only be 0
Is there any plan to improve this? Or are there any other considerations for such a design?

MrNeRF:
Fixes #1703: GUT (`--gut`) training now works with SH degrees 1 to 3. It also gets a good deal faster.

**Why SH was limited.** The GUT path uses the standalone optimizer, which had no step for the SH-rest moments in their joint codec; only the fused fastgs backward could update them. That code now takes its gradient from a functor, and a standalone kernel feeds it the gradients the GUT backward already accumulates. Codes, value bounds and packed moments come out bitwise identical to the fused path (new test).

**Speed.** London fisheye, images_2, 3M splats, 5000 iterations, end to end: **203 s -> 154 s** (fastgs on the same recipe: 99 s). London images_8, 7k, 1M: 113 s -> 91 s.

What changed, in order of effect:
- The per-pixel backward runs two pixels per thread: half the warp reductions and global atomics per splat and tile; the pair shares one unpacking of the splat. Backward kernel 2.07 -> 1.27 ms per iteration on the garden bench.
- Per-splat matrices are built once per batch in shared memory instead of per pixel; the RGB backward goes from 122 registers with a stack frame to 88 with none, the forward from 86 to 48.
- The forward stages splat colours in shared memory instead of re-reading them per pixel, in half-size batches (0.45 -> 0.34 ms, six blocks per SM).
- Fisheye and distorted-pinhole cameras run the same tuned kernel body as undistorted pinhole (the specialisation is keyed on global shutter, not on the camera model).
- Raw quaternions, log scales and opacity logits go straight into the kernels; the per-iteration tensor-op normalisation, the separate activation-backward kernels and the CHW/HWC permutes are gone. The degree-0 SH backward is a 10-register kernel.
- Host path: intrinsics and distortion go through a pinned staging buffer; the tile sort runs over the grow-only capacity with sentinel keys so the host only needs the intersection count to detect overflow, and waits for it after the fill, sort and offsets are queued (every frame renders all intersections); the contiguous Adam parameters step in one launch. The shared-memory attribute is set once per kernel, and a failure is an error instead of a silently skipped render. The per-iteration gradient clears are one memset.

**Measured and not kept** (each with numbers in the commit history): shared-memory gradient accumulation per block, a warp-leader VJP, staging the per-splat ray-origin term, a per-frame matrix pre-pass, pushing the backward to five blocks per SM. The backward is bound by its global atomics, not by occupancy or ALU, and every extra byte of staged data costs a block per SM.

**Quality.** Unchanged at every step: garden SH0 27.96, london fisheye SH0 24.23. Garden with `--gut` by SH degree: 0: 27.96, 1: 28.11, 2: 28.21, 3: 28.46 (SSIM 0.898 -> 0.906); exported PLYs carry the SH-rest fields. London fisheye SH3: 24.48. Forward image bit-identical to before on the fixture; gradients match to 2e-7 relative. Tests cover the shN Adam parity, the fixture render/gradient parity, zero steady-state allocations, and the ungated SH initialisation.

**Issues.** Closes #1703 and closes #834 (SH3 trains on GUT, the exported PLY carries the SH-rest fields, the viewer passes its SH degree to the GUT backend). Not addressed here: #1534 is a stale `gut` setting surviving a strategy reset into igs+ (a GUI fix); #713 / #796 concern node transforms on the viewer's GUT path, which applies them since the Vulkan rework and were not re-verified in this PR; #742 is inherent to per-ray training (a model trained with GUT looks different in a plain EWA viewer).


MrNeRF:
Superseded by a fresh PR from the same branch with a shorter description.

Issue #1484, opened 2026-07-29
Plugin API version and feature list are declared twice, in two languages


MrNeRF:
**What's wrong**

The list of plugin features the host supports exists in two hand-maintained copies:

- `src/python/lfs_plugins/compat.py:19-28` — `PLUGIN_API_VERSION = "1.0"` and six feature strings
- `src/python/lfs/py_plugins.cpp:35-44` — the same version and the same six strings, appended to a list by hand

Plugin loading is gated on the Python copy; plugin authors read the C++ copy. If they drift, a
plugin can pass the manifest check and then find the feature missing at runtime.

Separately, `compat.py:19` hardcodes `LICHTFELD_VERSION = "0.5.3"` as a Python string literal
instead of deriving it from the CMake project version — so it has to be remembered at every release.

**How to fix**

One small `plugin_contract.json` at the repo root as the single source. Generate a header from it
for the C++ side; import it directly on the Python side. Derive `LICHTFELD_VERSION` from CMake.

OmarFaig:
Confirmed against current master — the three copies you named agree today (v1.0, six features, 0.5.3), so this isn't live drift yet, just an unenforced invariant. One correction to the count though: it's not two copies, it's closer to five. Beyond `compat.py` and `py_plugins.cpp`'s `plugins.attr("API_VERSION")`, there's a third hardcoded `PLUGIN_API_VERSION = "1.0"` at the top-level `lichtfeld` module in `src/python/lfs/module.cpp:2487` (separate from `lichtfeld.plugins.API_VERSION`), and the value is also literalized in the type stub at `src/python/stubs/lichtfeld/__init__.pyi:1530`. Nothing currently reads the module.cpp copy at runtime, but it's part of the declared public surface, so it can still drift silently.

The `plugin_contract.json` + codegen approach you proposed is the more complete fix, but I can't verify generated-header wiring without the CUDA toolchain, so I'd rather not put up a build-system change I can't build-test myself. Proposing a narrower fix that addresses the same harm: a contract test in `tests/python` that reads the C++ and stub files as plain text and asserts every copy matches `compat.py`'s values — the same pattern `test_verified_dead_code_cleanup.py` already uses for cross-language invariants that can't be imported. Drift then fails CI instead of reaching a plugin author at runtime.

This is the cheaper guarantee, not the complete one, and it stays passing after a future codegen refactor. Want the test now, or would you rather hold for the full contract file?

PR #1789, opened 2026-08-23
Guard plugin API version and feature contract against cross-language drift

OmarFaig:
The plugin API version and feature list are hand-maintained in two places, `compat.py` and `py_plugins.cpp`, and `LICHTFELD_VERSION` duplicates the CMake project version as a separate hardcoded string. While investigating, I found the duplication is actually wider than the issue describes: a third copy of `PLUGIN_API_VERSION` lives on the top-level `lichtfeld` module in `module.cpp`, and the value is literalized a fourth time in the type stub. Plugin loading gates on the Python copy in `compat.py`, while plugin authors read the C++ and stub copies at development time, so drift produces a plugin that passes the manifest check and then finds a feature missing or a version mismatch at runtime.

This adds a contract test in `tests/python/test_plugin_api_surface.py` rather than the `plugin_contract.json` codegen the issue proposes, because I can't verify generated-header wiring without the CUDA toolchain, and the test is the cheaper guarantee that stays valid after a future codegen refactor. It follows the existing cross-language text-assertion pattern already used in `test_verified_dead_code_cleanup.py` — reading the C++, stub, and CMake files as plain text and asserting their values against `compat.py`'s constants, by both set and count so neither a missing nor an extra feature slips through. Both failure modes (a stray feature, a version mismatch) were verified by deliberately introducing drift and reverting.

See the discussion on #1484 for the maintainer's response on scope.

Refs #1484

### DIFF
--- a/tests/python/test_plugin_api_surface.py
+++ b/tests/python/test_plugin_api_surface.py
@@ -5,6 +5,7 @@
 from importlib import import_module
 from pathlib import Path
 from types import ModuleType, SimpleNamespace
+import re
 import sys
 
 import pytest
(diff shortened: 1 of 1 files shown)


MrNeRF:
Good scope call. The codegen single source from #1484 is still the end state, so the issue stays open. Eventually this needs a proper cleanup as it has a bit diverged over time due to quick progress.

PR #1848, opened 2026-08-26
Asset Manager follow-up: thumbnails for every project, safer catalog, found-location relink

MrNeRF:
Follow-up to #1770.

**Thumbnails**
- Every saved project now has a thumbnail. The viewport capture is used when there is one; otherwise the first image of the dataset is embedded at save time, so CLI and headless saves get one too.
- Projects that already exist without a thumbnail show their first dataset image when the file knows where its dataset is. Older CLI-trained files do not record that, so they keep a name placeholder until they are saved once from the GUI.

**Catalog**
- Adding a folder scans it immediately. A project saved into a catalog folder shows up without pressing Refresh.
- A project that was moved or renamed on disk is shown as missing together with the place it was found, and one click relinks it. The catalog never moves an entry on its own.
- A project whose file now belongs to a different project can be relinked with Locate File.
- Unreadable catalog rows are skipped with a short notice instead of leaving the catalog empty. The catalog from before the migration is kept permanently as `library.json.legacy.bak`.
- Projects that need repair or were saved by a newer version get their own labels.
- Background scans are cancelled when the panel closes; a refresh during a scan runs afterwards instead of being dropped.

**Preferences**
- The working folder and Asset folder settings reject relative paths and expand a leading `~`.

**Fixes and cleanup**
- Several rows in the panel never appeared because RmlUi treats text values as false in `data-if`; they now use boolean bindings and a test guards against it.
- Placeholder cards show the project name.
- Removed the leftover watched-directories stylesheet, unused strings in all languages and dead styles; added the missing translations; a hint against pointing the Asset folder at a whole drive.
- File inspection from Python no longer holds the interpreter lock during disk reads.


MrNeRF:
Heads-up: the current tip (19d7ddf0d) has a regression: opened projects render in degraded mode ("model tensors must use Vulkan-external storage") because the lazy training-session restore skips the step that put the model's shN into Vulkan-external storage. Fix in progress; do not test the viewer on this tip until the next push.

MrNeRF:
Fixed in 7c1d3117f: opened projects render again (hydration now encodes the model's rest SH into Vulkan-external storage itself). Verified on five projects, zero degraded-mode errors.

PR #1853, opened 2026-08-26
Create a new project UUID when using Save As

### LINKED ISSUE #1852: Saving a project under new name does not show up in asset manager
* open LFS
* load a project through the asset manager
* use file, save as to save the project under a new name in the same directory
--> the new project does not show up on the asset manager

Root cause:
Asset manager is using the UUID to display/filter the different projects.  However, File, Save As, does not store/createa a new UUID.

Suggested solution: when File, Save As is used, create a new UUID so that the asset manager can pick up the new file.

shadygm:
## Summary

- assign a new project UUID when a titled project is saved to a different path, allowing Asset Manager to catalog both files
- rekey project singleton chapters while preserving UUIDs for first saves, same-path saves, and recovery publication
- carry the new identity through live-training snapshot saves and keep document rebinding transactional
- add document-level and visualizer regression coverage

## Testing

- CMAKE_BUILD_PARALLEL_LEVEL=22 cmake -B cmake-build-release -S . -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=OFF -DLFS_VCPKG_MAX_CONCURRENCY=22 -DLFS_CUDA_COMPILER_CACHE=OFF
- cmake --build cmake-build-release -j22
- manual Asset Manager Save As workflow

Fixes #1852

### DIFF
--- a/src/io/include/io/project_container.hpp
+++ b/src/io/include/io/project_container.hpp
@@ -523,6 +523,10 @@ namespace lfs::io::project {
     struct CompactionOptions {
         ReaderOptions compatibility;
         lfs::core::Uuid new_file_uuid;
+        // Save As staging may replace the project identity before rewriting
+        // the project-keyed singleton chapters. Ordinary compaction leaves
+        // this null and preserves the source project UUID.
+        lfs::core::Uuid new_project_uuid = {};
         lfs::core::Uuid commit_uuid;
         lfs::core::Uuid snapshot_uuid;
         std::uint64_t creation_time_unix_ns = 0;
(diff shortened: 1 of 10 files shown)


MrNeRF:
Pushed 9fe275471 with two small test fixes. The production change is fine as is.

1. The new visualizer test used private members of VisualizerImpl and did not compile. Added the friend line in visualizer_impl.hpp like the other VisualizerImplResetTest cases.
2. The new document test expected find(fourcc, old_uuid) to be nullptr in the Save As copy. find() also returns tombstoned rows, so it found the (correctly) tombstoned old row and failed. The test now checks that the new-key row is live and the old-key row is absent or tombstoned.

Note: the tests never ran before because the PR was built with BUILD_TESTS=OFF and CI only runs lichtfeld_format_tests. With the fixes: ProjectDocumentTest.* 60/60, VisualizerImplResetTest.*SaveAs* 17/17 locally.

Issue #1845, opened 2026-08-25
Proposal: screen-space depth window for selection (replacing the fixed-width depth box)


alexmgee:
Prior discussion was requested for larger features so I figured I'd post this as an issue to begin with.

I've built a replacement for the selection depth filter's fixed-world-width box. It's a screen-space window centered at a percentage of the viewport, constant on-screen size at any Near/Far, with X/Y offset and a three-state visualization mode (Off / Dim outside / Hide outside). It's complete and heavily tested (CUDA/CPU golden parity across pinhole, ortho and equirect; scripted regression locks; hands-on acceptance), and I'd like to contribute it.

I'd planned it as a sequence of PRs so each review stays manageable: (1) the screen-space window itself (2) per-axis scale plus drag-to-draw the window with corner handles and (3) per-panel windows in split view with a sync toggle. 1 and 2 are finished and ready and working, 3 is underway.

While building this I found and fixed a camera-consistency defect in my own first cut (command selections could build the mask from one camera and depth-filter with another), and reported a few pre-existing issues separately (#1785, #1786, #1788). I'm happy to adjust scope or approach to fit the project's direction before opening the PR. I have been wanting to contribute at a much larger level for a long time and the existing depth-selection was a paint point for me so I started there. 

I made a crude demo video to show it in action: https://youtu.be/R5gxGW4uAQc?si=h6eySg-2NSYzmSqD

MrNeRF:
Awesome! As said on discord this will be a great addition. Please create prs vs the dev branch for now as we are chasing the next release and stabilize master. This will be merged into master after the release of 0.6

MrNeRF:
I think that's completed right? Thanks @alexmgee

PR #1860, opened 2026-08-27
Make the selection depth window size and position use screen units instead of in-world units

alexmgee:
Implements step 1 of the three-PR sequence proposed in #1845.

Demo video: https://youtu.be/R5gxGW4uAQc

### The problem

The selection depth region already tracked the camera. `SelectionTool::syncDepthFilterToCamera`
anchored `depth_filter_transform` to the viewport camera's rotation and translation and set the
box to `[-depth_far, -depth_near]` in Z, so it was never a static world box and it already
expressed a near/far band.

What it was not was screen-shaped. Its cross-section came from `frustum_half_width_` — a **world**
half-width — with the height derived from that by dividing by the viewport aspect ratio, giving a
prism of constant world cross-section swept through depth. Because that size was expressed in
world units, what it covered on screen was a by-product of the projection rather than something
you could state directly. Under a pinhole camera the region covers less and less of the viewport
the farther away it sits, and it moves with `focal_length_mm`. Under an orthographic camera it
holds the same share of the viewport at any distance — screen scale there does not depend on
depth — but that share moves with `ortho_scale`, the pixels-per-world zoom. Under equirectangular
a world box has no screen rectangle at all. In none of the three could you say "select what lies
between here and there, inside this part of my screen" and have it mean one fixed thing.

This PR redefines the cross-section as a fraction of the framebuffer. `depth_filter_scale` is now
a share of the render size, converted to world half-extents at the far depth through the pixel
focal lengths, with a separate pixels-per-world path when the camera is orthographic. The result
is that the region keeps a constant footprint in the viewport across the whole near/far band
whatever the projection: under a pinhole camera it sweeps out a frustum, under an orthographic
camera it stays a prism, and under equirectangular it is an angular region. In each case the
predicate that commits a selection tests the projected splat centre against the same screen
rectangle rather than against a world box.

### Where this is going

This is the first of three steps, kept deliberately separate so each is reviewable:

1. **This PR** — the screen-space window itself: semantics, overlay, toolbar and input, the
   compatibility API, and a camera-consistency repair (below).
2. **Next** — per-axis window scale and drag-to-draw handles.
3. **Later** — per-panel windows with a sync toggle, only after step 2 is reviewed.

Step 2 is built and implemented, but there is no passing verification record for it on this
base; step 3 is designed but not yet built. Step 2 is held back rather than bundled here, because
reviewing both together would mean reading substantially more diff across C++, CUDA/Slang, Python,
UI and input in one sitting.

### What is in this PR

- Screen-space depth-window types and settings plumbing.
- The window routed through the GPU lanes, with a Slang containment test.
- The screen-window predicate applied when committing selections.
- The window drawn as a 2D screen-space overlay.
- Selection panel controls: a Size slider for the window's on-screen scale, applied uniformly;
  X and Y sliders that move the window rather than resize it; and a button cycling the
  outside-window visualisation between off, dim and hide.
- Python API, toolbar and locale wiring, plus a Z-sign fix in the Python depth-filter fallbacks.
- Shift+Alt+Scroll bound to the window size.
- **A camera-consistency repair** — see the disclosure below.

### Design notes
Fresh sessions intentionally default to Dim outside: enabling the depth window makes its active region immediately legible while preserving the surrounding scene as context; the toolbar still cycles to Off or Hide.

### Defect found and fixed before submission

While preparing this branch we found a real defect in the selection service and fixed it here
rather than shipping around it.

**The defect.** A selection command can name a camera (`camera_index`). The command built its
mask from that camera, but the depth filter independently re-resolved the *viewer* camera and
filtered against that instead. Whenever the two disagreed, the mask and the filter were computed
from different cameras. The same class of mismatch also affected colour selection, which filters
during seed picking — before commit — so a repair confined to the rectangle path would have left
that path wrong.

**The fix.** One projection context is now resolved once per selection pass and used, held by
const reference, by mask creation, hover seeding and depth filtering alike. Nothing downstream
re-resolves. `camera_index = -1` means "the current viewer". Six of the nine registered
`selection.*` tools — `rect`, `polygon`, `lasso`, `ring`, `brush` and `click` — take that as their
omitted-argument default, matching `render.capture`, where omitting the index captures the live
viewport. `selection.by_description` is the exception: it still reads
`args.value("camera_index", 0)` and its published schema documents `default: 0`. On production
paths an out-of-range index now fails with a specific error instead of silently degrading to the
viewer. The one exception is the test-only hovered-gaussian override: while it is armed,
`selectRing` and `selectByColorAt` let a non-negative index that names no existing camera resolve
to the viewer instead of failing. Nothing in the application arms it.

**Signature change you will see in the diff.** Five of the six selection entry points on
`SelectionService`, and all five on `SceneManager`, previously defaulted `camera_index` to `0` —
dataset camera zero; `selectByColorAt` already defaulted to `-1`. Combined with the stricter
validation above, calling e.g. `selectRing(x, y, mode)` with no camera on a scene that has no
dataset cameras would now fail rather than quietly fall back to the viewer. All ten defaults
therefore move to `-1`, so an omitted argument means "the viewer" consistently in C++ and across
those six MCP tools. Both classes are exported (`LFS_VIS_API`), so these defaults do sit in public
headers. Every caller of those ten entry points already passes an explicit index, so nothing in
the repository changes behaviour; the one call that relies on a default is `selectByColorAt` in
`src/visualizer/operator/ops/selection_ops.cpp`, whose default was already `-1` and is unchanged
here.

**Regression tests.** Two tests reproduce the mismatch on the rectangle and colour paths and pass
here; they were written against the broken behaviour first, though this branch carries no red-run
artifact for that ordering. Additional tests pin the out-of-range failure and one degenerate case:
an in-range camera that cannot produce a valid projection now returns a defined error rather than
silently substituting the viewer.

### Two upstream fixtures translated

Two pre-existing tests encoded the *old* world-space contract:

- `TestingScreenPositionsFallbackAppliesDepthFilterInPointCloudMode`
- `ColorSelectionAppliesDepthFilterToSimilarityMask`

They set a world AABB of ±0.25, which under the screen-window predicate becomes a camera-depth
band of [-0.25, 0.25] and therefore rejects everything. They have been **translated to the new
contract with their original assertions preserved byte-for-byte**. Three values changed: the depth
band, now [8.0, 8.875], which keeps the near splat (depth 8.5442) and rejects the far one
(9.2063); the X and Y extents, from ±0.25 to ±0.5; and the window scale, pinned to 1.0 from its
0.35 default. The offsets were pinned as well, at the zero they already held. This path reads the
band, the window scale and both offsets — only the X and Y of the filter box go unread — so it is
the scale pin that leaves depth as the only thing able to decide the outcome. The band lives in
one helper per test file.

A third test, `RingsCommitAppliesDepthFilter`, passed but *vacuously*: its hovered-id override
puts only one gaussian in the mask, so any rejecting band — including one that rejects
everything — produced the same empty result. It now has a positive-control sibling using the same
band: the negative rules out "the filter never ran", the positive rules out "the filter rejected
everything". Neither alone excludes both.

### Verification

Built on Windows against this branch rebased on `dev`, Release, Ninja 1.12.0.
C++ is compiled with MSVC toolset **14.51.36231**; CUDA **12.9 (V12.9.41)** is compiled by nvcc
using MSVC toolset **14.44.35207** as its host compiler. That host toolset is pinned by our local
build scripts, not by the project's CMake, and `--allow-unsupported-compiler` permits the pairing
rather than selecting it. CUDA architectures: `native`.

**Targeted suites — 37 pass / 0 fail**, run against this branch's tip:
`SelectionServiceInteractionsTest`, `SelectionOperatorModalTest`, `SelectionCommandDispatchTest`,
`SelectionScreenWindow` (the CPU/CUDA parity golden, 2 tests on this branch), and
`McpAppUtilsTest`.

This is a **targeted subset**, not the full binary, and it does not cover everything this PR
touches: the Python API, the toolbar, locale wiring, session state, input binding and the
screen-space overlay are not exercised by these five suites. No full-suite run was made against
this branch, so we are not claiming a full-suite pass and make no claim about the state of any
suite outside the five listed above.

### Known limitations

- Window state is global; there is no per-panel state (step 3).
- No per-axis scale and no drag-to-draw handles yet (step 2).
- Scale and offset are session-only and are not exposed over MCP at all — `depth_filter_scale`
  and the two offsets appear nowhere in the MCP tool surface. Withholding them is deliberate, not
  an oversight. Note that the published render-settings schema documents only the two
  visualisation fields (`hide_outside_depth_box`, `depth_filter_viz_mode`), while the settings
  endpoints also read and accept `depth_filter_enabled`, `depth_filter_min` and
  `depth_filter_max` without documenting them; that undocumented surface predates this PR and is
  untouched by it.

### Upstream work already incorporated

This branch is rebased on `dev`, so it includes and is consistent with #1794 (depth filter
collapsing to a 1 cm slab) and #1798 (depth range surviving a project reopen). One commit from
our own history was dropped during the rebase because upstream refreshed the committed Python
stubs first — it carried only upstream-owned drift.

### Follow-ups

Step 2 next, then step 3 after step 2 is reviewed and merged.


src/python/lfs_plugins/selection_controls.py
```diff
@@ -384,18 +493,80 @@ def _set_depth_far(self, value):
...
+            pass
+
+    def _viz_mode_label(self):
+        key, fallback = _VIZ_MODE_LABELS.get(int(self._viz_mode), _VIZ_MODE_LABELS[_DEFAULT_VIZ_MODE])
+        return _ui_label(key, fallback)
+
+    def _viz_mode_icon(self):
+        return _VIZ_MODE_ICONS.get(int(self._viz_mode), _VIZ_MODE_ICONS[_DEFAULT_VIZ_MODE])
+
+    def _set_depth_scale_percent(self, value):
```
MrNeRF, line 548: Caught the slider echo in the act here: three scroll notches on the window size give 35 → 37 → 39 → 41 instead of 40.5. The slider is writing its rounded percent back after every notch, so scrolling gets snapped to whole percents. Near/far dodge this with their `_visible` / `_last_state_key` / holdoff guards — the three new setters just need the same treatment. Same echo gremlin we chased in #1794/#1798.

src/visualizer/rendering/rendering_manager.cpp
```diff
@@ -27,6 +27,29 @@
 namespace lfs::vis {
 
     namespace {
+        // Sanitizes only the screen-window fields. Near/far normalization is
+        // deferred until the screen-window path owns depth_filter_min/max: the
+        // legacy tool and GUI bindings rewrite those values per frame, and a
+        // sanitizer in that loop shifts its equilibrium.
+        void sanitizeSelectionWindowSettings(RenderSettings& settings) {
```
MrNeRF, line 38: Here's the nastier one I found: flip `depth_filter_enabled` on via render settings while the box still has its constructor defaults, and the new decode turns that box into an empty band (near −100, far 0). The scene vanishes, selections return nothing, and the overlay rect keeps drawing like everything's fine. I did it through MCP `render.settings.set` — before this PR that write was harmless. Worth teaching the sanitizer (or the enable path) to recognize the old positive-Z default and normalize it.

src/rendering/rasterizer/vulkan/shader/src/slang/vertex_shader.slang
```diff
@@ -335,16 +337,35 @@ SplatActiveState compute_splat_active_state(
         }
     }
 
+    float3 projected_center = project_point(mean_ws, cam);
```
MrNeRF, line 340: One subtle mismatch: this test goes through `project_point`, which picks up the temporal upscaler's jittered center — the CUDA test and the overlay use the plain center. So while temporal is settling, a splat parked right on the window edge will flicker in and out of dim even though its selection never changes. Using the unjittered center just for this check would settle it.

src/rendering/selection_ops.cu
```diff
@@ -683,6 +684,107 @@ namespace lfs::rendering {
...
+            const float image_height = static_cast<float>(height);
+            float px = 0.0f;
+            float py = 0.0f;
+            float depth = view_z;
+
+            if (camera_model == static_cast<std::uint32_t>(ScreenWindowCameraModel::Pinhole)) {
+                px = 0.5f * image_width + pixel_focal_x * view_x / view_z;
+                py = 0.5f * image_height + pixel_focal_y * view_y / view_z;
+            } else if (camera_model == static_cast<std::uint32_t>(ScreenWindowCameraModel::Orthographic)) {
+                px = 0.5f * image_width + ortho_scale * view_x;
```
MrNeRF, line 756: If `ortho_scale` ever ends up ~0, Vulkan quietly substitutes a default but this kernel runs with the raw value — then every splat projects to dead center, and the window either takes everything or nothing depending on where it sits, while the viz shows something else entirely. A weird state to get into, but one clamp on the host side and the lanes can't disagree.

src/visualizer/selection/selection_service.cpp
```diff
@@ -1193,12 +1280,23 @@ namespace lfs::vis {
             return {true, 0, {}};
         }
 
+        const auto viewer_context = resolveViewerViewportContext();
+        const auto projection_context_opt = viewer_context
+                                                ? projectionContextFromViewerContext(*viewer_context)
+                                                : std::nullopt;
+        if (!projection_context_opt) {
+            LOG_WARN("SelectionService: selectAllFiltered failed: no valid projection context");
```
MrNeRF, line 1373: These filtered/volume selects now refuse to run whenever the viewer viewport can't be resolved — even with the depth filter off, when nobody would use that projection anyway. Select-all used to survive a minimized window / first frame. Maybe only demand the snapshot when the filter is actually on?

src/visualizer/rendering/rendering_types.hpp
```diff
@@ -293,6 +292,10 @@ namespace lfs::vis {
         glm::vec3 depth_filter_min = glm::vec3(-50.0f, -10000.0f, 0.0f);
         glm::vec3 depth_filter_max = glm::vec3(50.0f, 10000.0f, 100.0f);
         lfs::geometry::EuclideanTransform depth_filter_transform;
+        float depth_filter_scale = 0.35f;
+        float depth_filter_offset_x = 0.0f;
+        float depth_filter_offset_y = 0.0f;
+        int depth_filter_viz_mode = 1;
```
MrNeRF, line 352: Noticed fresh sessions now start on "dim outside" where the old behavior was "show everything" (old projects migrate to off, that part's right). If that's a deliberate choice I'm fine with it — just say so in the PR text, since it changes what people see the first time they hit X.

src/python/lfs/py_selection.cpp
```diff
@@ -89,6 +163,104 @@ namespace lfs::python {
...
+            auto settings = rm->getSettings();
+            configure_depth_filter(settings, enabled, depth_near, depth_far,
+                                   std::max(informational_half_width, 0.05f));
+            settings.depth_filter_scale = clamped_scale;
+            settings.depth_filter_offset_x = clamped_offset_x;
+            settings.depth_filter_offset_y = clamped_offset_y;
+            rm->updateSettings(settings);
+        }
+
+        [[nodiscard]] std::tuple<float, float> fallback_depth_near_far(const vis::RenderSettings& settings) {
```
MrNeRF, line 263: Tiny one: with no tool alive, this reads the default box as near 0 / far 0 (it used to report far 100). Only startup/headless scripts ever land here, but a read-then-reapply script would build a zero-width band. Session restore already special-cases the old default — same guard would do.

tests/test_selection_screen_window.cpp
```diff
@@ -0,0 +1,428 @@
...
+            }
+        }
+        return projected;
+    }
+
+    bool cpuReferenceInside(const ProjectedPoint& projected, const ScreenWindowConfig& config) {
+        const float width = static_cast<float>(config.width);
+        const float height = static_cast<float>(config.height);
+
+        // KEEP IN SYNC: the screen-window formula lives in four places — this CPU
```
MrNeRF, line 141: Take it or leave it: love the golden — but of your four KEEP IN SYNC copies, the slang one is the only copy no test ever runs, and both parity nits above live exactly there. No action needed for this PR; maybe just flag it in the comment so future-us knows which copy is on the honor system.

MrNeRF (changes requested):
@alexmgee — this one was a pleasure to review. I built it on Linux (your 37 tests all pass here too), then just played with it for a while — zooming, dragging sliders, cycling the viz modes, poking it over MCP. The window really does just sit there on screen doing exactly what it promises, and watching the splats cut off exactly at the white border in hide mode was the moment I stopped hunting for math bugs. The camera fix underneath is the real deal too, and the tests you wrote for it are better than most of ours.

I did manage to break it in two places, and I have a handful of smaller notes — everything inline. Nothing here touches the design, it's all fixable in a round.

One note I can't attach inline because the file isn't in this diff: `tests/python/test_selection_controls_panel.py` — the stub there still only knows the legacy range API, so that suite never even constructs the new sliders. The echo issue I flagged inline sails right past it. Teaching the stub the window API would let you pin the guards down with a test.

src/visualizer/rendering/rendering_manager.cpp
alexmgee: Fixed in 54dd437: updateSettings migrates the constructor-default box when only the enable flag arrives, and the session-restore ordering hole this exposed (pre-existing) is closed. Also verified live over MCP: enable-only now decodes to near 0 / far 100.

src/python/lfs_plugins/selection_controls.py
alexmgee: Fixed in 54dd437: the five slider entry points reject replays while the echo holdoff is armed; text commits bypass only the holdoff. A commit-dedup mark an earlier draft added alongside is removed again — Enter-then-blur may write the same canonical value twice, matching upstream's two-listener behavior (the stale-buffer race between Enter and a later blur is inherited upstream behavior, accepted rather than proven harmless). Covered by the controller suite, including an upstream-compat case and a cross-session regression.

src/rendering/rasterizer/vulkan/shader/src/slang/vertex_shader.slang
alexmgee: Fixed in 54dd437, and more broadly than the first attempt: containment now projects with the displayed camera's real unjittered intrinsics, carried from capture (before applySceneViewJitter) through overlay slot 12 to the shader and as explicit parameters to both CUDA kernels. Jitter moves the draw sample, never the boundary — pinned by a temporal unit test. One honest caveat: live settling at a boundary splat under motion wasn't reproducible hands-on; the invariant is pinned at the unit level.

src/rendering/selection_ops.cu
alexmgee: Fixed in 54dd437: both the host caller and the kernel entry point substitute DEFAULT_ORTHO_SCALE over the whole invalid domain (!isfinite || <= 1e-5), matching the Vulkan lane. Covered by gtest across the domain, plus a live check at ortho_scale = 1e-7 returning a sane partial selection.

src/visualizer/selection/selection_service.cpp
alexmgee: Fixed in 54dd437: the five projection guards are now conditional on the depth filter actually needing a projection; the three sites that legitimately require one are unchanged.

src/python/lfs/py_selection.cpp
alexmgee: Fixed in 54dd437: the disabled getter reports the exact (0, 100) disabled band, so a read-then-reapply script can no longer build a zero band.

src/visualizer/rendering/rendering_types.hpp
alexmgee: Deliberate. Logic added to the PR description under Design notes.

tests/test_selection_screen_window.cpp
alexmgee: Done. All four copies now state the shared contract: rectangle framebuffer-centred, splat projected with real unjittered intrinsics, jitter never moves the boundary. They all should also name each other correctly.

alexmgee:
On the note about tests/python/test_selection_controls_panel.py: done, and extended. The stub knows the window API (the echo test you predicted now exists), the controller suite is 84 tests, and the C++ side grew to 72, including off-centre/asymmetric CPU–CUDA parity and five end-to-end explicit-camera cases that run the real projector kernel.

MrNeRF:
Awesome. Thank you!

PR #1870, opened 2026-08-28
Fix rendering of Python-added splats

thomas-koch-core11:
**Summary**
Splats added through the Python `add_splat()` API don't render when SH value quantization is enabled because shN is parked in memory no accessible by the Vulkan renderer.

**Details**
`add_splat` migrates the new model into the viewer allocator, but `keepFloatShNInPooledCuda` deliberately parks float32 `SplatData.shN` in pooled CUDA as a q16 workspace rather than importing a float rest buffer into Vulkan. VkSplat can't bind that workspace.

**Fix**
Call `apply_shN_value_quant` after migration and before `addSplat`, so shN becomes Vulkan-visible q16 codes plus bounds.

### DIFF
--- a/src/python/lfs/py_scene.cpp
+++ b/src/python/lfs/py_scene.cpp
@@ -9,6 +9,7 @@
 #include "core/path_utils.hpp"
 #include "core/property_registry.hpp"
 #include "io/loader.hpp"
+#include "lfs/training/sh_value_storage.hpp"
 #include "python/python_runtime.hpp"
 #include "visualizer/gui_capabilities.hpp"
 #include "visualizer/operation/undo_entry.hpp"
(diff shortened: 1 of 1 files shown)


MrNeRF:
I thin this is lready fixed on master since #1844 (Aug 25): the same encode runs inside migrateSplatTensorsToAllocator, which add_splat calls. Your line always hits the already-quantized early exit on current master. I reproduced your symptom only after removing the existing encodes by hand, and your line does fix it there, so you likely tested a pre Aug-25 build, no? Maybe I am missing something? Otherwise I would close it.

thomas-koch-core11:
Yes, I tested a pre Aug-25 build.
Just tested, #1844 fixes it.

PR #1871, opened 2026-08-28
Fix/python callback gil lifetime

thomas-koch-core11:
Two independent fixes in the Python bindings, plus a regression test. Both are cases where native code owned an `nb::object` and could release it on a thread that does not hold the GIL.

`py_store.cpp`-fix:
The notification lambda's `weak_ptr::lock()` sat above the `GilAcquire` scope, so its local `shared_ptr` was released after the GIL had been dropped. This occurs when a Python callback unsubscribes itself.

`module.cpp`-fix:
Native hook registries may copy and destroy callbacks off-GIL. Callbacks are now held behind `shared_ptr<nb::object>` with a
GIL-acquiring deleter, with only the pointer captured.

### DIFF
--- a/src/python/lfs/module.cpp
+++ b/src/python/lfs/module.cpp
@@ -104,6 +104,7 @@
 #include <fstream>
 #include <functional>
 #include <future>
+#include <memory>
 #include <sstream>
 #include <string>
 #include <unordered_map>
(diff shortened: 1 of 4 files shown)


MrNeRF:
Alright, thx! Found even another similar issue.

PR #1882, opened 2026-08-28
Improve the Plugin Marketplace experience

shadygm:
## Summary

- Add a UI design language guide covering windows, panels, controls, dialogs, spacing, and theme usage.
- Modernize the Plugin Marketplace with unified styling, responsive gallery cards, a denser list view, and clearer status and confirmation states.
- Prevent long plugin names, paths, tags, descriptions, and statuses from overlapping while preserving existing marketplace functionality.

## How to test

- Build and launch LichtFeld Studio.
- Open **Tools → Plugin Marketplace**.
- Resize the window and confirm gallery cards reflow while remaining aligned with both sides of the manual-install section.
- Switch between gallery and list views and verify long text truncates cleanly without overlapping.
- Verify filtering, sorting, list expansion, manual URL installation, and installed-plugin actions.
- Review `docs/docs/development/ui-design-language.md`.

New look and feel: 
image

image


MrNeRF:
That's great. Haven't reviewed yet but this might get towards master or I will just cherry pick later.

MrNeRF:
Really nice work, the new marketplace looks great. I went through the whole thing hands-on (resize, both views, install/uninstall, themes) and it all works. Two things need fixing before merge though:

1. `test_plugin_marketplace_uses_dirty_update_policy` fails now — it still asserts the old "dirty" policy. Just update it to interval/250 like the asset manager test does.

2. The card grid breaks on high DPI. The old code divided the measured width by `lf.ui.get_ui_scale()`, the new code doesn't but still writes the value back as dp. So at scale 1.5 everything is sized twice and cards get clipped at the panel edge. Restoring the divide fixes it.

Smaller stuff, take it or leave it:
- Python still emits `mp-warning-text` (lines 644/720) but the class got renamed to `mp-warning-note` in the css, so the stars/downloads in list view lost their orange color.
- The status text gets rewritten every 250ms tick even when nothing changed, which keeps the window repainting while idle. Guard the `set_text`/`set_attribute` calls with a "did it change" check and the panel can sleep again. The interval itself is fine — it's actually what makes resize reflow work.
- The "Supported formats" content still picks up the shared `.section-content` style (negative margins + 4dp bar), so the format lines don't line up with the URL input. Small local override fixes it.
- `.card-grid` is missing from the `.hidden` overrides you added for the other elements.

The doc is a great addition btw. Funny detail: it says "author dimensions in dp, reserve px for runtime values" — which is exactly the rule the grid bug above breaks :)

PR #1895, opened 2026-08-29
Add optional lazy-loaded NVIDIA DLSS scene reconstruction

francescofugazzi:
## Summary

Add NVIDIA DLSS Super Resolution as an optional scene-reconstruction backend for the Vulkan viewport.

The integration preserves the existing Native, Spatial, and Temporal reconstruction contracts while keeping NVIDIA NGX outside the main executable and `lfs_visualizer`. DLSS is provided through a separately built plugin, discovered from an application-owned path and activated only when selected.

## External plugin architecture

The integration introduces a versioned, vendor-neutral C ABI between LichtFeld Studio and optional scene-reconstruction plugins.

The boundary uses:

- fixed-width scalar values and versioned structures;
- opaque plugin and per-view handles;
- explicit ownership and lifecycle operations;
- caller-provided diagnostic buffers;
- Vulkan handles represented without exposing vendor types;
- no C++ STL objects, exceptions, allocators, NGX types, or implementation-specific ownership across the module boundary.

The NVIDIA implementation is built as:

```text
Windows:
scene_upscalers/nvidia/lfs_scene_upscaler_nvidia_dlss.dll

Linux:
lib/scene_upscalers/nvidia/liblfs_scene_upscaler_nvidia_dlss.so
```

The corresponding NVIDIA runtime is staged beside the plugin rather than linked into or placed arbitrarily beside the main application.

Dynamic discovery uses exact application-relative paths. Windows uses a restricted DLL search policy, while Linux opens the explicit plugin path. The current working directory and general system search path are not used to discover optional backends.

## Two-stage lazy loading

DLSS initialization is split into two stages because NGX must declare required Vulkan extensions before instance and device creation.

During bootstrap, LichtFeld loads only the external plugin module to:

- validate the ABI and plugin identifier;
- obtain required Vulkan instance extensions;
- obtain required Vulkan device extensions;
- retain static bootstrap metadata.

Bootstrap does not initialize the NGX runtime, query DLSS optimal settings, create DLSS features, allocate output images, or create per-view GPU state.

The complete NGX runtime is initialized lazily when NVIDIA DLSS is selected. Feature support, optimal render extents, per-view features, output resources, and runtime-owned history are created only at that point.

Safe mode does not discover or load optional scene-reconstruction plugins.

## Runtime catalog and preferences

The optional backend is registered with the stable identifier:

```text
nvidia-dlss
```

It exposes backend-specific presets:

- Quality
- Balanced
- Performance

The catalog values provide stable bootstrap scales for UI and persistence. Once NGX is active, the exact render extent is obtained from the vendor optimal-settings query for the selected output size and preset. Results are cached by output extent and quality to avoid repeating the vendor query every frame.

The backend and remembered preset participate in the existing reconstruction contracts used by:

- Preferences;
- Python UI bindings;
- MCP rendering settings and catalog discovery;
- preference reset;
- safe mode;
- scene-reconstruction status reporting;
- the live reconstruction benchmark.

Unavailable optional identifiers are rejected by mutation APIs instead of being silently accepted.

## Temporal input contract

DLSS consumes the reviewed Temporal reconstruction inputs already used by the built-in backend:

- the exact projection jitter applied to the rendered color input;
- current-to-previous motion in top-left pixel coordinates;
- independent main, split-left, and split-right view identities;
- explicit camera-cut, resize, projection, scene-generation, backend, and preset transitions;
- stable render and output extents;
- per-view history keys.

Switching between Temporal and DLSS changes the history key, preventing one backend from inheriting incompatible history from the other.

Perspective views use the exact applied jitter. Orthographic views remain unjittered while retaining valid depth and motion inputs.

## Depth, motion, and image resources

VkSplat publishes positive linear view depth, while NGX consumes raster depth. A dedicated Vulkan compute pass converts each valid rendered subregion into normalized, non-inverted `R32_SFLOAT` depth:

```text
near = 0
far  = 1
```

The DLSS pipeline uses:

- perceptually encoded LDR `RGBA8` color;
- `RG16F` low-resolution motion vectors;
- normalized `R32F` raster depth;
- full-resolution `RGBA8` output.

Valid render extents, allocation extents, image layouts, and synchronization boundaries are propagated explicitly through VkSplat, CUDA/Vulkan interop, motion production, depth conversion, DLSS evaluation, presentation, split view, and viewport capture.

## Viewport and split-view integration

DLSS is integrated into the normal viewport and split presentation paths.

Each logical view owns an independent NGX feature and associated resources. Split rendering is transactional: both panels must resolve successfully through DLSS or both panels remain on Native presentation. A partially reconstructed split frame is never displayed.

Backend changes, preset changes, resize, camera cuts, scene replacement, and invalidated temporal contracts reset or recreate only the affected state.

Output dimensions below NGX's 32-by-32 minimum use a temporary Native presentation without latching a backend failure. Returning to a supported extent restores the requested DLSS path automatically.

## Failure and fallback behavior

Requested backend, effective backend, runtime readiness, and fallback state remain separate.

If the LichtFeld-owned plugin is absent, NVIDIA DLSS is omitted from the runtime catalog.

If the plugin is present but the NVIDIA runtime is missing, incompatible, or unsupported:

- the DLSS request remains remembered;
- the effective presentation falls back atomically to Native;
- the NGX failure reason is reported;
- the effective DLSS-to-Native fallback is reported separately;
- the failure is latched rather than retried every frame;
- selecting Native and then NVIDIA DLSS performs an explicit retry.

The Native fallback is never persisted over a valid DLSS preference.

Loaded modules remain resident until ordered application shutdown. Selecting Native releases completed DLSS GPU resources but does not attempt an unsafe module unload while vendor callbacks or runtime state may still exist.

## Build configuration

Ordinary developer builds remain vendor-free by default:

```text
LFS_ENABLE_NVIDIA_DLSS=OFF
```

To enable the external plugin, obtain the SDK from:

```text
https://github.com/NVIDIA/DLSS
```

and configure with:

```sh
git clone https://github.com/NVIDIA/DLSS external/nvidia-dlss-sdk

cmake -S . -B build \
  -DLFS_ENABLE_NVIDIA_DLSS=ON \
  -DLFS_NVIDIA_DLSS_ROOT=/path/to/NVIDIA-DLSS-SDK
```

CMake does not download the SDK or accept its license during an ordinary developer configuration.

Explicitly enabled builds require a complete SDK checkout. Missing headers, platform libraries, or runtime artifacts produce an actionable configuration error containing the official SDK link instead of silently producing a build without the requested backend.

Debug builds use the SDK development import library and runtime. Release, RelWithDebInfo, and MinSizeRel builds use the corresponding release artifacts.

## Portable distribution

Portable builds force the optional NVIDIA plugin on.

The Windows nightly portable workflow checks out `NVIDIA/DLSS` at a pinned revision with Git LFS before configuration, builds the external plugin, and stages the matching NVIDIA runtime in the package.

A missing or incomplete SDK is fatal for a portable configuration so that distribution artifacts cannot silently omit the advertised backend.

The LichtFeld-owned plugin remains small and independent. The larger `nvngx_dlss.dll` is the unmodified NVIDIA runtime and is distributed subject to NVIDIA's SDK and runtime redistribution terms.

## Documentation

The scene-reconstruction and build documentation now describes:

- the external plugin ABI;
- bootstrap versus runtime activation;
- application-owned loading paths;
- depth, motion, jitter, and history contracts;
- backend and preset persistence;
- fallback and explicit retry behavior;
- SDK acquisition and local opt-in configuration;
- Debug and Release runtime selection;
- portable packaging and redistribution responsibilities;
- production-path reconstruction benchmarking.


MrNeRF:
Went deep on this one — read everything, built it against the real SDK and ran it live on a 4080 under Linux. It works: plugin discovered, NGX initializes lazily, requested/effective/fallback honest, presets switch cleanly, the saved DLSS preference survives restart, and switching temporal <-> dlss doesn't leak history across backends. Also checked all the temporal contract points from #1725 — motion direction, jitter identity, camera cut, reset behavior — they're all correct here. Really solid work, especially the ABI split.

I found a handful of issues and pushed fixes directly to your branch (3 commits) instead of a comment ping-pong, hope that's fine:

1. NGX wants its inputs in the Sample Image state and *restores* that state after evaluate. We handed motion + converted depth over in GENERAL and re-barriered them from GENERAL next frame, so host tracking and NGX disagree about the layout after the first evaluate. Both images now go to SHADER_READ_ONLY_OPTIMAL after their compute passes (temporal resolve binds the shared motion image the same way), and the plugin rejects GENERAL inputs.

2. The plugin was staged at build/scene_upscalers/nvidia, but on multi-config generators (VS) the exe lives in build/Release/ — so the loader never found it. Same story for Fedora-style lib64 installs. It now follows the exe dir and installs to plain lib/.

3. BUILD_PORTABLE=ON force-enabled DLSS and died fatally without an SDK checkout — for everyone, not just the nightly. An explicit -DLFS_ENABLE_NVIDIA_DLSS=OFF now wins (nightly passes ON explicitly, so nothing changes there).

4. Licensing: the plugin .cpp and the ABI header were GPL while statically linking NGX — NVIDIA's SDK terms (§4.e) don't allow that combination in a distributed binary. Both files are MIT now, host stays GPL, so the shipped module is clean. Flagging this openly since it touches your files' headers — shout if you disagree.

5. Smaller ones: a failed feature create leaked the NGX handle; a one-frame input hole (e.g. an interop timing miss) latched the Native fallback until you manually reselected DLSS — now it's just a quiet warmup frame; independent split panels reused the full-viewport optimal-settings answer instead of querying per panel; drag-resize recreated the NGX feature (with a full pipeline wait) on every size step — DLSS now pauses during the drag and resumes once; the python settings path accepted backends that aren't loaded; and the catalog assumed dlss is the last registry entry.

Also added CPU tests for the two things that would regress silently: the reset-flag mapping (a camera-cut on ordinary orbit would make DLSS useless) and the jitter round-trip.

Tests are green (72 upscaler/temporal/dlss tests, full suite matches the dev baseline), and I re-verified everything live after the fixes.

francescofugazzi:
Of course it's very fine anytime you push fixes :)
I synced with dev and fix merging conflicts

MrNeRF:
Thank you!
Oh, I thought I just do it as you are currently not available. Sorry about that. I wanted to play a bit with it and also have some metrics. Will post later.

Issue #1864, opened 2026-08-27
Training degenerates after 2,000–3,000 iterations on RTX 5070 Ti


shadygm:
### Description

On the current `master` branch, training the `kitchen` and `bonsai` scenes from the Mip-NeRF 360 dataset with the default parameters on an NVIDIA GeForce RTX 5070 Ti starts normally. After approximately 2,000–3,000 iterations, the reconstructed scene degenerates and develops severe visual artifacts.

### Steps to reproduce

1. Run LichtFeld Studio from `master` at commit `f04fd20d6b5a`.
2. Load either the `kitchen` or `bonsai` scene from the Mip-NeRF 360 dataset.
3. Leave all training parameters at their defaults.
4. Start training on an NVIDIA GeForce RTX 5070 Ti.
5. Continue training for approximately 2,000–3,000 iterations.

### Actual behavior

The scene initially trains normally, but then visibly degenerates and produces severe artifacts after approximately 2,000–3,000 iterations.

### Expected behavior

Training should remain stable and continue improving the reconstruction without the scene degenerating or developing artifacts.

### Additional observation

This may be related to spherical harmonics in scenes with bright colours, but that is only a hypothesis and has not been confirmed.

Training Start: 

<img width="2191" height="1281" alt="Image" src="https://github.com/user-attachments/assets/550091f4-bbb0-4e9a-bbb4-365942b8435a" />

Scene Degeneration: 

<img width="2191" height="1281" alt="Image" src="https://github.com/user-attachments/assets/d27faf56-adb9-4a98-97db-0999e8ddd0c8" />

MrNeRF:
Is this only in the gui? If you save the scene does it look good? You are training on full resolution? Because I can't reproduce it unfortunately :(

shadygm:
Training on full resolution, not gui only, exporting to supersplat shows the same thing. I'll investigate it a bit myself. 

<img width="1215" height="1225" alt="Image" src="https://github.com/user-attachments/assets/8f565488-d309-4ee5-b8f1-f10096360911" />

shadygm:
Resolved by #1920 and #1924

PR #1913, opened 2026-08-31
Fix unstable spherical harmonics updates during training

shadygm:
## Summary

New SH bands had fresh optimizer moments but inherited an old Adam timestep. This over-scaled their first updates during SH level changes, matching the instability reported in #1864.

## Changes

- Track separate Adam step counters for each SH band.
- Apply band-local corrections in fused and standalone optimizer paths.
- Freeze inactive packed SH cells and their optimizer moments.
- Preserve checkpoint compatibility with versions 1–3.
- Add regression coverage for progressive SH activation.


shadygm:
wrong fix

PR #1923, opened 2026-08-31
Fix GUI evaluation split initialization

shadygm:
## Summary

  - Refresh camera train/evaluation splits when GUI evaluation settings change before training starts.
  - Apply the selected Test Every value on the first run.
  - Preserve existing camera splits when evaluation settings are unchanged.


MrNeRF:
checked, tested, looks good to me. Thank you!

PR #1932, opened 2026-08-31
Add per-axis depth-window sizing and drag-to-draw controls

alexmgee:
Implements step 2 of the three-PR sequence proposed in #1845, following #1860.

### The problem

Step 1 gave the depth window a stated footprint on screen, but that footprint still had one scale. Its height followed its width through the viewport aspect ratio, so it could only be a scaled copy of the viewport. The only way to change it was through the Size, X and Y sliders. You could say how much of the screen the window covered, but not choose a differently shaped window or place that shape by hand.

This PR splits that scale into independent X and Y fractions and makes the window directly manipulable. Shift+Alt+drag draws a new rectangle. Its corners and edges resize it, and a central region moves it. The same per-axis values now flow through settings, IPC, CPU and CUDA containment, and the Slang overlay, so the rectangle being manipulated is the rectangle used to commit the selection under pinhole, orthographic and equirectangular cameras.

### Where this is going

This is the second of three steps, kept separate so each remains reviewable:

1. #1860 established the screen-space window and projection contract.
2. **This PR** adds per-axis size and direct manipulation.
3. **Later** adds per-panel windows in split view with a sync toggle.

Step 3 is designed and in the process of being built. Keeping it out avoids combining the input, rendering and API changes here with a second state-ownership change.

### What is in this PR

- The single depth-window scale split across render settings, IPC, sanitization, selection filtering, CUDA and shader containment, with a new parameter lane that does not renumber existing lanes.
- A Shift+Alt drag operator with draw, move, edge/corner resize, Ctrl constraints, undo, cursors, handles, live reveal and cleanup on commit, cancel, modifier release, focus loss, operator replacement, GT comparison and shutdown.
- `DEPTH_WINDOW_DRAG` appended as action 86, bound in selection-tool mode for both Shift+Alt+left-drag and Ctrl+Shift+Alt+left-drag, exported through C++ and Python, named in all ten locales, and migrated in keymap profile version 27 with one final version write.
- Python and toolbar support for both axes, including the widened getter, compatible setter, controller echo handling and reference-relative Size readout.

The complete change is 69 files, 3,654 insertions and 257 deletions across ten commits.

### Design notes

Size is deliberately reference-relative. The shape most recently drawn is 100%; every committed draw establishes a new reference, even when its ratio matches the previous shape, and a ratio-changing resize also re-bases. Pure proportional scaling does not: a window scaled to 200% remains 200% after release.

The live reveal is containment, not a provisional selection colour. During a drag, eligible splats inside the moving window keep their normal appearance against the dimmed outside; committed splats keep their selection colour, there is no tint, and the stale exact preview is suppressed. The exact CUDA-filtered result replaces it at release. Rebuilding that exact result measured about 43 ms per refresh on a representative scene, so drag frames instead bypass the selection-overlay fast path and run full renders at the normal interactive cadence.

The move target is proportional rather than the whole interior: 40% of the box on each axis, with a 24 px minimum and a 50% cap. Handles retain precedence. The remaining interior shows a crosshair and begins a fresh draw.

Ctrl held at draw onset makes a 1:1 screen-pixel square. Pressing it during a draw captures the live ratio; corner resizing constrains to the shape being resized. Edge resizing remains free and one-dimensional with or without Ctrl, and releasing Ctrl returns to free-form movement. Both modifier triggers are registered so entering the gesture with Ctrl already held works rather than relying on a mid-drag transition. A fresh draw begins once the pointer has moved at least 5 px from the press; a click, with or without Ctrl, creates nothing and leaves the existing window untouched.

The Python getter groups `scale_x` and `scale_y` together for natural reading. The setter appends optional `scale_y` after the existing six positional arguments. That asymmetry is deliberate: reads have the new shape, while existing isotropic writes keep their positional meaning.

### GT comparison

While GT comparison is active, the entire depth-window feature is suppressed: the gesture cannot start, an active drag is cancelled, the rectangle and handles draw on no panel, hover and cursor feedback stay inert, toolbar sliders hide, and dim/hide filtering is suspended. Window state is not mutated and returns exactly when comparison ends.

This boundary does not alter the shipped GT selection contract, including projection onto the presented GT view and depth filtering in the selection lane. It suspends only the depth window's surfaces and render effect.

The broader interaction is deferred. The intended form includes one unclipped window over the GT panel, live sliders, and potentially direct manipulation, but the drag operator still needs one authoritative mapping from panel coordinates to the presented GT content. Entering Edit mode discards the required dataset, while programmatic settings can still co-activate the features. Models loaded through a dataset `init_path` are represented as a point-cloud preview until training starts (#1905), so the combined workflow cannot be verified end to end through a supported path. It can be restored later, possibly with step 3, once those ownership and mapping questions have one answer.

### Defects found and fixed before submission

Two defects surfaced during hands-on verification and were fixed here rather than documented around.

**The frozen reveal.** Selection-only frames take a fast path that re-rasters cached overlay flags. The drag changes containment flags in the projection pass, so the rectangle moved while the reveal appeared frozen. Per-frame instrumentation traced the frame through that cache path. Active depth-window drags now bypass it, forcing the projection pass to refresh the live containment reveal; ordinary selection-only frames retain the fast path.

**The axis mix-up.** One C++ apply path passed the X scale for both axes after correctly clamping X and Y separately. An anisotropic write could therefore become isotropic at the final settings update. The path now passes the Y value in the Y slot, with an exact regression test that writes unequal axes and checks their positions independently.

### Interface changes and migrations

1. **The single scale becomes two axes.** `depth_filter_scale` and `SelectionScreenWindow::scale` are replaced by `depth_filter_scale_x`/`depth_filter_scale_y` and `scale_x`/`scale_y`. This lands now because the interface arrived in the immediately preceding, unreleased step and per-axis sizing was already announced. **Migration:** read and write both axes; use equal values for the old isotropic result.
2. **`get_depth_filter_window()` returns seven values.** The order is `(enabled, near, far, scale_x, scale_y, offset_x, offset_y)`; offsets move from slots 4/5 to 5/6. The repository's sole caller and the local plugin tree were audited. The getter groups axes while the setter appends optional `scale_y` for positional compatibility. **Migration:** unpack seven values; existing setter calls remain valid.
3. **Size has a new basis.** It is relative to the last drawn shape, with every draw and ratio-changing resize re-basing to 100%, while proportional scaling tracks. This is the useful basis once shape is editable. **Migration:** no code change; the displayed percentage now describes the drawn reference.
4. **GT comparison suppresses the whole depth-window feature, and the gesture defines new cursor regions.** State survives unchanged, while Shift+Alt suppresses the selection cursor and `SEL +` ring/label, shows crosshair, resize or move cursors as appropriate, and uses the proportional move zone above. Defining these together makes gesture ownership unambiguous. **Migration:** none.
5. **Fresh polling is the GT-entry boundary.** A cached `OperatorRegistry::poll()` result can briefly be stale after a GT toggle because rendering/GT state was not added as a poll dependency; fresh entry polls and modal cancellation are guarded, and no current user-facing path depends on the stale value. **Migration:** none.
6. **Drag-time reveal is approximate.** It is the no-tint containment reveal described above, with the exact CUDA-filtered result at release and full drag-frame renders replacing a roughly 43 ms synchronous exact rebuild. **Migration:** none; an asynchronous exact rebuild remains a possible follow-up.
7. **Some visual feedback is hands-on tested only.** Solid accent handle squares, the center glyph, cursor transitions, and selection ring/label suppression have no image-based assertions. **Migration:** none.
8. **`SelectionScreenWindow` gains renderer-interface fields.** The installed C++ struct carries `scale_x`, `scale_y` and internal `drag_preview`. The shader parameter-table layout is unchanged, so an old/new shader mix loses the approximation rather than moving lanes; the struct change is still an ABI consideration. **Migration:** rebuild C++ consumers, initialize both scales, and leave `drag_preview` false outside direct manipulation.

### Verification

Automated coverage exercises parameter layout and preview packing; unequal-axis CPU/CUDA containment across all camera models, including off-centre and asymmetric intrinsics; move-zone regimes and handle precedence; draw, move, corner/edge resize and Ctrl transitions; commit, cancel, undo, modifier release, focus loss, replacement and GT-toggle cleanup; brush-preview invalidation even when its point count is unchanged; the GT transition hook without disturbing unrelated modals; both migrated key triggers and the single version stamp; the seven-value API; controller echo handling; and all Size re-basing cases.

The complete hands-on run passed 12/12 blocks: draw and live reveal; handles and anchoring; Ctrl constraints; ordinary selection gestures; isotropic scaling and the reference-relative readout; undo; GT suppression/restoration; cancel and commit semantics; shared-window split view; orthographic and equirectangular cameras; visualization and toolbar regimes; and restart/default behaviour. The review-round changes were re-checked hands-on afterwards: clicks and Ctrl clicks create nothing, small and constrained draws land at the anchor, scroll stops as a whole at the limits with the Size readout tracking, undoing a draw restores the readout, and the unchanged gestures behave as before.

Built on Windows, Release, with fresh binaries and shader outputs. The application and tests builds completed with zero errors. **Targeted C++ — 190/190 pass, 0 skip. Python controller, keymap and panel suites — 104/104 pass.** Formatting and Python syntax checks are clean. This is targeted coverage, not a claim that every repository suite was run.

### Known limitations

- A cached operator poll can be briefly stale across a GT toggle as described above.
- Handle appearance, cursor transitions and selection-ring suppression are verified hands-on rather than by visual assertions.
- Selection edits ride temporal accumulation, a pre-existing renderer characteristic. Active drag frames force full renders for the live reveal.

### Follow-ups

Step 3 adds per-panel windows and a sync toggle. The deferred GT interaction can then be reconsidered, including whether its panel mapping should expose the selection resolver, duplicate the mapping, or move all consumers to a shared helper.

Two upstream follow-ups found during this work are outside this PR. Leaving crop mode can leave the app store's active tool at Selection while the native tool registry is empty, so a subsequent depth-field commit raises a Selection error modal. And since #1905 deliberately represents a Gaussian dataset init PLY as a point-cloud preview until training starts, a feature request proposes an opt-in, viewer-owned splat preview for faithful pre-training comparison.


MrNeRF (changes requested):
I read the whole diff. This is a good PR: the description matches the code, the per-axis plumbing is consistent everywhere I checked (CPU, CUDA and shader use the same containment formula, the new shader lane is appended so nothing renumbers, the keymap migration follows the existing pattern, the Python setter stays backward compatible), and the test coverage is real.

One thing must be fixed before merge:

- An interrupted drag can leave changes behind with no undo. When another modal operator replaces the drag mid-gesture, the registry destroys the operator without calling cancel() (operator_registry.cpp, active_modal_ = std::move(op)). The destructor only clears the preview latch; it neither restores the old window nor pushes an undo entry. So a half-finished drag silently becomes permanent and Ctrl+Z cannot take it back. Either restore the before state or push the undo entry on that path. The replacement test only checks the preview flag, so it misses this.

Small, would be nice in the same PR:

- Scroll zoom distorts the shape at the limits. adjustWindowScale() clamps X and Y separately at 0.05/1.0, so a wide window that hits the 100% wall on one axis keeps growing on the other. The Python slider computes a common safe factor; the scroll path should too.

Fine as follow-ups:

- After undoing a draw with the same ratio, the Size readout keeps the undone draw as its 100% reference (undo does not bump depth_window_draw_generation).
- A Ctrl click without movement ends up as a 5% x 5% window through the sanitizer floor, which is not a square on a non-square viewport.

Fix the undo leak and this is good to merge.

MrNeRF:
Thank you! Looks fine!

PR #1938, opened 2026-09-01
Autosave no longer rewrites unchanged projects in a loop

MrNeRF:
The app was quietly burning CPU on autosaves that saved nothing new.

Two cases, both measured on an idle session:
- After importing a dataset, the app wrote a full project save every 1.2 seconds, forever — about 50 saves and 50 preview-image re-encodes per minute, with zero changes.
- With a project open, the autosave sidecar was republished every timer tick indefinitely, even though its content never changed.

Now the autosave scheduler checks whether anything actually changed since the last successful write (scene edits, parameter edits, dirty flags). If nothing did, it does nothing. A real edit still autosaves within the usual few seconds, and crash recovery behaves exactly as before. Bonus: parameter-only edits now count toward autosave eligibility, which they previously never did.

Verified live: one write after dataset import then silence; zero writes on an idle open project; an edit triggers a sidecar within 5 seconds and then it goes quiet again. Project-document and format test suites pass.

### DIFF
--- a/src/io/project/project_document.cpp
+++ b/src/io/project/project_document.cpp
@@ -3216,6 +3216,7 @@ namespace lfs::io::project {
         std::span<const std::byte> preview_png = options.preview_png;
 #if !defined(LFS_FORMAT_TEST_TARGET)
         if (!is_autosave &&
+            !options.leave_unbound &&
             (options.commit.kind == CommitKind::Explicit ||
              options.commit.kind == CommitKind::Recovered)) {
             const auto project_root =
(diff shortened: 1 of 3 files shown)


MrNeRF:
Note on the measured cadence: this box has autosave_interval_seconds=1 in its local config (left over from E2E rigs), which is why the loop ticked every ~1.2s. With the default interval the waste is proportionally smaller but the defects are the same: the dataset-session path re-encoded the preview image and rewrote every chunk on each tick, and the sidecar was republished with identical content on every timer expiry.

PR #1939, opened 2026-09-01
Add optional lazy-loaded AMD FSR 3.1 scene reconstruction

francescofugazzi:
## Summary

Add AMD FSR 3.1 as an optional scene-reconstruction backend for the Vulkan viewport.

The integration preserves the existing Native, Spatial, Temporal, and NVIDIA DLSS reconstruction contracts while keeping FidelityFX outside the main executable and `lfs_visualizer`. FSR is provided through a separately built plugin, discovered from an application-owned path and activated only when selected.

## External plugin architecture

The integration reuses and append-only extends the versioned, vendor-neutral scene-upscaler C ABI.

The boundary continues to use:

- fixed-width scalar values and versioned structures;
- opaque plugin and per-view handles;
- explicit ownership and lifecycle operations;
- caller-provided diagnostic buffers;
- Vulkan handles represented without exposing vendor types;
- no C++ STL objects, exceptions, allocators, or FidelityFX types across the module boundary.

The ABI extension adds the camera values required by vendor temporal reconstruction while preserving the prefix consumed by the existing NVIDIA DLSS plugin.

The AMD implementation is built as:

```text
Windows:
scene_upscalers/amd/lfs_scene_upscaler_amd_fsr3.dll

Linux:
lib/scene_upscalers/amd/liblfs_scene_upscaler_amd_fsr3.so
```

FidelityFX FSR 3.1 and its Vulkan backend are linked only into this optional plugin. The main executable and `lfs_visualizer` do not link against the FidelityFX SDK.

Dynamic discovery uses the exact application-relative plugin path. The current working directory and general system search path are not used to discover the optional backend.

## Lazy runtime activation

During Vulkan bootstrap, LichtFeld loads only the external plugin module to:

- validate the ABI and plugin identifier;
- collect any required Vulkan bootstrap metadata;
- retain the validated plugin interface;
- register the optional runtime backend.

Bootstrap does not initialize the FidelityFX Vulkan backend, allocate SDK scratch memory, create FSR contexts, allocate full-resolution outputs, or create per-view shared resources.

The complete FidelityFX runtime is initialized lazily when AMD FSR 3.1 is selected. Render-resolution queries, per-view contexts, shared resources, depth and motion resources, and output images are created only when required.

Safe mode does not discover or load optional scene-reconstruction plugins.

## Runtime catalog and preferences

The optional backend is registered with the stable identifier:

```text
amd-fsr3
```

It exposes the FidelityFX reconstruction presets:

- Quality
- Balanced
- Performance

The catalog provides stable bootstrap scales for UI and persistence. Once the FidelityFX runtime is active, the exact render extent is obtained from the SDK quality-mode query for the selected output size and preset.

The backend and remembered preset participate in the existing reconstruction contracts used by:

- Preferences;
- Python UI bindings;
- MCP rendering settings and catalog discovery;
- preference reset;
- safe mode;
- scene-reconstruction status reporting;
- the live reconstruction benchmark.

Unavailable optional identifiers are rejected by mutation APIs instead of being silently accepted.

## Temporal input contract

FSR consumes the reviewed temporal reconstruction inputs already used by the built-in and DLSS backends:

- the exact projection jitter applied to the rendered color input;
- current-to-previous motion in top-left pixel coordinates;
- normalized raster depth;
- independent main, split-left, and split-right view identities;
- explicit camera-cut, resize, projection, scene-generation, backend, and preset transitions;
- stable render and output extents;
- per-view history keys.

VkSplat depth is rasterized with the jittered camera projection. FSR motion is therefore reconstructed from the same current and previous jittered projection pair, and the plugin enables the FidelityFX motion-vector jitter-cancellation contract.

The dispatch jitter represents the actual camera-projection sample and is independent from texture storage or presentation orientation.

Motion is supplied in pixels, with a unit SDK scale. FidelityFX converts that displacement to its normalized internal representation using the active motion-vector extent.

Switching between Temporal, NVIDIA DLSS, and AMD FSR 3.1 changes the history key so that one backend cannot inherit incompatible history from another.

## Reconstruction convergence

Viewport convergence follows the FidelityFX jitter phase count derived from the actual render-to-output ratio.

The standard presets therefore settle across:

- 18 samples for Quality;
- 23 samples for Balanced;
- 32 samples for Performance.

The Halton sequence wraps at the corresponding SDK phase length instead of stopping after the built-in backend’s default eight samples.

Each FSR evaluation consumes a newly published color and depth generation rendered with the matching jitter sample. GUI-only presentation frames reuse the last resolved output rather than redispatching unchanged color, depth, and frozen jitter through FidelityFX.

Camera cuts, backend changes, preset changes, render-size changes, scene replacement, failures, and explicit resets invalidate the cached result and the corresponding SDK history.

## Depth, motion, and image resources

The FSR pipeline uses:

- perceptually encoded LDR `RGBA8` input color;
- `RG16F` low-resolution motion vectors;
- normalized, non-inverted `R32F` raster depth;
- full-resolution `RGBA8` output;
- neutral pre-exposure;
- the exact camera near plane, far plane, and vertical field of view.

The plugin uses fixed render and output extents. Contexts are recreated when those extents change instead of advertising unsupported dynamic-resolution behavior.

The optional RCAS post-pass is disabled. Quality, Balanced, and Performance control the FidelityFX reconstruction ratio without adding a second sharpening operation to high-frequency Gaussian detail.

Reactive and transparency-and-composition masks remain absent until the renderer can provide semantically correct material signals. Aggregate Gaussian coverage is not treated as a substitute for those inputs.

LichtFeld does not define a canonical metre-per-scene-unit scale, so the SDK receives the neutral `1.0` view-space scale.

Valid render extents, allocation extents, source generations, image layouts, and synchronization boundaries are propagated explicitly through VkSplat, CUDA/Vulkan interop, motion production, depth conversion, FSR evaluation, presentation, split view, and viewport capture.

## Viewport and split-view integration

AMD FSR 3.1 is integrated into the regular viewport, training viewport, Independent Dual split presentation, and PLY comparison paths.

Each logical view owns an independent FidelityFX context and associated resources. Split rendering is transactional: both panels must resolve successfully through FSR or both panels remain on Native presentation. A partially reconstructed split frame is never displayed.

The resolved Vulkan image layout is carried by the FSR result instead of being inferred by presentation consumers.

Equirectangular projection, orthographic projection, and appearance-correction presentation remain on Native because their projection or camera contracts do not match the current FidelityFX integration.

Output dimensions below the SDK’s 32-by-32 minimum use a temporary Native presentation without latching a backend failure. Returning to a supported extent restores the requested FSR path automatically.

## Failure and fallback behavior

Requested backend, effective backend, runtime readiness, and fallback state remain separate.

If the LichtFeld-owned plugin is absent, AMD FSR 3.1 is omitted from the runtime catalog.

If the plugin is present but its ABI, FidelityFX runtime, Vulkan device, resources, or evaluation path is unavailable:

- the FSR request remains remembered;
- the effective presentation falls back atomically to Native;
- the FidelityFX failure reason is reported;
- the effective FSR-to-Native fallback is reported separately;
- the failure is latched instead of retried every frame;
- selecting Native and then AMD FSR 3.1 performs an explicit retry.

The Native fallback is never persisted over a valid AMD FSR preference.

Loaded modules remain resident until ordered application shutdown. Selecting Native releases completed FSR GPU resources without attempting an unsafe module unload while SDK callbacks or runtime state may still exist.

## Build configuration

Ordinary developer builds remain vendor-free by default:

```text
LFS_ENABLE_AMD_FSR3=OFF
```

To enable the external plugin, obtain FidelityFX SDK v1.1.4 from:

```text
https://github.com/GPUOpen-LibrariesAndSDKs/FidelityFX-SDK/releases/tag/v1.1.4
```

and configure with:

```sh
git clone --branch v1.1.4 --recurse-submodules \
  https://github.com/GPUOpen-LibrariesAndSDKs/FidelityFX-SDK \
  external/fidelityfx-sdk

cmake -S . -B build \
  -DLFS_ENABLE_AMD_FSR3=ON \
  -DLFS_AMD_FSR3_ROOT=/path/to/FidelityFX-SDK
```

CMake does not download the SDK or accept its license during an ordinary developer configuration.

Explicitly enabled builds require a complete SDK checkout. Missing headers, license material, or Vulkan FSR libraries produce an actionable configuration error containing the official SDK link instead of silently producing a build without the requested backend.

On Windows, `LFS_AMD_FSR3_BUILD_SDK=ON` builds the required static FSR upscaler and Vulkan backend libraries from an isolated copy of the user-provided SDK when usable libraries are not already available.

Prebuilt Windows or Linux libraries can be supplied through:

```text
LFS_AMD_FSR3_LIBRARY_DIR
```

## Portable distribution

Portable configurations enable the optional AMD FSR 3.1 plugin by default while retaining an explicit opt-out.

The Windows nightly portable workflow checks out FidelityFX SDK v1.1.4 at a pinned revision before configuration, builds the optional plugin, and stages it in the application-owned AMD scene-upscaler directory.

The SDK implementation remains isolated inside the optional plugin. No separate FidelityFX runtime is placed beside the main executable.

The FidelityFX SDK license is staged with the plugin and installed under the AMD FidelityFX third-party license directory. The SDK and the LichtFeld-owned AMD plugin use MIT-compatible licensing, while LichtFeld’s main application remains under its existing license.

## Documentation

The scene-reconstruction and build documentation now describes:

- the optional AMD plugin architecture;
- bootstrap versus runtime activation;
- application-owned loading paths;
- FidelityFX SDK acquisition and local opt-in configuration;
- depth, motion, jitter, phase-count, and history contracts;
- source-generation reuse;
- backend and preset persistence;
- supported viewport and split-view modes;
- Native fallback and explicit retry behavior;
- portable packaging and third-party license staging.


MrNeRF:
So this is the last pr from https://github.com/MrNeRF/LichtFeld-Studio/pull/1503 ? Can we close the superseded one?

francescofugazzi:
> So this is the last pr from #1503 ? Can we close the superseded one?

not yet, i should apply also as a choice for exporting videos using the upscalers from sequencer. But if you prefer we can close everything and I will do that in another moment

MrNeRF:
No, all good. Just asking.

MrNeRF:
I will make sure that it also works with linux and with respect to the licensing questions I will work on this next week.

PR #1940, opened 2026-09-01
Saving projects is 2.2x faster

MrNeRF:
Saving a 1.4GB project took 8.1 seconds. It now takes 3.7.

The save pipeline was reading and checksum-verifying the entire file three separate times, and flushing it to disk twice. Now it verifies the source while copying it, runs one full checksum validation of the staged file before the atomic rename (in parallel across blocks), and syncs once.

No safety was removed: the file is still fully checksum-verified and durably flushed before it replaces the old one, and a failed validation still aborts the save. The written payload is byte-identical to before. The full save path also now has permanent timing logs, so future regressions are visible instead of a black box.

Verified: 3 timed runs each way, saved file reopens with all 4M splats, format and project test suites pass.

### DIFF
--- a/src/io/include/io/project_container.hpp
+++ b/src/io/include/io/project_container.hpp
@@ -586,8 +586,13 @@ namespace lfs::io::project {
         [[nodiscard]] static lfs::Result<void>
         compact(const std::filesystem::path& path, const CompactionOptions& options = {});
         // Compacts source_path into a new destination without first cloning
-        // the complete source file. The source is only read; publication of
-        // destination remains transactional and fully validated.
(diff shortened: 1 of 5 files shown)


MrNeRF:
Review items addressed: compact_to now holds writer locks on both source and destination for the entire compaction (deterministic lock order; a single lock when the paths coincide), with Save As falling back to a private copy if another process holds the source lock; the compact_to doc comment now spells out that private_staging defers full validation and durability to the caller's final commit. Format suite 105/106 (standing skip), project suites 94/95 (standing skip), error-debt census clean.

PR #1941, opened 2026-09-01
SOG export is 1.9x faster and no longer drops colors of transparent splats

MrNeRF:
Exporting 3.3M splats to .sog took 18.7 seconds. It now takes 10.

What changed:
- The GPU color-compression step (the biggest cost) now runs concurrently with all the CPU packing and image encoding instead of before it.
- The seven WebP images encode in parallel, each starting as soon as its data is ready.
- The zip no longer re-compresses the already-compressed WebP images.
- The WebP encoder uses a faster lossless mode.

Bonus correctness fix found on the way: the old encoder call silently discards color data for fully transparent splats, so they came back with a wrong color after export+import. The new encoding preserves them exactly — measured reconstruction error is now equal or lower on every attribute.

The five deterministic images are pixel-identical to before (aside from the transparent-splat fix). The SH images differ between runs because the GPU clustering was already nondeterministic; quality is verified by reconstruction error against the source model. SOG tests pass. Detailed timing logs are now built in, so this path is no longer a black box.

Note: this is 1.9x, slightly under the 2x bar from the overnight brief — shipped on request. The remaining wall is the GPU k-means itself (8.2 of the 10 seconds); a measured kernel-tiling pass is the next lever if we want more.

### DIFF
--- a/src/io/cuda/kmeans.cu
+++ b/src/io/cuda/kmeans.cu
@@ -10,6 +10,7 @@
 #include <cmath>
 #include <cuda_runtime.h>
 #include <numeric>
+#include <optional>
 #include <random>
 #include <thrust/device_ptr.h>
 #include <thrust/sequence.h>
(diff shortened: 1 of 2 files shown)


MrNeRF:
Addressed the review items and the error-debt gate: CUDA event timers are now created only when debug logging is enabled, with every cudaEvent* status handled and any timer failure demoted to a debug log (timing disabled for that run, export unaffected); WebP futures are joined in archive order and appended as each finishes (single pass); zip compression is per entry — stored for .webp, deflated for meta.json; per-image timing strings are built only when debug logging is on. Re-measured after the changes: 9.5s for the 3.3M-splat scene (was 18.7s).

PR #1944, opened 2026-09-01
Launching after a crash no longer takes seconds

MrNeRF:
If the app ever crashed with a big model open, every later launch first read and checksummed the leftover recovery data in full — twice — before showing a window: 3.4 seconds instead of 0.8 on this machine, on every start until the recovery prompt was answered.

Now startup only reads the recovery headers, on a background thread after the first frame. The full check runs when you actually click Recover. The prompt and its behaviour are unchanged.

Measured: launch with a 1.2GB leftover recovery project 3.4s -> 0.83s (same as a clean launch); clean launch unchanged. Recovery, autosave, project and MCP suites pass.

### DIFF
--- a/src/io/include/io/project_recovery.hpp
+++ b/src/io/include/io/project_recovery.hpp
@@ -40,6 +40,13 @@ namespace lfs::io::project {
         std::vector<std::string> diagnostics;
     };
 
+    struct RecoveryInspectionOptions {
+        // Detection only needs the container index and small metadata chapters.
+        // Payload CRC/decode validation remains available for callers that need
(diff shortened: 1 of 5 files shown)


MrNeRF:
Addressed all three review items: the candidate timer's destructor is noexcept with guarded logging; the recovery worker now takes the jthread stop_token and checks it between directories/candidates so teardown's request_stop() is honored; a failed background scan leaves the candidate optional empty (warn-logged) so the synchronous fallback still offers recovery. Recovery/autosave/project suites 127/128 (1 standing skip), error-debt census clean.

PR #1947, opened 2026-09-01
Add a seperate .licht inspect and restore CLI Tool

dreamTent:
- Add `tools/inspect_licht.py` to inspect a `.licht` projects:
  -  generations, training iteration, Gaussian count (including ByteShuffle older checkpoints), and which generation tombstoned a previous chapter.
- Restore an older still-on-disk generation or training iteration to a new file with `--output` and `--generation` / `--iteration`.


MrNeRF:
Thank you!

PR #1951, opened 2026-09-01
Add an FOV slider with reset to the HTML viewer

jacobvanbeets:
Adds a camera field-of-view slider to the HTML export viewer's settings panel, plus a button to reset the FOV back to the exported default.

## Changes
- `viewer_template.html`: FOV slider row (10-120 deg, step 0.5) with a value readout and a "Reset FOV" button in the settings panel
- `index.js`:
  - `state.fov` / `state.defaultFov` initialized from `settings.cameras[0].initial.fov` (the exported FOV)
  - `applyCamera` applies the user-adjusted FOV every frame (animation tracks keep driving their own FOV); the orthographic frustum size now derives from the effective FOV so ortho mode stays in sync with the slider
  - slider + reset wiring and tooltips in `initUI`
- `index.css`: slider styling with the existing orange accent

No C++ changes needed: the exporter reads the viewer files from disk at export time, so the next export picks this up automatically.

## Verification
- `node --check` on the bundle passes
- `index.js` diff is line-ending clean (the file's original CRLF lines are preserved, so the diff only contains the feature lines)
- Verified in a browser with a fresh HTML export: the slider changes the FOV live in orbit, fly, and ortho modes, and "Reset FOV" restores the default

### DIFF
--- a/src/visualizer/gui/resources/viewer/index.css
+++ b/src/visualizer/gui/resources/viewer/index.css
@@ -330,6 +330,24 @@ button {
   margin: 8px 0;
   background-color: #666;
 }
+#settingsPanel > .settingsRow > .fovLabel {
+  cursor: default;
+  white-space: nowrap;
+}
+#settingsPanel > .settingsRow > #fovSlider {
+  flex-grow: 1;
+  margin: 8px 4px;
+  accent-color: #F60;
+}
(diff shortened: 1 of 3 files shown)


MrNeRF (approved):

MrNeRF:
Thx!

PR #1956, opened 2026-09-02
 Introduce v2 theme families and adaptive viewport chrome

francescofugazzi:
## Summary

This PR evolves the theme system from independent flat presets into versioned theme families while preserving compatibility with existing themes and preferences.

It introduces:

- v2 theme-family manifests with semantic dark and light variants;
- system-aware automatic variant selection;
- family and variant selection in Preferences and View;
- optional gradients with palette-derived fallbacks;
- the new high-legibility Signal Day and Signal Night family;
- broader theme coverage across the RmlUI application chrome;
- configurable Solid, Translucent, and Frosted viewport controls;
- configurable viewport utility-toolbar placement;
- multiple sidebar, split-view, input-routing, hover, layout, and contrast fixes;
- theme-aware Python Console chrome without changing its Output and Terminal rendering.

## Theme family format v2

The v2 format groups related variants in one JSON family file.

It supports:

- semantic `dark` and `light` variant categories;
- short user-visible names such as `Night`, `Day`, `Mocha`, and `Latte`;
- stable family and variant IDs;
- family and variant ordering and localization keys;
- shared properties inherited by family variants;
- explicit fallback presets;
- single-variant families;
- optional fonts when already supplied by a theme;
- optional gradients;
- JSON parsing, serialization, validation, hashing, and hot reload.

Bundled themes were migrated to this format:

- LichtFeld Dark and Light now form one LichtFeld family;
- Catppuccin Mocha and Latte now form one Catppuccin family;
- Gruvbox and Nord remain valid single-variant families;
- Signal adds dedicated Night and Day variants.

Existing v1 manifests and standalone theme JSON files remain supported.

## Theme selection and automatic mode

Preferences now shows one family entry rather than separate dark and light themes.

When a family contains multiple variants:

- the available variant names are displayed using their custom short labels;
- variants remain semantically categorized as dark or light;
- Auto follows the operating-system application color preference;
- the selector buttons share the available width equally;
- Auto is hidden when system detection is unavailable.

System detection is implemented through:

- the Windows application color preference;
- SDL3 system-theme detection on Linux when the desktop session exposes it.

Unknown or unsupported environments fall back safely to dark without exposing a non-functional Auto option.

The selected family and explicit or automatic appearance mode are persisted. Legacy preset IDs remain loadable, and the legacy `set_theme` path may still persist a plain preset ID.

`View > Theme` now uses nested family and variant submenus. The View menu also displays the active family and appearance mode directly, so the current choice remains visible without opening every submenu.

## Optional gradients

Themes may optionally define gradients for:

- window bodies;
- panel bodies;
- window title bars;
- section headers;
- hovered section headers;
- progress and selected states;
- scrubber tracks;
- scrubber fills;
- histogram headers;
- histogram fills;
- histogram selections.

Gradient support includes:

- per-property validation and fallback;
- serialization and cache hashing;
- hot-reload propagation;
- palette-derived defaults for flat and legacy themes;
- RmlUI token generation for the current consumers.

Gradients are never mandatory. Themes without them retain a coherent flat appearance.

The training progress-bar easter egg remains intentionally unchanged.

## Signal theme family

Signal Night and Signal Day provide a distinct blue-violet visual language rather than a recolored LichtFeld theme.

Signal includes:

- stronger separation between application regions;
- layered window, panel, header, and selected-state gradients;
- dedicated Sequencer and Asset Manager surfaces;
- coordinated menu, status-bar, right-dock, and viewport chrome;
- readable light and dark variants;
- contrast-aware text and icon treatment;
- short `Night` and `Day` selector labels.

Theme-specific styling remains constrained to theme-driven surfaces. Semantic application controls retain their established identity unless explicitly designed otherwise.

## Application-wide theme coverage

Theme and gradient tokens now cover:

- floating windows and title bars;
- Preferences sections and controls;
- application menu chrome;
- status bar;
- hosted and docked panel chrome;
- right-sidebar tabs and content surfaces;
- Sequencer modules, transport controls, and scrubbers;
- Asset Manager surfaces, cards, rows, controls, icons, and scrollbars;
- histogram headers, comparison bars, and history controls;
- progress and selected-control states;
- viewport toolbars and contextual tool strips;
- orientation-gizmo actions;
- Python Console toolbar, tabs, splitter, and Packages panel.

Flat palette-derived fallbacks remain available for themes without enhanced properties.

This PR improves the surrounding Scene Graph frame and right-dock integration, but does not claim a complete conversion of every legacy internal Scene Graph color.

## Asset Manager fixes

- Separate layout rules from palette-specific styling.
- Remove remaining fixed dark surfaces from the themed chrome.
- Theme cards, rows, controls, borders, warnings, selections, icons, resize handles, and scrollbars.
- Replace the fixed-color add-folder SVG with a theme-colored glyph.
- Improve light-theme icon and control visibility.
- Improve surface layering in Signal Day and other light variants.

## Sequencer fixes

- Route the Quality scrubber through the shared theme scrubber tokens.
- Prevent local fixed gradients from masking the active theme.
- Theme the principal Sequencer modules and transport row.
- Keep Quality, Speed, FPS, and Resolution controls visible at narrow sidebar widths.
- Improve control and label readability in light and dark variants.

## Right sidebar and Scene Graph fixes

- Restore stable Scene, History, and Logging tab geometry.
- Keep Rendering, Training, and plugin tabs distinguishable.
- Update the active-tab indicator immediately when a tab is clicked.
- Remove stale transition state from tab selection.
- Re-run hover hit testing after layout changes.
- Avoid mismatched hover coordinates and disappearing hover states.
- Keep the right dock visually continuous while preserving internal separation.
- Remove inconsistent heavy content borders.
- Frame the Scene Graph consistently on all edges.
- Preserve readable contrast across flat and gradient themes.

## Viewport controls and icon fixes

- Make menu, toolbar, and gizmo icons contrast-aware.
- Provide readable normal, hovered, selected, and disabled states.
- Improve disabled toolbar icons in light themes.
- Improve the Home and visibility controls below the orientation gizmo.
- Make those controls compact and partially transparent while retaining contrast.
- Apply the same adaptive treatment to selection, transform, mirror, crop, and other contextual toolbars.
- Make toolbars size to their actual controls.
- Restore the selection panel’s window-relative sizing from #1860 so its depth controls can grow with the viewport.
- Align duplicated split-view controls with their corresponding viewport.
- Keep translucent and frosted effects constrained to the actual toolbar geometry.

## Configurable viewport chrome

Appearance preferences now provide:

- Solid;
- Translucent;
- Frosted.

The selected mode affects visible viewport toolbars, contextual tool panels, and orientation-gizmo actions.

The implementation:

- derives tint, borders, and icon contrast from the active theme;
- preserves each theme’s visual identity;
- composites a cached, low-resolution Vulkan backdrop beneath eligible controls;
- avoids a full-resolution blur path;
- feeds the actual renderer capability back into the theme tokens;
- resolves unsupported Frosted requests to the Translucent presentation;
- preserves the user’s saved Frosted preference if runtime support is temporarily unavailable;
- requires no additional properties from custom themes.

The preference is persisted and exposed through the native preferences model, Python UI bindings, Preferences panel, and bundled localizations.

## Viewport utility-toolbar placement

The vertical utility toolbar supports three persisted modes:

- **Top** — placed near the upper viewport edge with intentional spacing from the application chrome;
- **Centered** — vertically centered and used as the default;
- **Free** — repositioned vertically through a dedicated double-line drag handle.

Free placement:

- stores a normalized position;
- remains stable across viewport resizing and UI-scale changes;
- is clamped to the available viewport bounds;
- preserves pointer capture until dragging ends;
- keeps duplicated split-view toolbars synchronized with their own viewport bounds.

## Independent split-view fixes

- Prevent `Shift+V` from activating independent split view when the scene is empty.
- Apply the same guard to the title-bar split-view action.
- Avoid creating an empty secondary camera, gizmo, axes, divider, or toolbar.
- Preserve the ability to close an already active split.
- Hide orphan secondary controls if the editor context becomes empty.
- Preserve normal scene-clear behavior, returning the renderer to one viewport.

## Retained-menu input routing

Nested theme menus made an existing input-routing problem easier to reproduce: selecting a menu item could also activate a Python Console or panel control underneath it.

The central input path now:

- preserves menu pointer ownership when an action closes its dropdown during mouse-down;
- retains that capture through the matching mouse release;
- masks pointer input before dispatching the frame to underlay panels;
- preserves keyboard and text-input delivery;
- also protects outside-close clicks from leaking into underlying controls.

This fix applies consistently to retained application menus rather than special-casing the Python Console.

## Python Console theme integration

The surrounding Python Console chrome now follows the active theme:

- ordinary toolbar controls use themed surfaces, borders, text, hover, and disabled states;
- bottom tabs are compact, centered, and clearly selected;
- the splitter uses a subtle accent edge rather than a heavy full-width blue hover bar;
- the Packages panel and search field follow the active palette.

The established semantic controls remain unchanged across theme families:

- Run retains its original green identity;
- Stop retains its original red identity;
- active toolbar controls retain their original blue identity.

The Output and Terminal rendering surfaces remain explicitly outside theme overrides and preserve their existing terminal presentation.

## Review follow-up

The review findings are addressed by:

- refreshing the generated Python UI stubs;
- restoring the selection-panel sizing introduced by #1860;
- marshalling `set_theme`, `set_theme_family`, and `set_viewport_chrome_style` onto the viewer thread;
- releasing the Python GIL while synchronous UI-thread work completes;
- propagating actual Vulkan frosted-glass availability into theme presentation;
- logging caught theme-loading exceptions;
- avoiding additions to the Error Debt baseline;
- removing unused presentation tokens;
- clarifying legacy theme-preference persistence;
- narrowing the Scene Graph coverage claim.

## Documentation

Documentation now covers:

- theme families and semantic appearance modes;
- the v2 schema and every runtime-supported property;
- v1 and standalone-theme compatibility;
- custom visible variant names;
- Auto detection and fallback behavior;
- optional gradients and their current consumers;
- Solid, Translucent, and Frosted viewport modes;
- toolbar placement modes and persisted Free positioning;
- RmlUI viewport and split-toolbar geometry;
- preference persistence;
- current custom-theme development and installation workflow;
- missing-family and missing-variant recovery.

A complete `theme-family-v2.json` example is included and contains all currently supported runtime properties with human-readable descriptions.

Managed end-user theme import, deletion, and a dedicated user-theme directory remain future work.

## Compatibility and fallback behavior

- Existing v1 and standalone themes remain supported.
- Legacy IDs and saved preferences remain loadable.
- Missing optional properties inherit from the selected dark or light fallback.
- Missing gradients use palette-derived defaults.
- Invalid individual gradients do not invalidate the rest of the theme.
- Invalid or missing families are skipped safely.
- Missing saved families fall back to LichtFeld Dark.
- Single-variant families work without Auto or unnecessary variant buttons.
- Unknown system-theme detection falls back to dark.
- Unsupported frosted rendering falls back visually to Translucent.
- Output and Terminal rendering remain unchanged.

## Tests and validation

Coverage includes:

- v2 family parsing and validation;
- legacy theme compatibility;
- missing and malformed fallback paths;
- family and variant metadata;
- custom visible variant names;
- Auto capability detection;
- preference persistence;
- nested View-menu schemas and hover behavior;
- equal-width appearance buttons;
- optional-gradient parsing and consumer coverage;
- Asset Manager theme isolation;
- Sequencer scrubbers and responsive controls;
- right-sidebar tab behavior;
- viewport toolbar geometry and contrast tokens;
- viewport chrome preference bindings;
- actual frosted-capability fallback;
- Python UI-thread marshalling and stub completeness;
- empty-scene split prevention;
- toolbar placement and drag persistence;
- retained-menu pointer capture;
- Python Console theme boundaries;
- documentation-template property coverage;
- safe-mode preference behavior.

Validation performed during development includes:

- focused Python menu, theme, Preferences, toolbar, sidebar, and console tests;
- native theme and menu tests;
- `git diff --check`;
- `git-clang-format` checks on changed C++ hunks;
- Windows build validation;
- iterative Windows runtime testing across light and dark themes, sidebar tabs, responsive panels, viewport controls, toolbar placement, split view, frosted chrome, retained menus, and Python Console chrome.

Recent focused results:

- 84 tests passed for viewport toolbar placement;
- 88 tests passed and 1 skipped for menu input routing;
- 48 tests passed for the final Python Console integration.

A broad isolated Windows Python run completed with 1498 passed and 54 skipped. Its remaining 24 failures are outside the modified console/theme paths and include Windows-specific symlink, path, newline, encoding, and existing source-contract cases.


src/visualizer/gui/rmlui/resources/viewport_overlay.rcss
```diff
@@ -630,10 +627,8 @@ body {
 
 #selection-block .viewport-selection-panel {
     min-width: 0;
-    /* Window-relative: the panel tracks the viewport width and the sliders
-       grow into it (capped below), instead of collapsing to the icon row. */
-    width: 65%;
-    max-width: 1500dp;
+    width: auto;
```
MrNeRF: This undoes #1860 (Alex, 2026-08-29): the panel was made window-relative on purpose so the depth sliders grow with the viewport, see the comment you removed above. If content-sized is what you want, make the case in the PR; otherwise keep the 65% / 1500dp cap.

src/visualizer/gui/rml_viewport_overlay.cpp
```diff
@@ -1041,6 +1046,72 @@ namespace lfs::vis::gui {
...
+                append_region(button, 5.0f);
+        }
+
+        Rml::ElementList panels;
+        document_->GetElementsByClassName(panels, "viewport-transform-panel");
+        for (Rml::Element* panel : panels)
+            append_region(panel, 8.0f);
+
+        if (!regions.empty())
+            renderer->RenderFrostedGlass({regions.data(), regions.size()});
```
MrNeRF: Return value dropped. When this returns false the RCSS still uses the frosted opacities (0.44 / 0.50 in rml_theme.cpp), so the fallback is thinner chrome with no blur behind it. Track the result and switch the tokens to the translucent values, or resolve the preference to translucent once the renderer says no.

src/python/lfs/py_ui.cpp
```diff
@@ -4903,6 +4903,30 @@ namespace lfs::python {
             []() -> std::string { return vis::currentThemeId(); },
             "Get current stable theme id");
 
+        m.def(
+            "set_theme_family",
+            [](const std::string& family_id, const std::string& mode) {
+                return vis::setThemeFamilySelection(family_id, mode);
```
MrNeRF: This activates the RmlUi theme on whatever thread calls it. From the Python console that is the script worker, and it crashes (segfault in Rml::StyleSheetNode::BuildIndex under Rml::Context::ActivateTheme). Same for set_viewport_chrome_style in py_ui_theme.cpp. Post both to the UI thread. set_theme above has the same problem, feel free to fix it in the same go.

src/visualizer/theme/theme.cpp
```diff
@@ -878,187 +1242,316 @@ namespace lfs::vis {
...
+            std::ifstream file;
+            if (!lfs::core::open_file_for_read(lfs::core::utf8_to_path(path), file))
+                return false;
 
-            return true;
+            json j;
+            file >> j;
+            t.name = j.value("name", "Custom");
+            return applyThemeJson(t, j, true, false);
         } catch (...) {
```
MrNeRF: New catch-all that swallows the error and then gets added to error_debt_baseline.json so the census passes. Catch std::exception, log what failed, and leave the baseline alone. Same at line 1494.

MrNeRF (changes requested):
Thanks Francesco, this is a big one and most of it is good. I built the branch on Linux (dev still pulls OpenUSD and OpenImageIO from vcpkg, so that took a while), ran the theme and menu gtests (20 green) plus the 87 Python tests from the PR, and drove it live on a 4090 with a 5M splat scene: all eight variants, the three viewport-control styles, the Preferences buttons and the View menu. Auto works here on Linux through the portal, the Night/Day naming reads well, the light themes are finally usable, and the frosted blur renders as advertised. Signal is a nice addition.

Four things before this can go in:

1. **It does not build on Linux.** The committed Python stub is stale: `get_viewport_chrome_style` and `set_viewport_chrome_style` are missing from `src/python/stubs/lichtfeld/ui/__init__.pyi`, so the default target fails at `python_stub_check.stamp`. Run `cmake --build build --target refresh_python_stubs` and commit the result.

2. **The selection panel change reverts #1860.** Alex made that panel window-relative on purpose four days ago so the depth sliders grow with the viewport, and the comment you removed says why. This PR silently turns it back into a content-sized box. If you think that was wrong, say so here and we decide, but it should not ride along unannounced. Inline.

3. **Frosted glass falls back only halfway.** The RCSS takes the thinner frosted opacities from the preference string alone, while the overlay ignores what `RenderFrostedGlass` returns. On a swapchain without transfer-source usage the toolbars get more see-through and no blur, the opposite of what the docs promise. Feed the capability back into the token set so unsupported hardware really lands on the translucent look. Inline.

4. **The new Python bindings crash when called from a script thread.** `set_theme_family` and `set_viewport_chrome_style` run the RmlUi theme switch on the calling thread. From the Python console (and MCP `editor_run`) that is a worker thread racing the renderer; I got a double free once and then a segfault in `Rml::StyleSheetNode::BuildIndex` under `Rml::Context::ActivateTheme` <- `setThemeFamilySelection` <- `PythonConsoleState::runScriptAsync`. The GUI paths are on the right thread, so clicking is fine, and the old `set_theme` has the same weakness, but let's not add two more. Marshal these to the UI thread. Inline.

Smaller:
- Don't grow `tools/error_debt_baseline.json` with two new `catch (...) { return false; }` sites. Log the exception instead. Inline.
- "Scene Graph" is listed under the themed surfaces, but `scene_tree.rcss` still has about 120 hard-coded colors. Either theme it or drop the claim.
- `preferences-and-user-storage.md` says the theme preference is always `family:mode`; the legacy `set_theme` path (used by "Reset this section") still writes the plain preset id. Both load fine, the sentence just needs softening.
- Five tokens are emitted and never used: `layered_shadow.1`, `panel.body_bg`, `panel.body_bg_or_transparent`, `right_panel.tab_active_bg`, `right_panel.separator`.

Next time please split work like this: the family format and selection UI, the gradients and Signal, and the viewport chrome with the toolbar fixes are three PRs. 5,400 lines in one is hard to review properly and impossible to bisect.

francescofugazzi:
Pushed four follow-up commits addressing the review and the subsequent UI validation:

- `6d7cf4568` — address the requested manifest v2 review changes
- `235835903` — add configurable viewport utility-toolbar placement
- `76ef1fb48` — prevent menu clicks from reaching underlay panels
- `c212821c5` — integrate the Python Console chrome with themes

All blocking review points are covered:

- refreshed the generated Python UI stubs;
- restored the window-relative selection-panel sizing introduced by #1860;
- marshalled `set_theme`, `set_theme_family`, and `set_viewport_chrome_style` onto the viewer thread while releasing the Python GIL during the synchronous wait;
- propagated the actual Vulkan frosted-glass capability into theme presentation, so unsupported configurations resolve to the translucent appearance without changing the saved preference;
- replaced the new catch-all exception handlers with logged `std::exception` handling;
- restored the Error Debt baseline to the upstream counts;
- removed the unused presentation tokens;
- clarified the legacy plain-preset preference path in the documentation;
- narrowed the Scene Graph wording so the PR does not claim a complete conversion of its legacy internal colors.

The follow-up also adds:

- Top, Centered, and persisted Free positioning for the viewport utility toolbar;
- a dedicated drag handle in Free mode, with clamping and synchronized split-view placement;
- central pointer capture for retained menus, preventing a menu selection from also activating a control underneath it;
- theme-aware Python Console toolbar, tabs, splitter, and Packages chrome;
- preservation of the original green Run, red Stop, and blue active-button identity;
- an explicit boundary keeping the Output and Terminal renderers unchanged.

PR #1957, opened 2026-09-02
Fix Windows link failures for lfs_mcp and lfs_visualizer

jacobvanbeets:
Windows Release builds of `lfs_mcp.dll` and `lfs_visualizer.dll` failed at link time with:

- `LNK2005`: `log_arena_failure_vram_snapshot` defined in both `lfs_core.dll` and `lfs_core_cuda.lib`
- `LNK2019` x6: unresolved `VmmDeviceBuffer` inline members (default ctor, `data()`, `committed_bytes()`, `granularity_bytes()`, `operator bool()`, `kGranularityBytes`)

Both are MSVC `dllimport`/`dllexport` boundary issues (Linux builds pass because visibility attributes don't create import thunks).

## Root causes & fix

**1. `log_arena_failure_vram_snapshot` (LNK2005).** Declared `LFS_CORE_API` in `memory_arena.hpp` — i.e. every consumer compiles it as `__declspec(dllimport)` from `lfs_core.dll` — but defined in `memory_arena.cu`, which is part of the static `lfs_core_cuda`. When `memory_arena.cu.obj` was pulled into a consumer link, the local definition collided with the DLL import. The definition now lives in `memory_arena_owner.cpp` (built into `lfs_core`), next to its siblings `global_arena_manager()` / `shutdown_global_arena_manager()`, so it is genuinely exported from `lfs_core.dll` as declared.

**2. `VmmDeviceBuffer` (LNK2019/LNK2001/LNK4217).** The class is declared `LFS_CORE_API`, but `vmm_device_buffer.cpp` was compiled into `lfs_core_cuda` without `LFS_CORE_EXPORTS`. That TU saw the class as `dllimport`, so MSVC emitted `__imp_` references instead of definitions for the header-inline members — and since no `LFS_CORE_EXPORTS` TU ever defined them, `lfs_core.dll` never exported them and the import thunks didn't exist. `vmm_device_buffer.cpp` now compiles into `lfs_core` (with `LFS_CORE_EXPORTS`), so all members — inline, defaulted ctor, `kGranularityBytes`, and the out-of-line ones — are exported from `lfs_core.dll` and imported by consumers (`gsplat_backend_lfs`, `lfs_training`, …). This mirrors the existing `OperationId`/`SmallFields` pattern in `error.hpp`.

## Verification

Full Windows Release build (ninja, same toolchain as CI): `lfs_core.dll`, `lfs_mcp.dll`, and `lfs_visualizer.dll` now link cleanly; no LNK2005/LNK2019/LNK4217 remaining.

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -71,6 +71,7 @@ set(CORE_SOURCES
         host_metrics.cpp
         failure_report.cpp
         cuda/memory_arena_owner.cpp
+        cuda/vmm_device_buffer.cpp
         memory_pressure.cpp
         memory_domain.cpp
         operator_callbacks.cpp
--- a/src/core/cuda/CMakeLists.txt
+++ b/src/core/cuda/CMakeLists.txt
@@ -10,7 +10,6 @@
 set(LFS_CORE_CUDA_SOURCES
     memory_arena.cu
     exportable_storage.cpp
-    vmm_device_buffer.cpp
     lanczos_resize/lanczos_resize.cu
     undistort/undistort.cu
     kernels/selection_ops.cu
--- a/src/core/cuda/memory_arena.cu
+++ b/src/core/cuda/memory_arena.cu
@@ -18,44 +18,6 @@
 
 namespace lfs::core {
 
-    void log_arena_failure_vram_snapshot(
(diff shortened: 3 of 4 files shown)


MrNeRF:
Thank you @jacobvanbeets

PR #1959, opened 2026-09-02
Fix stack overflow when embedding dataset into project (on Windows)

thomas-koch-core11:
Move the buffers to heap-allocated std::vector<char> instead.


MrNeRF:
Thx for the fix!

PR #1967, opened 2026-09-02
Refresh the error-debt baseline after the logger timer change

MrNeRF:
The error-debt gate fails on master since #1966: the logger's guarded timer scopes added three empty catch blocks in src/core/logger.cpp without refreshing the baseline. This regenerates tools/error_debt_baseline.json (no code change). Every open PR's gate goes green once this is in.


MrNeRF:
Superseded: the baseline refresh landed with #1965, and the census is clean on master now.

PR #1977, opened 2026-09-03
Add native LPIPS evaluation metric

MrNeRF:
Adds native LPIPS v0.1 with a VGG16 backbone to held-out image evaluation, alongside PSNR and SSIM. Results appear in CSV/text reports, per-image CSV, the status bar, notifications, the Python store and benchmark summaries. Missing LPIPS remains unavailable; a valid zero remains a real score.

Inference uses LichtFeld's CUDA tensor engine: fp16 for normal evaluation and fp32 for reference checks. Large images use overlapping tiles and a check of available GPU memory. This reduces memory pressure but does not guarantee that evaluation cannot run out of memory. Evaluation uses the same rounded 8-bit values as saved PNGs, and resuming at the final iteration now performs the requested evaluation.

Weights are [LPIPS VGG16 v0.1 in native LFW format](https://github.com/MrNeRF/LichtFeld-Studio/releases/download/model-lpips-v1/lpips-vgg16-v0.1.lfw), approximately 59 MB. They are downloaded on demand, checked against a pinned SHA-256 and cached in `~/.lichtfeld/onnx/lpips-vgg16-v0.1.lfw`. `LFS_LPIPS_WEIGHTS` selects a local override; `--no-download` permits cached weights without fetching them. GUI preparation runs on the initialization worker after pending settings are applied. The exporter requires PyTorch/torchvision/LPIPS; the app does not require them at runtime.

The branch includes current master and resolves the application startup conflict. Review fixes correct rectangular tile allocation and memory estimates, prevent checksum-rejected default files from being reopened, join CUDA streams for packed convolution weights, preserve the download policy through checkpoint loading, validate model/input shapes and fix missing-weight test skips. Cleanup consolidates duplicate fast inference paths, removes private benchmark/fixture assumptions and separates timing from correctness tests.

Validation on Linux, RTX 4080, CUDA 13.3, Release:

- Built the app, test executable and separate NN benchmark target.
- LPIPS/reference, two-axis tiling, input preservation, metrics, NN, tensor-view and full MoGe parity/footprint checks passed.
- The broader C++ run passed 4,349 tests and skipped 54 in 78 seconds. Three tests requiring absent `tests/data/bike.ply` were explicitly excluded. The default 60-second CTest limit was too short; the other five fast CTest groups passed.
- CUDA memcheck passed three selected tests with zero errors. Racecheck of complete 16×16 fast inference reported zero hazards; a larger, slow racecheck was interrupted without a result.
- Bicycle training evaluated 25 images; final-iteration resume reproduced PSNR/SSIM/LPIPS. Missing weights left LPIPS blank while evaluation completed.
- The committed crop's fast/full-precision LPIPS difference was about 0.000025. A 512×6000 memory probe fell from 1,664 MiB retained to 512 MiB, within the revised estimate. These are measured cases, not universal accuracy or memory guarantees.

Windows runtime/build validation is left to CI; PowerShell summary changes were not executed locally. The input convention matches the original 3DGS evaluator (`Identity` on [0,1] inputs). Masked LPIPS evaluates images with excluded pixels zero-filled; it is not a mean over selected pixels only.


MrNeRF:
Merged master (training-manager start path) and addressed the review: LPIPS is now an optional value end to end so 0.0 is kept and unavailable stays blank, the dispatch caches are race-free, the weights preparer honours its argument, and saved PNGs round to 8 bits instead of truncating (the quantiser produced values one ULP below k/255 under fast-math division, which cost one level per pixel in every saved eval image). Unrelated formatter churn and benchmark defaults reverted; the NN bench builds only on demand. LPIPS, metrics, image-io and MoGe test groups pass locally on an RTX 4080.

PR #1986, opened 2026-09-03
Add owner-scoped plugin localization catalogs

francescofugazzi:
## Summary

Add optional, owner-scoped localization catalogs to the existing Python plugin system.

Plugins can now ship runtime translations without modifying LichtFeld's core locale files or relying on global `loc_set()` overrides.

## Catalog contract

- Catalogs live in an optional `locales/` directory.
- `locales/en.json` is required only when localization catalogs are present.
- Additional catalogs use lowercase filenames for available LichtFeld languages, such as `it.json` or `de.json`.
- Authors declare relative, optionally nested keys.
- The host derives a canonical owner ID from `project.name` and resolves keys under `plugins.<owner-id>.*`.
- Lookup preserves legacy overrides, then uses the active plugin language, plugin English, and finally the unresolved key.
- Non-English catalogs may omit English keys but cannot introduce additional keys.
- Format placeholders, conversions, and format specifications must match the English catalog.

## Lifecycle and hot reload

Catalogs are validated and registered before the plugin module is imported, allowing translated strings during import and `on_load()`.

Ownership tokens remove registrations during normal unload, hot reload, failed import or activation, and cancelled loading.

Changes to `locales/*.json` participate in the existing plugin hot-reload watcher.

Localization changes publish the existing thread-safe language generation signal. UI refresh remains frame-driven, avoiding direct `GuiManager` mutation from plugin-loading worker threads.

## Local validation

The existing command validates catalogs without importing plugin code:

`LichtFeld-Studio plugin check <name>`

Validation covers strict UTF-8, malformed JSON, duplicate or colliding keys, owner and language identifiers, English fallback availability, placeholder parity, and bounded file, entry, key, value, and nesting limits.

## Compatibility

- The Python plugin API remains `1.0`.
- Existing localization override APIs remain unchanged.
- Plugins without `locales/` follow the existing load path.
- Existing plugins remain installable.
- No changes are made to DLSS, FSR, CUDA, rendering, Sequencer, trainer behavior, native plugin loading, or distribution.

## Validation

User-verified on Windows:

- project builds successfully;
- tests pass;
- LichtFeld Studio starts successfully;
- existing plugins can still be installed.

Additional checks:

- 56 relevant Python tests passed;
- changed C++ hunks pass `git-clang-format`;
- `git diff --check` passes.


MrNeRF (changes requested):
Thanks Francesco, the feature itself works nicely. I built it merged onto today's dev on Linux, ran the plugin Python suites (316 green including your new file) and drove it live on a 4080 with a demo plugin shipping `locales/en.json` and `locales/de.json`: strings resolve, a missing key comes back as the key, switching the app to German switches the panel (with the English-only key falling back and one warning in the log), editing only `de.json` hot-reloads the plugin and the panel shows the new text, a catalog with a placeholder mismatch is refused with a clear message at load and `plugin check` reports the same, and unload puts the key back. Nice work on the lifecycle.

Two things before it can go in:

1. **It no longer merges onto dev, and the new C++ tests fail there.** `localization_manager.{hpp,cpp}` conflict with the `language_generation_` counter dev added in the meantime. After resolving that, all five `PluginLocalizationTest` cases fail at `tests/test_plugin_localization.cpp:32`: since #1965 (63074465b) `LocalizationManager::initialize()` requires a `locale_index.json` next to the locales directory, and the fixture only writes `en.json`/`it.json`, so `initialize()` returns false. Please rebase and add the index to the fixture (and to the `reset` test at line 141).

2. **Owner ids collide.** `canonical_plugin_id` (`plugin_localization.py:37`) maps `my_plugin`, `my.plugin`, `My-Plugin` and `my__plugin` all to `my-plugin`. Two such plugins either share one namespace silently, or, if they define the same key, the second one fails to load with "already registered" because of a plugin it has nothing to do with. Detect the collision at discovery time, or include something unique in the owner id.

Smaller:
- A lone surrogate in a catalog (`{"message": "\ud800"}`) passes `json.loads`, then `child.encode("utf-8")` at `plugin_localization.py:217` raises `UnicodeEncodeError`, which `validate_plugin_catalogs` (line 122) does not catch, so `plugin check` dies with a traceback instead of a validation error. Reproduced.
- `"{0} {}"` passes the placeholder check but raises at format time; NUL bytes inside values are accepted although the strings end up as `const char*` on the C++ side.
- `registerPluginCatalog`/`unregisterPluginCatalog` (`localization_manager.cpp:301` onward) never bump a generation; only the Python binding (`py_ui.cpp:3673`) publishes the visualizer signal. A C++ caller of the new API leaves the GUI stale. Cheapest is to bump the core counter dev now has and let the binding stay as is.
- `locales/` and the files in it are followed through symlinks (`plugin_localization.py:62`), so a catalog can live outside the plugin directory. Resolve and require containment, or say symlinks are fine on purpose.
- `zh-hans.json` is accepted but never selected because the host only knows `zh`; validating against the languages the host ships would catch typos.
- Doc nit: the docs say "currently shipped examples include `en`, `it`, and `de`" while nothing ships any; drop or rephrase.

One thing I could not pin down: once, after a few hot reloads with the plugin's tab open, `lf.plugins.unload` followed by `lf.plugins.load` from the Python console crashed in `PanelRegistry::register_panel`. Five retries on this branch and the same sequence on master were clean and the backtrace has no localization frames, so I am not counting it against the PR, just mentioning it in case you see it too.

PR #1999, opened 2026-09-03
Fix logger.cpp compilation on MSVC

thomas-koch-core11:
Similar fix as in mapped_file.cpp: The file now defines NOMINMAX and WIN32_LEAN_AND_MEAN before including <windows.h>.


thomas-koch-core11:
Already solved in 959e9c9

PR #2001, opened 2026-09-03
Let headless training start from an untrained .licht with embedded dataset via --data-path

thomas-koch-core11:
### Summary
Let headless training start from an untrained .licht with embedded dataset via `--data-path`.

Headless mode could not use either: `--resume` required a bound checkpoint and the embedded-dataset extraction lived only in the visualizer.

- Move the extraction into `lfs_io` (io/embedded_dataset.hpp) so headless and the GUI share it. The function body is unchanged.
- Accept an untrained .licht on `--data-path`. application.cpp resolves its dataset from the REFS folder or the embedded copy, applies the PRMS dataset options, and continues as an ordinary dataset run. Projects with a checkpoint are rejected with a hint to use `--resume`.
- Embed the images folder the project was loaded with instead of the first detected one, so downscaled datasets round-trip correctly.

### Future Work
- As mentioned in the second point above, if a project already contains checkpoints from a previous training run, an error is printed that `--resume` should be used. Instead, it could automatically just resume training.
- Starting training from .licht projects that don't have images embedded does not work yet.

### DIFF
--- a/src/app/application.cpp
+++ b/src/app/application.cpp
@@ -26,6 +26,7 @@
 #include "core/user_paths.hpp"
 #include "diagnostics/vram_profiler.hpp"
 #include "io/cache_image_loader.hpp"
+#include "io/embedded_dataset.hpp"
 #include "io/project_document.hpp"
 #include "io/project_recovery.hpp"
 #include "tcp/include/tcp_publisher.hpp"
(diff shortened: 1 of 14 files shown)


src/app/application.cpp
```diff
@@ -116,6 +117,113 @@ namespace lfs::app {
...
+            if (!dataset_root) {
+                return lfs::Status::failure(training_project_error(
+                    lfs::ErrorCode::NotFound,
+                    "The project's dataset is neither reachable nor embedded",
+                    LFS_SOURCE_SITE_CURRENT()));
+            }
+
+            // Adopt the stored dataset options; machine-specific paths and
+            // loading resources stay with the command line.
+            auto dataset = std::move(snapshot->dataset);
```
MrNeRF: Only the dataset block is taken from the project. The GUI saves the full pending parameter state, so strategy, iterations, max cap etc. set in the UI are in PRMS as well, but they get replaced by CLI defaults here. I would adopt the whole snapshot and then re-apply the explicit CLI overrides, same as --resume does.

src/app/application.cpp
```diff
@@ -116,6 +117,113 @@ namespace lfs::app {
...
+            dataset.data_path = std::move(*dataset_root);
+            dataset.images = std::move(images_folder);
+            dataset.output_path = params.dataset.output_path;
+            dataset.output_path_explicit = params.dataset.output_path_explicit;
+            dataset.output_name = params.dataset.output_name;
+            dataset.loading_params = params.dataset.loading_params;
+            params.dataset = std::move(dataset);
```
MrNeRF: Explicit CLI dataset flags get lost here: --images, --resize_factor, --max_width, --test-every, --min-track-length and the timelapse flags are ignored without a message. Checked: `-d project.licht --images images --test-every 4` still trains on images_4 with no eval split. The --resume path calls `core::param::apply_explicit_training_overrides(params, params.overrides)` after restoring; doing the same right after this line fixes it.

src/app/application.cpp
```diff
@@ -116,6 +117,113 @@ namespace lfs::app {
...
+            // From here on this is an ordinary --data-path run.
+            params.dataset_project.reset();
```
MrNeRF: Flow note: the result goes to a fresh project.licht in --output-path and the source project stays untrained, so the user ends up with two files. The GUI trains inside the same project. Worth deciding now whether `-d project.licht` should bind the source project as the destination; that also lines up with the auto-resume idea from your future work list.

MrNeRF (commented):
Built on current master and tested end to end: REFS path, embedded path (extraction and reuse), the rejects, resume of the produced project, and opening the moved project in the GUI. All works. Three points inline.

src/app/application.cpp
thomas-koch-core11: Fixed in [c0e7449](https://github.com/MrNeRF/LichtFeld-Studio/pull/2001/commits/c0e7449f999a6fad23395882f6f857be526d904b). Also added a test.

src/app/application.cpp
thomas-koch-core11: Fixed in [c0e7449](https://github.com/MrNeRF/LichtFeld-Studio/pull/2001/commits/c0e7449f999a6fad23395882f6f857be526d904b). Also added a test.

src/app/application.cpp
thomas-koch-core11: Fixed in [4739be0](https://github.com/MrNeRF/LichtFeld-Studio/pull/2001/commits/4739be04dce9fbf4cb7ba6106c93c8f93d3e33c2).

Now when training headless with a .licht in `--data-path`, the training result is written into the same .licht. Only when `-o` is given, a separate .licht is created and written to.
One thing that should be addressed in a follow-up: `-o` on an embedded project produces a fresh file without the embedded dataset, which was already the case for `--resume -o`. This issue is tracked in #2032

MrNeRF:
Thank you! ☺️

PR #2005, opened 2026-09-03
Add backend-neutral video reconstruction contract

francescofugazzi:
## Summary

This introduces the backend-neutral contract needed to support future video reconstruction backends in Sequencer exports without executing or exposing any vendor-specific implementation yet.

The current user experience remains unchanged: there is no new backend selector, and video export continues to use the existing native rendering and encoding path.

## What changed

- Added a versioned `VideoReconstructionSelection` containing stable backend and preset IDs plus an explicit `Abort` or `Native` fallback policy.
- Added a metadata-only backend catalog interface that must not load plugin code, create GPU resources, or probe devices during lookup.
- Added immutable reconstruction plan resolution with:
  - requested and effective backend/preset provenance;
  - provider version and digest;
  - input and output extents;
  - projection support;
  - required resource flags;
  - structured resolution and fallback reasons.
- Added the always-available `native` backend.
- Persisted the selection under `SEQR.preferences.reconstruction`.
- Migrated projects without the new field to the compatible `native/native` default.
- Resolved the plan before scene capture so invalid selections or explicit fallbacks are handled before rendering begins.
- Added a localized user-facing failure summary across all 10 shipped locales while retaining structured technical diagnostics in logs.
- Added contract and migration documentation.

## Compatibility

- Existing projects without reconstruction settings continue to open with native export selected.
- Existing Sequencer export behavior and UI remain unchanged.
- No vendor enum, numeric quality level, or derived render scale is persisted.
- No DLSS or FSR implementation is selected or executed by this change.
- No native plugin loader, native plugin ABI, or trainer integration is introduced.
- No existing Python plugin API behavior is changed.
- The public contract contains no CUDA types or CUDA-specific behavior.

This is intentionally limited to data, validation, persistence, and preflight planning. Runtime backend registration and reconstruction execution remain separate follow-up work.

## Validation

- Local build and native export smoke test completed by the author.
- Reconstruction selection serialization and migration coverage added.
- Plan resolution, missing backend, explicit fallback, invalid descriptor, projection, preset, and extent coverage added.
- Legacy Sequencer project compatibility coverage added.
- Localization contracts passed for all shipped locales.
- Error-debt baseline check passed.
- Formatting and `git diff --check` passed.


MrNeRF (commented):
Thanks Francesco, this is a clean, well-scoped contract and I have no blocker. I merged it onto today's dev (cc94095d6, merges without conflicts), built on Linux, ran the `VideoReconstruction*` and `P5SessionChapter*` gtests (45 green) and drove all four paths live on a 4080 through the Sequencer export button: a project saved before this PR opens and re-saves with the default selection written in; native export renders 301 frames; a project asking for a missing backend with `abort` shows the localized "Video Export Failed" dialog and writes no file; the same with `native` fallback exports and the log says `requested=com.example.reconstruction/quality, effective=native/native`. Identifier rules, size cap, version/shape checks, the Abort-vs-Native branches, extent rounding and the ten locales all check out.

Four things I would fix before this grows a second backend:

1. **A newer selection version refuses to open the whole project.** `session_state.cpp` validation and the `reconstruction` reader (line 1788 onward) turn `{"version": 2, ...}` into `DataLoss` and `prepareGuiSessionRestore` fails with "The project GUI session cannot be restored." I hit that live: `project_open` on such a file is rejected outright. A video export preference should not lock someone out of their scene. Treat an unknown version as "selection unavailable" (native plus a warning) and keep the hard failure for real corruption. The version-2 case in `test_p5_session_chapters.cpp` currently asserts the strict behaviour, so it would need to flip with the fix.

2. **The writer can save what the reader rejects.** The `reconstruction` field in `session_state.cpp` builds the JSON by hand instead of going through `serializeVideoReconstructionSelection`, so an invalid in-memory fallback lands on disk as `"fallback": "unknown"` (`video_reconstruction.cpp:409`) and the next open fails. Validate before writing, or use the serializer.

3. **There is no way to change or reset the selection.** The persisted value goes straight into the export event and nothing in the Sequencer UI, Python or MCP can set it. With the native-only catalog, a project carrying a non-native + `abort` selection can never export again short of editing the file. Meanwhile `lf.ui.export_video` (`py_ui.cpp:4752`) emits the event with the hard-coded native defaults from `events.hpp:156`, so a script and the export button disagree on which selection is used. Even a small Python setter plus a "reset to native" path would close both gaps.

4. **The fallback is silent.** `video-reconstruction-contract.md` says "No silent fallback is allowed", but on the `native` policy the only trace is the log line at `async_task_manager.cpp:2249`; the plan is not handed to the worker and nothing reaches the user or the provenance stamp. A toast or a line in the completion dialog is enough.

Smaller:
- `video_reconstruction.cpp:286` treats every projection value that is not `Perspective` as equirectangular; a switch with an `InvalidProjection` issue keeps it consistent with the fallback enum validation.
- The resolver accepts extents the encoder cannot take (only positive and even are checked at line 173); the GUI validates first today, but other callers of the public contract will not.
- The identifier grammar (1-128 bytes, lowercase start, `.`/`_`/`-`), the 8 KiB cap and the "unknown members are ignored" rule are not in the docs.
- dev already carries `lfs::vis::SceneUpscalerDescriptor/Preset/Selection` for the viewport (`native`, `spatial`, `temporal`, `nvidia-dlss`). Keeping this contract in `io` is right, but the follow-up should adapt that registry into `VideoReconstructionCatalog` rather than grow two backend lists.
- Pre-existing, not yours: the export failure dialog renders a literal `\n\n` from `runtime.video_export_failed`.

PR #2008, opened 2026-09-03
Add fast source and MSVC preflight gates for Windows builds

francescofugazzi:
Changes to project headers can fail late in the Windows build, while an overly broad preflight repeats unrelated compilation. This adds a portable DLL-export source check and a bounded replay of affected compile-database entries through MSVC `/Zs`.

The configured pass:

- follows quoted and angle-bracket project includes, including deleted headers and both sides of renames;
- compares PR merge checkouts with `HEAD^1`, excluding independent master changes, while push builds use the previous event commit;
- restricts changed paths and compilation entries to `src/`, `include/`, and `tests/`, excluding the selected build directory even when named `b/` or nested under a source root;
- retains generated headers for include resolution without treating newly generated configuration or ABI headers as source edits;
- bypasses sccache and removes dependency-reporting/PCH switches while preserving MSVC quoting, defines, and include arguments;
- limits concurrency to four processes and emits a GitHub warning when the 128-command CI budget skips replay.

Build-only changes do not request a complete replay unless `--all-commands` is supplied. CUDA, vendored, generated, vcpkg, and `_deps` compilation entries are excluded. Source lists show repository-relative filenames and duplicate command counts.

Windows Release/Debug CI and nightly portable packaging first build the `lfs_shader_headers` prerequisite, which exposes the existing CMake shader-generation graph. This prepares headers such as `viewport/frustum.vert.spv.h` before syntax-only replay, propagates generation failures, and lets the subsequent full build reuse the outputs. The prerequisite builds the shader compiler helper and headers without building the visualizer or application. The Python checker itself does not configure CMake, parse its cache, or invoke build targets. ABI headers are seeded at configuration by the now-merged #2018.

The branch includes upstream/master `dc266db24`, containing #2018 and #2028. As requested in the approved #2028 review, its existing profile checks are registered as `lichtfeld.cmake.vcpkg_profiles` with the `fast` CTest label. Documentation covers this registration, local prerequisite generation, configuration selection, and concrete DLL-export checking limitations.

Validation:

- All 40 targeted preflight, shader-dependency, and ABI-stamp tests pass on Python 3.10 after the upstream merge. The 36 preflight/shader tests also passed on Python 3.11.
- Real-Git regressions reproduce the build-only PR with fresh `b/` headers and verify that no unchanged shader consumer is replayed, while genuine source edits remain selected.
- A temporary CMake/Ninja fixture with `LANGUAGES NONE` and a Python header generator verifies generation, incremental reuse, missing-output regeneration, failure propagation, and consumer isolation.
- The actual vcpkg test registration was exercised in a temporary compiler-free CMake fixture: CTest discovery confirms the command, label and timeout, and all 30 Windows/Linux profile cases pass through CTest.
- Source-only contracts, workflow parsing/step ordering, and `git diff --check` pass.

No LichtFeld configuration, native C++/CUDA/GLSL compilation, or runtime execution was performed locally. Include discovery remains lexical; macro includes, ambiguous suffix matches, and linker/runtime behavior still require the full build and tests.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -150,6 +150,15 @@ jobs:
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
 
+      - name: Generate shader headers for preflight
+        shell: cmd
+        env:
+          VCPKG_RUNTIME_SUBDIR: ${{ matrix.build_type == 'Debug' && 'debug\bin' || 'bin' }}
+        run: |
+          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          set PATH=%CD%\%BUILD_DIR%\vcpkg_installed\x64-windows\%VCPKG_RUNTIME_SUBDIR%;%PATH%
+          cmake --build %BUILD_DIR% --target lfs_shader_headers --config ${{ matrix.build_type }} -j 4
+
       - name: Run configured MSVC preflight
         shell: cmd
         env:
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -153,7 +153,7 @@ jobs:
       - name: Run configured MSVC preflight
         shell: cmd
         env:
-          LFS_PREFLIGHT_BASE: ${{ github.event.pull_request.base.sha || github.event.before }}
+          LFS_PREFLIGHT_BASE: ${{ github.event_name == 'pull_request' && 'HEAD^1' || github.event.before }}
         run: |
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           python tools\windows_build_preflight.py --skip-source-checks --compile-commands %BUILD_DIR%\compile_commands.json --max-commands 128
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -14,6 +14,10 @@ on:
   workflow_dispatch:
 
 jobs:
+  windows-preflight:
+    if: github.repository == 'MrNeRF/LichtFeld-Studio' || github.event.pull_request.base.repo.full_name == 'MrNeRF/LichtFeld-Studio'
+    uses: ./.github/workflows/windows-build-preflight.yml
+
   test-plugin-venv:
     if: github.repository == 'MrNeRF/LichtFeld-Studio' || github.event.pull_request.base.repo.full_name == 'MrNeRF/LichtFeld-Studio'
     runs-on: windows-2022
@@ -85,6 +89,7 @@ jobs:
           Write-Host "PASS: uv sync resilient to wrong PYTHONHOME"
 
   build-windows:
+    needs: windows-preflight
     if: github.repository == 'MrNeRF/LichtFeld-Studio' || github.event.pull_request.base.repo.full_name == 'MrNeRF/LichtFeld-Studio'
     runs-on: windows-2022
     strategy:
@@ -144,6 +149,14 @@ jobs:
         run: |
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
+
+      - name: Run configured MSVC preflight
+        shell: cmd
+        env:
+          LFS_PREFLIGHT_BASE: ${{ github.event.pull_request.base.sha || github.event.before }}
+        run: |
+          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          python tools\windows_build_preflight.py --skip-source-checks --compile-commands %BUILD_DIR%\compile_commands.json --max-commands 128
 
       - name: Build
         shell: cmd
--- a/docs/docs/development/windows-build-preflight.md
+++ b/docs/docs/development/windows-build-preflight.md
@@ -10,7 +10,8 @@ replacement for building and testing the application.
 
 ## Source checks
 
-The dependency-free check runs on Windows or Linux and requires only Python:
+The dependency-free check and its self-tests run on Windows or Linux with
+Python 3.10 or newer:
 
 ```sh
 python tools/windows_build_preflight.py --source-only
@@ -24,6 +25,12 @@ present in the DLL import library.
 The rule is intentionally conservative. It ignores broad `using namespace`
 imports, class methods, inline functions, and `constexpr` helpers rather than
 guessing at C++ semantics.
+
+In particular, `tests/test_project_lifecycle.cpp` and
+`tests/test_p5_session_chapters.cpp` use unqualified calls after nested
+`using namespace` directives. Removing `LFS_VIS_API` from
+`capturePanelCameraProjectState` is not detected by this source-only rule.
+The Windows link step is still needed to catch that missing export.
 
 Run the checker self-tests after changing its matching rules:
 
@@ -43,7 +50,7 @@ python tools/windows_build_preflight.py `
   --changed-since upstream/master
 ```
 
-The tool follows project-local quoted includes from each changed header to the
+The tool follows project-local quoted and angle-bracket includes from each changed header to the
 affected C and C++ translation units. It then replays their existing commands
 with MSVC's `/Zs` option. The Microsoft compiler therefore parses the same
 sources with the configured defines and include paths, but does not generate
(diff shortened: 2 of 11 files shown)


tools/windows_build_preflight.py
```diff
@@ -0,0 +1,843 @@
...
+    {
+        ".git",
+        ".pytest_cache",
+        "build",
+        "build-debug",
+        "external",
+        "node_modules",
+    }
+)
+INCLUDE_RE = re.compile(r'^\s*#\s*include\s*"([^"]+)"', re.MULTILINE)
```
MrNeRF: Quoted includes only, here and in the git grep pattern below. src/core/include/core/tensor.hpp:60-61 includes the tensor internals with angle brackets, so the graph sees 9 of 450 consumers for tensor_serialization.hpp and 28 of 463 for tensor_impl.hpp. Accept both forms; the resolution step already ignores anything that does not map to a project file, so system headers stay out.

tools/windows_build_preflight.py
```diff
@@ -0,0 +1,843 @@
...
+        else:
+            diff_base = merge_base.stdout.strip()
+    else:
+        diff_base = "HEAD"
+        if os.environ.get("GITHUB_ACTIONS") == "true" and not requested_base:
+            force_all = True
+
+    changed: set[Path] = set()
+    if not force_all:
+        diff = _git(root, "diff", "--name-only", "--diff-filter=ACMRT", diff_base, "--")
```
MrNeRF: ACMRT excludes D, and --name-only prints only the new side of a rename, so a deleted or moved header never reaches the graph and its consumers are not selected. Drop the filter and add --no-renames; build_reverse_include_graph already takes non-existent paths through extra_dependencies.

.github/workflows/windows.yml
```diff
@@ -145,6 +150,14 @@ jobs:
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
 
+      - name: Run configured MSVC preflight
+        shell: cmd
+        env:
+          LFS_PREFLIGHT_BASE: ${{ github.event.pull_request.base.sha || github.event.before }}
```
MrNeRF: base.sha is master from when the PR was opened, so this diffs against the fork point and replays every master commit merged since. Both runs of this PR show it: no C++ changed, 7 and then 62 TUs replayed. For pull_request use HEAD^1 of the merge checkout (fetch-depth is already 0); keep github.event.before for push.

tools/windows_build_preflight.py
```diff
@@ -0,0 +1,843 @@
...
+            for command in commands
+        }
+        for future in as_completed(futures):
+            command, returncode, output = future.result()
+            if returncode != 0:
+                failures.append((command, returncode, output))
+    return sorted(failures, key=lambda item: item[0].file.as_posix().lower())
+
+
+def generated_header_build_commands(root: Path, database: Path) -> list[list[str]]:
```
MrNeRF: This exists because #2018 stopped generating the stamp at configure time. I asked for that in #2018; once it generates at configure like git_version.h, remove this and prepare_generated_headers. Naming a target of another PR and parsing CMakeCache.txt to build it is more than a preflight should carry.

tools/windows_build_preflight.py
```diff
@@ -0,0 +1,843 @@
...
+        for index in range(len(parts)):
+            suffix_index["/".join(parts[index:])].append(path)
+
+    reverse: dict[Path, set[Path]] = defaultdict(set)
+    for source, include in _iter_local_includes(root, project_files):
+        include_path = Path(include.replace("\\", "/"))
+        direct = _lexical_absolute(source.parent / include_path)
+        resolved: Path | None = direct if direct in dependency_files else None
+        if resolved is None:
+            matches = suffix_index.get(include.replace("\\", "/").lower(), [])
```
MrNeRF, line 505: The suffix index does not know the generated headers under build/include, so `#include "config.h"` in src/visualizer maps to src/rendering/rasterizer/vulkan/src/config.h. An edit to the Vulkan header then selects 59 unrelated TUs. Seed the index with the build directory's generated headers, or resolve against the command's /I paths.

docs/docs/development/windows-build-preflight.md
```diff
@@ -0,0 +1,130 @@
...
+```sh
+python tools/windows_build_preflight.py --source-only
+```
+
+It checks a source contract that otherwise tends to fail only when the Windows
+test executable is linked: a non-inline visualizer free function explicitly
+imported or fully qualified by a test must declare `LFS_VIS_API` so it is
+present in the DLL import library.
+
+The rule is intentionally conservative. It ignores broad `using namespace`
```
MrNeRF, line 25: Two tests hit this today: tests/test_project_lifecycle.cpp and tests/test_p5_session_chapters.cpp call visualizer free functions unqualified after nested using-directives, and removing LFS_VIS_API from capturePanelCameraProjectState passes the check. Name them here so nobody expects coverage there.

tools/windows_build_preflight.py
```diff
@@ -0,0 +1,843 @@
...
+            return 2
+        selected = select_compile_commands(root, commands, changed, forced_all)
+
+        if not selected:
+            print("MSVC preflight: no affected configured C/C++ translation units")
+            return 0
+        if arguments.dry_run:
+            _print_selected_compile_commands(root, selected)
+            return 0
+        if arguments.max_commands and len(selected) > arguments.max_commands:
```
MrNeRF: A skipped replay looks like a green step. Print a ::warning:: annotation here so the skip shows up in the run summary.

MrNeRF (changes requested):
Ran the checker's tests and `--source-only` here, measured the include graph against Ninja's dependency log on a full Linux build (778 TUs, 621 headers), and read the preflight step logs of both CI runs of this PR.

Three things before this can go in:

1. **The include graph misses angle-bracket project includes.** Only `#include "..."` is parsed. tensor.hpp pulls the tensor internals with `<core/tensor/internal/...>` and 262 files include tensor.hpp, so an edit to tensor_serialization.hpp selects 9 of its 450 real consumers, tensor_impl.hpp 28 of 463, cuda_error.hpp 105 of 483. Micro recall over all headers is 0.74. A broken tensor internal passes the preflight and fails 40 minutes later in the full build, which is the case this PR exists for. Inline.

2. **Deleted and renamed headers are never discovered.** `--diff-filter=ACMRT` drops D and `--name-only` shows only the new side of a rename; staged rm, unstaged rm and the old path of a git mv all give an empty change set. The unit test for the removed-header case feeds select_compile_commands a hand-built set and never goes through discovery. Inline.

3. **On CI the replay covers master's delta, not the PR's.** `github.event.pull_request.base.sha` is the base tip from when the PR was opened, so the diff contains every master commit merged since. This PR changes no C++, yet the step replayed 7 and then 62 TUs (main.cpp, abi.cpp, the py_*.cpp files, from the merges of that afternoon), 4.5 min per configuration, and the count grows with the age of the PR until the 128 budget silently switches the check off. Inline.

On 9c38b1716: the header preparation works around #2018 generating its stamp only at build time. I asked for that to be fixed in #2018 (generate at configure, like git_version.h). Once it is, drop generated_header_build_commands and prepare_generated_headers; the preflight should not know a specific target by name and read CMakeCache.txt to build it. Inline.

Smaller:
- `#include "config.h"` in src/visualizer resolves to the Vulkan config.h instead of the generated build/include one, so an edit to the Vulkan header over-selects 59 unrelated TUs. Inline.
- The documented blind spot is real on the tree: test_project_lifecycle.cpp and test_p5_session_chapters.cpp call visualizer free functions unqualified after nested using-directives, and removing LFS_VIS_API from capturePanelCameraProjectState passes `--source-only`. Name those in the doc. Inline.
- The budget skip exits 0 with one log line; a `::warning::` annotation would make it visible in the run summary. Inline.
- /Yc, /Yu and /Fp pass through into the /Zs command. No PCH in the tree today, cheap to strip defensively.
- build/vcpkg_installed nanobind and _deps sources count as project TUs under `--all-commands`; only external/ is excluded.
- The new tests use TestCase.enterContext, which is 3.11+; 17 errors on 3.10 here. Fine for CI on 3.12, just state the floor.
- test_windows_build_preflight.py is not in LFS_PYTHON_FAST_TESTS, so only the Ubuntu job runs it.

Cost for reference: the Ubuntu gate is 10 s. The replay is about 4.3 s per TU at four jobs, up to about 9 min per configuration at the 128 budget on every green PR, and the full build compiles those TUs again. With 1 to 3 fixed the selection stays small for typical PRs and that is fine.

MrNeRF (approved):
All seven points are addressed. Measured on this tree against Ninja's dependency database (787 C++ TUs): with angle-bracket includes tensor_serialization.hpp goes from 9 to 450 of 462 consumers and tensor_impl.hpp from 28 to 463 of 475; export.hpp recall from 0.84 to 0.95. The 28 checker tests pass on 3.10 and 3.12, --source-only passes on master, deleted and renamed headers reach the graph, PR runs diff against HEAD^1, and the ABI header workaround is gone. Goes in after #2018 so the configure-time stamp is on master first.

tools/windows_build_preflight.py
```diff
@@ -0,0 +1,799 @@
...
+    if not force_all:
+        # Treat renames as delete/add so old include spellings remain reachable.
+        diff = _git(root, "diff", "--name-only", "-z", "--no-renames", diff_base, "--")
+        if diff.returncode != 0:
+            force_all = True
+        else:
+            for path in diff.stdout.split("\0"):
+                if path:
+                    changed.add((root / path).resolve())
+            untracked = _git(root, "ls-files", "-z", "--others", "--exclude-standard")
```
MrNeRF: With BUILD_DIR=b in windows.yml this lists the whole build tree. Drop everything outside src/, include/ and tests/ here, and anything under the compile database's directory, before it reaches the include graph.

.github/workflows/windows.yml
```diff
@@ -145,6 +150,14 @@ jobs:
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
 
+      - name: Run configured MSVC preflight
+        shell: cmd
+        env:
+          LFS_PREFLIGHT_BASE: ${{ github.event_name == 'pull_request' && 'HEAD^1' || github.event.before }}
+        run: |
+          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+          python tools\windows_build_preflight.py --skip-source-checks --compile-commands %BUILD_DIR%\compile_commands.json --max-commands 128
```
MrNeRF, line 168: Fails on master content already: vulkan_viewport_pass.cpp includes viewport/frustum.vert.spv.h, generated by compile_shader() at build time. Nothing between Configure and Build can replay that TU.

MrNeRF (changes requested):
Approved too early: the Windows Release run on 2e98c1589 fails inside the new step itself (job 100978151742). Two causes, both visible in that log.

1. The PR changes no C++, yet 56 TUs were selected. CI's build dir is `b`, which .gitignore does not cover, so the untracked-file scan reports every generated file under it as changed; b/include/config.h, git_version.h and lfs_core_abi_stamp.h then pull in their transitive includers. The changed set needs the same src/include/tests filter as the command list, and the compile database's own directory excluded outright.

2. src/visualizer/rendering/passes/vulkan_viewport_pass.cpp fails with C1083 on viewport/frustum.vert.spv.h. Those headers come out of compile_shader() during the build, so any replay that runs between configure and build cannot see them, and every future PR that touches a header on that TU's include path fails the same way. Either the replay builds the generating targets it needs first, or it leaves CI and stays a local tool. The source-only gate is fine on its own.

Inline.

PR #2012, opened 2026-09-03
Viewport rail follows the bottom dock and re-flows instead of scrolling

MrNeRF:
Three viewport rail defects, all reproduced with real pointer input on a 1920x1168 window.

**Rail highlight out of sync with the bottom dock.** Clicking a dock tab, closing the active tab with its x, or hiding the sequencer while the histogram stays open left the rail highlighting the wrong panel. The rail is refreshed by the overlay document sync, which only runs on `viewport_toolbar_generation` (published by `set_panel_enabled`) or a forced size/theme sync. `PanelLayoutManager::setBottomDockActiveTab`, the auto-activation in `renderBottomDock`, and `setShowSequencer` never published anything, so the perf log showed the tab click rendering with `document_dirty=false`. Closing the last tab only looked right because the viewport resize forced a sync. The setters now publish on real change (no-op calls stay silent), with a gtest covering both.

**Rail scrolled and clipped buttons at short viewport heights.** `.toolbar-vertical` had `max-height: 90%; overflow-y: auto`, so a tall dock or a small window produced a scrollbar and a half-visible last button. The rail now re-flows instead: single column when it fits; two columns (editing tools | panel toggles, split by group with a vertical separator) when it does not; compact 26dp buttons with 18dp icons when even that does not fit; overflow hidden below roughly 200dp of viewport height. The tier is measured in `RmlViewportOverlay::updateToolbarRailLayout` after the context update against `viewport height - 24dp`, only when viewport size, DPI, toolbar roots, or toolbar records change. Both toolbar roots (primary and split-view secondary) get the same treatment.

**Histogram button labels sat at the top of their boxes.** `Undock` and friends were inline-block buttons with a taller min-height. They are now flex-centered; because RmlUi drops bare text inside a flex container, the bound labels are wrapped in `span.btn-label`, the same pattern the asset manager uses.

Verified: `lfs-build-guard` build, `*PanelLayout*:*BottomDock*:*ViewportOverlay*` gtests (16), toolbar/overlay/localization/panel pytest (55), and an xdotool session on `--view 3k.ply`: tab click, active-tab close, sequencer close, last-tab close all keep the rail in sync with the loop idling afterwards; two-column rail at the 65% dock height, single column back at the minimum; Undock/Dock/Close/Clear/Delete labels present and centered docked and floating; zero error lines.

### DIFF
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -6513,6 +6513,7 @@ namespace lfs::vis::gui {
         const float bottom_dock_edge_grab_h =
             std::max(PanelLayoutManager::SPLITTER_H * current_ui_scale_,
                      8.0f * current_ui_scale_);
+        const float bottom_dock_grip_h = PanelLayoutManager::DOCK_GRIP_H * current_ui_scale_;
         const bool pointer_over_bottom_dock =
             panel_layout_.isBottomDockVisible() &&
             pointInRect(panel_input.mouse_x, panel_input.mouse_y,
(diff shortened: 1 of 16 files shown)


MrNeRF:
Added b7e3be415: the dock resize grip. The bar under the tab strip was a leftover element inside the histogram panel with nothing bound to it; before #1983 it happened to sit inside the dock's top-edge resize band, and the tab strip moved it 29 px below. The grip is now dock chrome owned by RmlBottomDock, drawn at the dock's top edge above the tabs for every bottom-dock panel; the resize band covers the whole strip, the bar highlights on hover and during the drag, and tab clicks below it still work. Verified by dragging it with xdotool, gtests and pytest green.

PR #2018, opened 2026-09-04
Avoid unnecessary CMake regeneration for core edits and runtime assets

francescofugazzi:
Ordinary core source edits and runtime resource changes currently cause unnecessary CMake regeneration or copies. This moves their refresh work to build-time helpers while preserving ABI validation and resource discovery.

The core ABI header is now also seeded during configuration, so clangd, CLion, and direct syntax checks can use it before the first build. Actual multi-config generators receive one header per configuration; single-config generators use their build type even if a dependency populated `CMAKE_CONFIGURATION_TYPES`.

Headers live under `build/include/core-abi/<config>/`, with the directory itself used for an unnamed configuration. Configuration removes the legacy shared `build/include/lfs_core_abi_stamp.h`, preventing stale fallback through the public include path. The core target owns its private ABI include path and generation dependency.

The always-run helper still refreshes the stamp before the application, core, and test consumers build. Core content hashes, Git identity, compiler/platform/CUDA/architecture/configuration/triplet inputs, and the runtime stale-library guard remain intact. Unchanged content preserves the header timestamp. Core source/header edits do not become configure dependencies.

Runtime resource synchronization:

- discovers assets at build time, copies changed files, and removes stale managed files;
- preserves directory/extension scope and optional legacy overlay behavior;
- validates destination containment and handles recently written Windows files;
- avoids unconditional RmlUI directory replacement.

Single-config non-Debug builds also skip unused vcpkg Debug DLL membership checks. Normal CMake inputs, overlay tracking, Release DLL tracking, and Python stub membership checks remain enabled.

The branch is rebased onto master `1cc94ba82`, including the #2012 icons and the upstream Unicode assertion fix. The configure-time seed allows #2008 to remove its special header-target build workaround.

Validation of the review corrections:

- Four CMake script-mode tests pass on Python 3.10 and 3.11: all single-config variants including an empty build type, multi-config separation, legacy-header removal, unchanged timestamps, core-content changes, and configure/build stamp parity.
- Tests are registered in the fast Python suite with the configured CMake executable.
- Merge-tree checks accept this PR, corrected #2008, and the local Release-only vcpkg branch together without conflicts.
- `git diff --check` passes; ABI documentation is updated.

No application configuration, native C++/CUDA build, or runtime execution was performed locally for this revision. The maintainer's [Linux review of the preceding revision](https://github.com/MrNeRF/LichtFeld-Studio/pull/2018#pullrequestreview-5111065879) separately covers no-op/core-edit timings, resource changes, the stale-library guard, multi-config builds, and GUI execution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -497,13 +497,14 @@ configure_file(
         @ONLY
 )
 
-# App/core ABI tripwire. Recompute it before ABI consumers build, rather than
-# making every core source/header a configure dependency. Define this after the
-# effective CUDA architecture selection so the stamp represents the actual
+# App/core ABI tripwire. Seed headers for IDEs and syntax checks at configure
+# time, then refresh before ABI consumers build without making core source/header
+# edits configure dependencies. Define this after the effective CUDA architecture
+# selection so the stamp represents the actual
 # target configuration. Each multi-config configuration gets its own header so
 # Debug and Release cannot race or overwrite each other.
-set(LFS_CORE_ABI_INCLUDE_DIR "${CMAKE_BINARY_DIR}/include/$<CONFIG>")
-set(LFS_CORE_ABI_HEADER "${LFS_CORE_ABI_INCLUDE_DIR}/lfs_core_abi_stamp.h")
+include(cmake/InitializeCoreAbiStamp.cmake)
+lfs_initialize_core_abi_stamps()
 add_custom_target(lfs_core_abi_stamp ALL
     BYPRODUCTS "${LFS_CORE_ABI_HEADER}"
     COMMAND ${CMAKE_COMMAND}
@@ -596,12 +597,10 @@ add_executable(${PROJECT_NAME}
     resources/lichtfeld-studio.manifest
 )
 
-foreach(_lfs_core_abi_consumer IN ITEMS ${PROJECT_NAME} lfs_core)
-    add_dependencies(${_lfs_core_abi_consumer} lfs_core_abi_stamp)
-    target_include_directories(${_lfs_core_abi_consumer} BEFORE PRIVATE
-        "${LFS_CORE_ABI_INCLUDE_DIR}"
-    )
-endforeach()
+add_dependencies(${PROJECT_NAME} lfs_core_abi_stamp)
+target_include_directories(${PROJECT_NAME} BEFORE PRIVATE
+    "${LFS_CORE_ABI_INCLUDE_DIR}"
+)
 
 # The binary only carries code for these architectures; the app refuses to start on anything
 # lower instead of dying inside the first kernel launch (#1540).
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,35 +162,6 @@ add_custom_target(lfs_git_version ALL
     VERBATIM
 )
 
-# App/core ABI tripwire. Hash every lfs_core implementation and interface input,
-# because a dirty-worktree rebuild cannot rely on the Git commit alone to
-# distinguish a newly linked app from a stale shared library.
-file(GLOB_RECURSE LFS_CORE_BUILD_INPUTS CONFIGURE_DEPENDS
-    "${CMAKE_SOURCE_DIR}/src/core/*.cpp"
-    "${CMAKE_SOURCE_DIR}/src/core/*.cu"
-    "${CMAKE_SOURCE_DIR}/src/core/*.cuh"
-    "${CMAKE_SOURCE_DIR}/src/core/*.hpp"
-)
-list(APPEND LFS_CORE_BUILD_INPUTS
-    "${CMAKE_SOURCE_DIR}/src/core/CMakeLists.txt"
-)
-list(SORT LFS_CORE_BUILD_INPUTS)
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -526,6 +497,35 @@ configure_file(
...
 )
 
+# App/core ABI tripwire. Recompute it before ABI consumers build, rather than
+# making every core source/header a configure dependency. Define this after the
+# effective CUDA architecture selection so the stamp represents the actual
+# target configuration. Each multi-config configuration gets its own header so
+# Debug and Release cannot race or overwrite each other.
+set(LFS_CORE_ABI_INCLUDE_DIR "${CMAKE_BINARY_DIR}/include/$<CONFIG>")
+set(LFS_CORE_ABI_HEADER "${LFS_CORE_ABI_INCLUDE_DIR}/lfs_core_abi_stamp.h")
+add_custom_target(lfs_core_abi_stamp ALL
```
MrNeRF, line 508: This only runs at build time. After a fresh configure the header does not exist, so compile_commands.json points at a missing file (clangd and CLion show main.cpp, abi.cpp and test_assert_hardening.cpp as broken) and anything that compiles before the first build, like the #2008 /Zs pass, fails with C1083 on those three. Run GenerateCoreAbiStamp.cmake once here at configure as well, per configuration for multi-config, like GenerateGitVersion.cmake at line 154.

tests/CMakeLists.txt
```diff
@@ -573,6 +577,7 @@ set_target_properties(lichtfeld_tests PROPERTIES
 
 # Include directories
 target_include_directories(lichtfeld_tests PRIVATE
+    ${LFS_CORE_ABI_INCLUDE_DIR}
```
MrNeRF, line 588: Ordering is right, but existing build dirs still carry build/include/lfs_core_abi_stamp.h from the old scheme and lfs_core exports that directory PUBLIC. A fourth includer without BEFORE would silently pick up the stale stamp. file(REMOVE) it at configure.

cmake/GenerateCoreAbiStamp.cmake
```diff
@@ -0,0 +1,79 @@
...
+    "${LFS_SOURCE_DIR}/src/core/*.cuh"
+    "${LFS_SOURCE_DIR}/src/core/*.hpp"
+)
+list(APPEND LFS_CORE_BUILD_INPUTS
+    "${LFS_SOURCE_DIR}/src/core/CMakeLists.txt"
+)
+list(SORT LFS_CORE_BUILD_INPUTS)
+
+execute_process(
+    COMMAND git rev-parse --short HEAD
```
MrNeRF, line 42: Refreshing this at build time means every commit, checkout or rebase recompiles abi.cpp, main.cpp and test_assert_hardening.cpp and relinks lfs_core and everything downstream (17 s cold, 2 s cached here). Fine as is, since provenance.cpp already forces that relink through git_version.h. If it ever bites, drop the hash from the material; the content hashes already define the ABI.

MrNeRF (changes requested):
Built and exercised the branch on Linux (Ninja, Release, GCC 14.1, CUDA 12.8) with a master build next to it for comparison.

- No-op rebuild: 0.08 s, only the two new always-run edges, stamp mtime unchanged.
- Core edit (splat_data.cpp): 1.8 s and one compile edge here vs 20.2 s on master, where CMake reruns and 40 objects rebuild (26 of them nvImageCodec).
- rcss and locale edit, new .rml, delete: synced without a CMake rerun, stale file removed, soft_shadow.png kept.
- `ninja lfs_core` then start the app: exit 2 with both stamps printed, same as master.
- Ninja Multi-Config: separate Release and Debug headers with different stamps.
- GUI with a 1.9M splat PLY: overlay and locales render, no error lines, clean quit.

One thing before this goes in: nothing writes the stamp header at configure time any more, so after a fresh configure it does not exist until the first build. clangd and CLion show the three includers as broken, and #2008 already had to grow a special case for it (9c38b1716 runs `cmake --build --target lfs_core_abi_stamp` from inside the preflight). Generate it once at configure as well, the way git_version.h is handled, and that code can go. Inline.

Smaller:
- Existing build trees keep the old build/include/lfs_core_abi_stamp.h; remove it at configure. Inline.
- The git hash in the stamp now recompiles three TUs and relinks per commit (17 s cold, 2 s cached here). Acceptable, lfs_core relinks per commit anyway through git_version.h. Inline.
- The `target_include_directories(lfs_core ...)` reach-in from the top level belongs in src/core/CMakeLists.txt.
- Four icons from #2012 are missing from this build because of the base; a rebase picks them up.

MrNeRF (approved):
Configure-time seeding per configuration, the core-abi/<config> directory and the legacy header removal are what I asked for. Checked on d36ce5c33: reconfiguring an existing tree writes include/core-abi/Release/lfs_core_abi_stamp.h in 2 s before any build and removes include/lfs_core_abi_stamp.h, the core-abi include dir comes first for main.cpp, abi.cpp and test_assert_hardening.cpp, a core edit refreshes the stamp and rebuilds the consumers without a CMake rerun, the tripwire starts the app, and the four script-mode tests pass on the vcpkg 3.12 interpreter and through the fast ctest tier.

Issue #2014, opened 2026-09-03
2026-09-02 master merge breaks depth-window Dim/Hide on dev: most splats no longer dim, classification is patchy, and the image shimmers at rest


alexmgee:
# The 2026-09-02 master merge breaks depth-window Dim/Hide classification on dev: displaced bright region, constant shimmer at rest, no re-tracking on camera moves

## Environment

- LichtFeld `dev@b541e3f92` (`model-sam2-v1-87-gb541e3f92`, "Merge branch 'master' into dev",  2026-09-02), Windows 11 Pro, MSVC/VS 2026 (`--allow-unsupported-compiler`)
- CUDA 12.9, RTX 3090 Ti (SM 8.6), Release build, clean worktree — zero local changes
- Vulkan viewport with built-in temporal scene reconstruction active (Temporal/Quality, 0.75 input scale)
- Repro scene: ~10M Gaussians (any scene of similar size reproduces immediately; no training involved, viewer only)

## Description

With the selection depth filter enabled, the Dim render effect appears patchy and broken: the majority of splats are no longer dimmed, and the screen appears to flicker/shimmer. See below:

Nothing selected by depth, Dim mode on:
<img width="2298" height="1607" alt="Image" src="https://github.com/user-attachments/assets/300cddb9-6733-4ca0-992b-86fb299879c2" />
Far pushed out to select everything in the box, Dim mode on:
<img width="2382" height="1609" alt="Image" src="https://github.com/user-attachments/assets/9da184e0-a4ff-472d-932b-6a5dd71a5d1e" />

The bizarre thing is that it seems to work different in the split screen mode. When testing the development of Step 3 of this depth selection tool, I notice that if I move the camera in the extra panel before implementing the depth selection tool, the dimming in that panel works fine:
<img width="2961" height="1882" alt="Image" src="https://github.com/user-attachments/assets/6a84e46b-0a58-4e6a-be2f-c0e66936b609" />

If I do not move the other panel's camera before initiating the depth selection mode, the dim effect is broken similarly in each:
<img width="2820" height="1727" alt="Image" src="https://github.com/user-attachments/assets/7cd23dd3-ccdf-4282-a533-bb2ba01d08e3" />

[...]

PR #2019, opened 2026-09-04
Fix depth-window rendering with consistent 8-bit overlay flags

francescofugazzi:
The host allocates one byte per splat for overlay flags, while the projection and raster shaders still use four-byte elements. This mismatch can access outside the logical buffer range and affects the overlay path involved in #2014.

This change keeps the intended 8-bit storage and makes the host and shaders agree across legacy, tile-batch, and macro rendering paths:

- Use byte-addressed buffers, packing four flags per 32-bit word with final-word padding. No additional Vulkan 8-bit storage feature is required.
- Clear flags before projection and write each byte with atomic OR to preserve neighboring flags and remove stale values when buffers are reused.
- Account for the clear and atomic access in synchronization, honor the projection write-enable gate, and retain flags during raster-only selection updates.
- Correct the shared scratch estimate to one byte per splat.
- Remove the retired `out_radii` shader output to match the existing host pipeline layout.
- Add static contract checks and scripted Vulkan dispatch regressions for padding, buffer reuse, write gating, byte addressing, and barriers.

**Thorough validation is still required before merge.** Rebuild the C++ code and affected Slang variants, run the dispatch regressions, and verify Off/Dim/Hide, dragging and resizing, Native/Temporal Quality, split views, and selection/crop overlays on real GPUs. Vulkan validation and performance checks are also needed; full resolution of all reported symptoms is not yet established.

Related to #2014.

### DIFF
--- a/src/rendering/rasterizer/vulkan/shader/src/slang/alphablend_shader.slang
+++ b/src/rendering/rasterizer/vulkan/shader/src/slang/alphablend_shader.slang
@@ -159,7 +159,7 @@ layout(binding=9) RWStructuredBuffer<float> pixel_depth_weight;
 layout(binding=14) ByteAddressBuffer selection_mask;
 layout(binding=15) ByteAddressBuffer preview_mask;
 layout(binding=16) StructuredBuffer<float4> selection_colors;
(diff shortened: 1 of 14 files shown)


MrNeRF:
Thanks for the fix!

PR #2020, opened 2026-09-04
Fix USD splat CLI opening and preserve Windows Unicode paths

### LINKED ISSUE #1750: Double-clicking an associated USD file fails: CLI view path does not accept .usd
Noticed while adding the .licht association (#1749-area work).

Windows registers .usd/.usda/.usdc/.usdz as LichtFeld file types, but the command-line view path (which double-click uses) accepts only .ply, .sog, .spz, .rad, .resume and mesh formats - so double-clicking a USD file launches the app and errors with "Unsupported file format". Either the view path should accept the USD family (the in-app importer handles them) or the association list should drop them. This predates the .licht work; the shared view-path helper makes the fix a one-list change once the right direction is picked.

francescofugazzi:
Opening a USD splat through a file association or the command line fails with
`Unsupported file format`, even though the existing importer supports it.

- Accept `.usd`, `.usda`, `.usdc` and `.usdz` through `-v`, `--view`, bare positional paths and directory imports.
- Use `wmain` on Windows and convert UTF-16 arguments to UTF-8 before parsing, preserving accented and non-Latin paths.
- Keep the existing Linux entry point and USD importer.
- Add parser regressions for extension case variations, spaces, Unicode paths, directory filtering and ordering.
- Preserve the separate `.licht` project flow.

Fixes https://github.com/MrNeRF/LichtFeld-Studio/issues/1750

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -24,6 +24,8 @@
 #include <cuda_runtime.h>
 #include <filesystem>
 #include <print>
+#include <string>
+#include <vector>
 
 namespace {
     // Apply CUDA driver-level VRAM-reduction knobs BEFORE the primary context exists.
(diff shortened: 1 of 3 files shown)


MrNeRF:
Thx!

PR #2022, opened 2026-09-04
Honor transforms.json mask paths and preserve nested mask lookup

### LINKED ISSUE #1422: transforms.json import ignores per-frame `mask_path`; mask lookup by bare filename fails on subfoldered datasets
## Summary

A nerfstudio `transforms.json` dataset whose images live in per-camera subfolders with repeated basenames (e.g. `images/front/000001.jpg` + `images/back/000001.jpg` with matching masks) fails to import with:

> [Invalid dataset] Mask for image '000001.jpg' is ambiguous across the dataset mask folders.
> Keep masks in the same relative subdirectories as the images or rename them uniquely.

even though the masks are in the matching subdirectories and each frame's `mask_path` is in the JSON. This affects the Blender/NeRF loader, which handles any `transforms.json` dataset whether you open the **dataset folder** or the file directly. COLMAP datasets (those with a `sparse/` model) are unaffected.

## Repro

A multi-camera / rig dataset (e.g. dual-fisheye) exported to nerfstudio `transforms.json`, with per-camera subfolders that share frame-numbered basenames:

```
dataset/                       # no COLMAP sparse/ -> routed to the Blender/NeRF loader
├── transforms.json            # frames: file_path images/front/000001.jpg (+ back/...),
│                              #         mask_path masks/front/000001.png (+ back/...)
├── images/{front,back}/000001.jpg
└── masks/{front,back}/000001.png
```

**Open this folder in LFS** (the normal way to load a nerfstudio dataset) → the error above. With no `sparse/` model present, `ColmapLoader::canLoad` declines and `BlenderLoader` takes it; dragging the `transforms.json` file directly is equivalent. Happy to attach a zip.

## Cause (master @ `faa73cd`)

- `transforms.cpp:510` stores only `_image_path.filename()`, so both frames collapse to `000001.jpg`; the per-frame `mask_path` field is never parsed.
- `blender_loader.cpp:154` resolves masks via `mask_cache.lookup(info._image_name)`, keyed on that bare filename.
- `RecursiveFileCache::lookup` (`filesystem_utils.hpp:243`) tries an exact relative-path match first, then a basename fallback. Since the key is always bare, the exact branch is unreachable; the shared basename `000001.png` hits the ambiguous branch. The error's "same relative subdirectories" advice is thus impossible to satisfy from this loader.
- The COLMAP loader passes the full relative name (`colmap.cpp:873`), so it is unaffected.

## Proposed fix (backward compatible; happy to submit a PR)

Either part alone fixes it; flagging both for your preference before I build:

1. **Honor `mask_path`**: parse the optional per-frame field in `transforms.cpp`, carry it on `CameraData`, and resolve it relative to the transforms file in `blender_loader.cpp` (reusing the existing dimension check), falling back to `MaskDirCache` only when absent.
2. **Make the exact-relative branch reachable**: pass each frame's image path relative to the images root (minus a leading `images/`) to `mask_cache.lookup`, instead of the bare filename.

## Motivation

- **Multi-camera / rig datasets** (dual-fisheye, 360, camera arrays): COLMAP rig-based reconstruction requires *identical* basenames across the per-camera folders during the solve, so the error's "rename them uniquely" advice conflicts with how the data must be built. Preserving the relative subdirectory (which the masks already mirror) is the natural fix.
- **Robustness**: mask discovery by filename-guessing has proven fragile — cf. #547, where a "no masks found" failure was fixed, then resurfaced in a later release and stayed flaky on nightlies. Honoring an explicit per-frame `mask_path` gives a deterministic resolution path that sidesteps those heuristics.
- `mask_path` is the nerfstudio-standard field, so honoring it lets these datasets load without restructuring.

francescofugazzi:
The transforms loader ignores per-frame `mask_path` and searches for masks by image basename. Datasets such as `images/front/000001.jpg` and `images/back/000001.jpg` can therefore produce ambiguous or incorrect mask associations.

- Parse and validate optional per-frame `mask_path`, resolving relative paths against the transforms JSON directory and preserving absolute paths.
- Give explicit mask paths precedence over discovery. Report missing explicit masks when masks are enabled instead of silently selecting a different file.
- Preserve image subdirectories for implicit mask lookup, stripping the leading `images/` component while retaining the legacy unique-basename fallback.
- Prefer matching relative paths across mask extensions and search folders before falling back to a basename, preventing masks from another camera from winning.
- Preserve Unicode paths in the shared file index and defer mask indexing until discovery is needed.
- Retain camera display names and split behavior, and associate mask paths without validating mask files when mask loading is disabled.
- Add importer and mask-cache regressions for explicit, implicit, mixed, nested, Unicode, missing, malformed and dimension-mismatched mask inputs.

Fixes https://github.com/MrNeRF/LichtFeld-Studio/issues/1422

### DIFF
--- a/src/io/formats/colmap.hpp
+++ b/src/io/formats/colmap.hpp
@@ -94,6 +94,8 @@ namespace lfs::io {
         float _center_y = 0.f;
         std::string _image_name;
         std::filesystem::path _image_path;
+        // Explicit transforms.json mask_path, resolved against the JSON directory.
+        std::filesystem::path _mask_path;
         bool _has_image = true;
         lfs::core::CameraModelType _camera_model_type = lfs::core::CameraModelType::PINHOLE;
         int _width = 0;
--- a/src/io/formats/transforms.cpp
+++ b/src/io/formats/transforms.cpp
@@ -50,8 +50,21 @@ namespace lfs::io {
 
         struct ValidatedTransformFrame {
             std::string file_path;
+            std::string mask_path;
             std::array<float, 16> matrix{};
         };
+
+        [[nodiscard]] std::string validated_frame_path(
+            const nlohmann::json& frame, const char* key, const size_t frame_index) {
+            if (!frame.is_object() || !frame.contains(key) || !frame[key].is_string())
+                throw std::runtime_error(std::format("Frame {} must contain a string {}", frame_index, key));
+            std::string path = frame[key].get<std::string>();
+            if (path.empty() || path.size() > MAX_TRANSFORMS_PATH_BYTES ||
+                path.find('\0') != std::string::npos) {
+                throw std::runtime_error(std::format("Frame {} {} is empty, too long or contains a null byte", frame_index, key));
+            }
+            return path;
+        }
 
         [[nodiscard]] float finite_json_float_value(
             const nlohmann::json& value,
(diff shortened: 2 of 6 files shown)


MrNeRF (commented):
Reviewed and exercised locally on Linux (RTX 4080).

- Static read: explicit `mask_path` resolves against the JSON directory exactly like `file_path`, absolute paths pass through, and the exact-before-basename pass in `MaskDirCache` only adds a few map probes per camera. The COLMAP path inherits the same ordering, which is the right call (an exact relative match in a later mask root now beats a basename match in an earlier one).
- `TransformsMaskTest`, `ColmapImageLayoutTest`, `PythonIOTest` and the mask-cache groups: 343 pass.
- Headless training on the issue's layout (`images/{front,back}/000001.jpg` with matching masks), once with per-frame `mask_path` and once relying on discovery: both load 4 cameras with 4 masks and train; no ambiguity error.
- A missing explicit mask fails with MISSING_REQUIRED_FILES only when masks are enabled, and is carried but not validated otherwise, which matches the .licht persistence (`has_mask()` checks existence).

Nit, no need to block: `detail::raw_lookup_key` is now the same thing as `lfs::core::path_to_generic_utf8` from #2021 (merged after this PR was opened); worth swapping when you rebase.

No blockers from my side. Windows CI was still pending when I looked.

PR #2028, opened 2026-09-04
Add opt-in Release-only vcpkg presets for Windows and Linux

francescofugazzi:
Standard vcpkg triplets build both Debug and Release dependencies even when the application only needs Release libraries. This adds optional native x64 Windows and Linux profiles that avoid building unused Debug dependencies.

The new configure/build presets are:

- `windows-release`
- `windows-relwithdebinfo`
- `linux-release`
- `linux-relwithdebinfo`

They use project-local triplet overlays with `VCPKG_BUILD_TYPE=release` for both target dependencies and host tools, preserving the standard triplet names and linkage policies. Existing `build` and `debug` presets retain their behavior.

Release and RelWithDebInfo share the same dependency recipe, while each build directory keeps its own `vcpkg_installed` tree. Compatible packages can be restored through the existing binary cache. The first installation may rebuild packages because the overlay changes their ABI; subsequent reuse still requires matching toolchains, features, and recipes.

RelWithDebInfo retains application debug information and its normal Vulkan validation defaults. It does not request Debug builds of vcpkg dependencies.

Profile validation runs before vcpkg installation and rejects incompatible configurations, multi-config generators, and mismatched target/host triplets. It uses `GENERATOR_IS_MULTI_CONFIG`, avoiding false rejection when OpenMesh populates `CMAKE_CONFIGURATION_TYPES` in a single-config Ninja build.

Documentation covers preset selection, cache behavior, dependency symbols, Debug fallback, and migration of an existing build directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,9 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
 endif()
 
+include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/VcpkgProfiles.cmake")
+lfs_validate_vcpkg_profile()
+
 project(LichtFeld-Studio VERSION 0.5.3 LANGUAGES CUDA CXX C)
 
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/SelectFastLinker.cmake")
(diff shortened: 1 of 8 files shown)


MrNeRF (approved):
Opt-in, and the standard build/debug path is untouched: the guard returns before doing anything unless the release-only overlay is selected, and CI keeps building with the stock triplet. Checked here: the 30 script-mode cases pass, -G really sets GENERATOR_IS_MULTI_CONFIG in script mode, every debug-dir reference in our CMake is behind EXISTS or a glob, and vcpkg install --dry-run with the overlay resolves the whole manifest plan for x64-linux. Not exercised here: the actual Release-only dependency install on Linux, that is a from-source rebuild of every port on this box. Nothing changes for anyone who does not pick the new presets, so this can land.

Smaller: register cmake -P tests/cmake/test_vcpkg_profiles.cmake as a ctest next to the fast tier, like the ABI stamp tests in #2018, so it runs without anyone remembering the command.

Issue #2030, opened 2026-09-04
Viewer crash in hand over from training to edit mode.


jacobvanbeets:
Reproduce
• Start training
• When training finished dialog shows up. click edit mode
• Observe viewer artifacts till GPU crash
Commit: dc266db24

<img width="691" height="868" alt="Image" src="https://github.com/user-attachments/assets/33809e2c-17b9-47c0-95a3-fed6f9c6b9fb" />

francescofugazzi:
If you still can reproduce the problem try PR https://github.com/MrNeRF/LichtFeld-Studio/pull/2039 if it solves it

MrNeRF:
@jacobvanbeets Can you retest latest master please?

jacobvanbeets:
Fixed. Thanks

PR #2039, opened 2026-09-05
Invalidate viewport state when training hands the same model to Edit Mode

francescofugazzi:
Entering Edit Mode moves the existing `SplatData` into a scene node. The
single-model scene path preserves its address, so pointer-only model tracking
could miss the handoff and bypass the renderer's existing model-change reset.

This change tracks the model source (training dataset or scene) alongside its
address. The handoff now invalidates viewport artifacts and reaches the existing
renderer reset path on the next render frame. Pause/resume and training
completion retain dataset ownership; repeated frames with the same source and
address do not trigger additional resets.

The model also stays owned by its scene node until `clearTrainer()` succeeds,
preventing a deferred teardown from destroying it through a local `unique_ptr`.

Related to https://github.com/MrNeRF/LichtFeld-Studio/issues/2030.

### Validation and limitations

**Further testing is required before treating this as a fix for #2030. I have
not been able to test it thoroughly or reproduce the original GPU crash.**

- Local Windows compilation succeeded; the requested GTests passed.
- Manual use performed so far showed no problems.
- Added lifecycle regressions for same-address source transitions and repeated
  frames, plus a SceneManager handoff regression preserving the model address.
- The SceneManager regression does not run a live trainer or Vulkan dispatch.
- Static review, formatting and `git diff --check` completed. The coding agent
  did not configure, compile or run the application/tests.

Please test on the affected setup: complete training and enter Edit Mode both
immediately and after a delay, repeat the transition, orbit/zoom, select/delete
and undo, and verify transforms, appearance correction, save/reopen and PLY
export. On dev, check Native and Temporal Quality separately. Full logs from
any artifacts, hangs or device loss would help confirm whether this addresses
the reported failure.

### DIFF
--- a/src/visualizer/rendering/rendering_manager_vulkan.cpp
+++ b/src/visualizer/rendering/rendering_manager_vulkan.cpp
@@ -1938,13 +1938,20 @@ namespace lfs::vis {
         };
         refresh_content_flags();
         size_t model_ptr = reinterpret_cast<size_t>(model);
+        // Edit-mode handoff moves the same SplatData into a scene node. Its
+        // address does not change, but the trainer's GPU handshake is gone.
+        // Use dataset ownership, not Running/Paused, so completion alone does
+        // not reset the renderer while the trainer still owns its fences.
+        const auto model_source = scene_manager && scene_manager->hasDataset()
+                                      ? ViewportFrameLifecycleService::ModelSource::Training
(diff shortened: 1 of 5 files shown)


MrNeRF:
Should pr into master no?

francescofugazzi:
> Should pr into master no?

ok rebased to master

MrNeRF:
Thank you. Looks good but still need confirmation

Issue #2029, opened 2026-09-04
Model fails to rotate during training. Grid, cameras and upper right gizmo seems to rotate.


jacobvanbeets:
As title says.  It happened once and wasn't able to reproduce. 
Commit: dc266db24
<img width="611" height="417" alt="Image" src="https://github.com/user-attachments/assets/8703315d-e890-46bb-8e24-851513d03900" />

PR #2041, opened 2026-09-05
Recover VkSplat shared scratch after training pause

francescofugazzi:
Related to #2029.

## Problem

Pausing training can detach the shared external rasterizer backing at the B3 boundary. VkSplat retained its installed state after that detach and could attempt to reuse or grow backing that was no longer attached to the arena.

After resuming training, this could leave the viewport presenting its cached splat image while the grid and camera overlays continued to move.

The original recovery also left retained-backing reinstall as a try-only operation. Under continuous training-frame activity, reinstall could therefore be starved even though scratch growth already used a bounded idle-window reservation.

## Changes

- Verify shared backing by device pointer before reuse or growth.
- Reinstall renderer-owned backing when the trainer has detached it.
- Recheck backing ownership after acquiring the renderer arena frame.
- Give reinstall, initial installation, and growth the same bounded 15 ms idle-window reservation.
- Keep the arena synchronization mutex through installation so a new training frame cannot consume the reserved idle window.
- Release only the caller's reservation on success, timeout, or failure.
- Preserve zero-timeout try-only behavior and the existing unbounded installation path for other callers.
- Return typed external-growth outcomes so busy and detached states can defer and retry without misleading error messages.
- Keep actionable diagnostics for CUDA synchronization and callback failures.

## Regression coverage

- Detached backing detection and reinstallation.
- Successful reinstall after an active training frame releases the arena.
- Reinstall timeout cleanup.
- Preservation of an existing render reservation.
- External backing growth results and timeout cleanup.
- Deterministic trainer/metrics lock-order contention.
- Shared-scratch setup occurring before the render frame is acquired.
- Retry handling for legitimate try-lock contention introduced by reservation observation in the tests.

## Validation

Reported from Windows Release:

- build succeeded;
- all `ArenaMetricsContentionTest.*` cases passed for 20 repeated iterations with `--gtest_break_on_failure`;
- `tests/python/test_vksplat_output_lifetime_regressions.py`: 9 passed;
- dataset was exercised through repeated training pause/resume and scratch growth without reproducing the frozen splat viewport.

A longer validation run on another GPU is still recommended because the original failure depends on runtime timing and GPU resource behavior.

### DIFF
--- a/src/visualizer/rendering/vksplat_viewport_renderer.cpp
+++ b/src/visualizer/rendering/vksplat_viewport_renderer.cpp
@@ -3499,6 +3499,21 @@ namespace lfs::vis {
         // in reimportSharedScratchIfGrown() once the render owns the arena frame
         // (which excludes training), so the block is stable when we read it.
 
+        // B3 pause/end can detach the backing on the training thread without
+        // changing this renderer's cached installation flag. Reinstall the same
+        // block before either the capacity fast path or growth uses it again.
+        if (shared_scratch_.block) {
+            shared_scratch_.installed_in_training_arena =
+                lfs::core::GlobalArenaManager::instance().get_arena().using_external_backing(
+                    shared_scratch_.block->device_ptr);
+            if (!shared_scratch_.installed_in_training_arena) {
+                if (!try_install_existing()) {
+                    return std::unexpected("VkSplat shared scratch training rasterizer arena is busy");
+                }
+                shared_scratch_.installed_in_training_arena = true;
+            }
+        }
+
         // Fast path: an installed block already large enough for this frame.
         if (shared_scratch_.block && shared_scratch_.bytes >= required_bytes &&
             shared_scratch_.imported_buffer.buffer != VK_NULL_HANDLE) {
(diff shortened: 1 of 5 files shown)


src/visualizer/rendering/vksplat_viewport_renderer.cpp
```diff
@@ -3499,6 +3499,21 @@ namespace lfs::vis {
...
 
+        // B3 pause/end can detach the backing on the training thread without
+        // changing this renderer's cached installation flag. Reinstall the same
+        // block before either the capacity fast path or growth uses it again.
+        if (shared_scratch_.block) {
+            shared_scratch_.installed_in_training_arena =
+                lfs::core::GlobalArenaManager::instance().get_arena().using_external_backing(
+                    shared_scratch_.block->device_ptr);
+            if (!shared_scratch_.installed_in_training_arena) {
+                if (!try_install_existing()) {
```
MrNeRF, line 3512: This reinstall is try-only: `try_install_external_backing` needs `active_frames_ == 0`, and this thread cannot install while it holds its own arena frame, so after a pause with training running continuously the retained block only comes back if this call lands in the gap between two training frames. That is the same starvation shape #2029 describes, moved from grow to reinstall: the viewport stays on the cached frame until an idle gap appears.

Give the reinstall the same bounded reservation you added to `grow_external_backing` (a `timeout_ms` on `try_install_external_backing`, or install-then-grow inside one reserved window when the backing is missing), and add a test in the shape of `ViewerGrowReservesIdleWindowAfterTrainingFrame`: hold a frame, release it once the reinstall has registered interest, assert the block is installed afterwards.

MrNeRF (commented):
Read statically, no build on my side.

The growth reservation, the pointer based installed check and the ownership re-check after the render frame are sound, and the rewritten contention test is a real improvement (one deterministic inversion, no wall clock budget). One thing to address before I would call #2029 fixed, inline below: the reinstall path still has the starvation shape this PR removes from growth.

Two small notes: the 15 ms window would be better as a named constant next to the other scratch tuning values, and the growth-before-frame ordering the reservation relies on is only stated in a comment; a test pinning that order would protect it. Please also give it a longer run on another GPU, the original report was intermittent.

francescofugazzi:
Addressed the review feedback in a follow-up commit:

- retained scratch backing reinstall now uses the same bounded idle-window reservation as growth, preventing continuous training frames from starving recovery after a B3 detach;
- the 15 ms reservation window is defined by the shared `kSharedScratchIdleReservationTimeoutMs` constant;
- added deterministic coverage that holds a training frame, waits for reinstall to register its reservation, releases the frame, and verifies successful installation;
- added coverage for timeout cleanup and preservation of an existing render reservation;
- pinned the required shared-scratch setup-before-render-frame ordering with a source-contract regression test.

Validation on Windows:

- Release build succeeded;
- all `ArenaMetricsContentionTest.*` cases passed for 20 repeated iterations with `--gtest_break_on_failure`;
- `tests/python/test_vksplat_output_lifetime_regressions.py`: 9 passed;
- manual dataset pause/resume testing completed without the viewport freeze.

A longer validation run on another GPU would still be useful.

PR #2042, opened 2026-09-05
Preserve Unicode in Windows console logs

francescofugazzi:
Complements #2021.

Windows console logs were written as UTF-8 bytes through stdio. When the active console used a legacy code page, valid diagnostics could appear as mojibake, for example `ÔÇö` instead of `—`.

The logger now detects an attached Windows console and writes the formatted UTF-8 log line through `WriteConsoleW` after converting it to UTF-16. This preserves Unicode without changing the console code page, console mode, or the user's terminal configuration.

Redirected stdout and stderr keep the existing byte-oriented UTF-8 path, preserving behavior for files, pipes, CI, and log collectors. Console writes retain CRT-style CRLF handling for multiline messages such as Python tracebacks.

Windows startup logs now include:

```text
LichtFeld logger initialized — UTF-8 console: ✓

### DIFF
--- a/src/core/logger.cpp
+++ b/src/core/logger.cpp
@@ -22,6 +22,7 @@
 #define NOMINMAX
 #endif
 #include <windows.h>
+#include "windows_console.hpp"
 #else
 #include <unistd.h>
 #endif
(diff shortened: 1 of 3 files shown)


MrNeRF:
thx!

PR #2045, opened 2026-09-05
Tensor: order synchronous host copies against the home stream

MrNeRF:
## Summary

`TensorLazyRuntimeTest.ErankExpressionMatchesTorchOnInheritedStream` fails now and then in full-suite runs and never in isolation. When it fails, a handful of scattered rows carry values computed from garbage input (erank 1.1 to 1.4 where 2.998 is expected), which points at the upload, not the kernels.

The cause is a CUDA rule for pageable copies: `cudaMemcpy` from pageable host memory returns once the source has been staged, while its DMA may still be in flight on the legacy default stream. A non-blocking stream neither waits for the legacy stream nor is waited on by it, so a tensor created with `from_vector` under a stream guard and consumed by kernels on that stream can be read before its data has landed. The same rule bites downloads: a `cudaMemcpy` device-to-host copy on the legacy stream does not wait for a producer kernel on a non-blocking home stream.

## Fix

`memcpy_ordered` (internal, `cuda_stream_context.hpp`) keeps the copy a synchronous legacy-stream `cudaMemcpy` and adds the event edge the rest of the library already uses in `Tensor::to`:

- download: the legacy stream waits for the tensor's home stream first
- upload: the home stream waits for the legacy stream afterwards

The host never waits on the home stream itself, so a gated or capturing home stream cannot deadlock (a first attempt that waited on the home stream deadlocked `TensorMultiStreamTest.ExplicitH2DViewMoveAssignmentGuardsDroppedPinnedSource`, whose gate opens only after the copy returns). Tensors without a home stream keep the plain `cudaMemcpy`.

Call sites: `from_vector`, the host-staged `fill_` paths, `copy_from` between CPU and CUDA, the host-staged conversions in `to(DataType)`, `at`, `set_bool`, `get_bool`, `to_vector_int` and `to_vector_int64`. The reads that already synchronize the device first (`item`, `debug_values`, `to_vector`) are untouched.

## Test

```
./build/tests/lichtfeld_tests --gtest_filter='*Tensor*:CudaEventPoolTest.*'
```

1,188 tests pass on this branch. Three tests fail in the build lane for reasons outside the change: the two Python integration tests find no built module there, and `AllocCounterTest.FreshLargeTensorIncrementsCounter` is order-dependent inside that broad filter and passes on its own. The stream, multi-stream and lazy runtime suites pass three times in a row (58 tests each). The busy-copy-engine unit test I tried could not force the race deterministically, so the proof for the flake itself is the full suite on the tensor-backend branch, where the same pattern is applied; I will note the result here when those runs finish.

### DIFF
--- a/src/core/tensor/cuda_stream_context.cpp
+++ b/src/core/tensor/cuda_stream_context.cpp
@@ -70,6 +70,18 @@ namespace lfs::core {
         }
     }
 
+    cudaError_t memcpy_ordered(void* const dst, const void* const src, const size_t bytes,
+                               const cudaMemcpyKind kind, const cudaStream_t stream) {
+        if (stream != nullptr && kind == cudaMemcpyDeviceToHost) {
+            bridgeStreams(stream, nullptr);
+        }
+        const cudaError_t status = cudaMemcpy(dst, src, bytes, kind);
+        if (status == cudaSuccess && stream != nullptr && kind == cudaMemcpyHostToDevice) {
+            bridgeStreams(nullptr, stream);
+        }
+        return status;
+    }
+
     cudaStream_t prepare_inputs_for_stream(
         const std::initializer_list<const Tensor*> inputs,
         const std::optional<cudaStream_t> requested_stream) {
--- a/src/core/tensor/internal/cuda_stream_context.hpp
+++ b/src/core/tensor/internal/cuda_stream_context.hpp
@@ -25,6 +25,17 @@ namespace lfs::core {
         std::initializer_list<const Tensor*> inputs,
         std::optional<cudaStream_t> execution_stream = std::nullopt);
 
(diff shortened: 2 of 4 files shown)


MrNeRF:
Proof from the tensor-backend branch, where the same ordering is applied through two small helpers instead of `memcpy_ordered`: before the change the full `lichtfeld_tests` run failed `TensorLazyRuntimeTest.ErankExpressionMatchesTorchOnInheritedStream` in two consecutive runs (rows 28714 to 28716 in one, 36906 to 36908 and 36933 among 8 rows in the other, all with values computed from garbage input); with the change it passes in two consecutive full runs, and the stream, multi-stream, lazy and event-pool suites pass with the gated-stream test that deadlocked the first attempt.

Issue #1910, opened 2026-08-31
Depth and normal supervision are silently ignored during GUT training


Niloux:
## Summary

When training with GUT enabled, depth supervision and normal supervision can still be enabled through the CLI/UI, and LichtFeld Studio loads/preprocesses the corresponding maps and logs that the losses are enabled. However, the GUT training path never computes these losses or propagates their gradients.

This is a silent correctness issue: users can reasonably believe that geometry supervision is active while training is actually photometric-only.

Observed on current `master` at commit `8e1a6ac23ede27d5d53fa9e2bab113008c553131`.

## Affected options

- `use_depth_loss` / `depth_loss_weight`
- `use_normal_loss` / `normal_loss_weight`
- `normal_consistency_weight`
- `normal_flatten_weight`
- The normal supervision start/end settings consequently have no training effect under GUT

## Source analysis

The training backend selection makes the FastGS-only loss gate false whenever GUT is enabled:

- [`fastgs_path = !params_.optimization.gut`](https://github.com/MrNeRF/LichtFeld-Studio/blob/8e1a6ac23ede27d5d53fa9e2bab113008c553131/src/training/trainer.cpp#L5826)
- [`run_fastgs_gaussian_backward = fastgs_path && update_gaussians_this_iter`](https://github.com/MrNeRF/LichtFeld-Studio/blob/8e1a6ac23ede27d5d53fa9e2bab113008c553131/src/training/trainer.cpp#L5883-L5886)

All auxiliary geometry loss blocks require `run_fastgs_gaussian_backward`:

- [Depth loss](https://github.com/MrNeRF/LichtFeld-Studio/blob/8e1a6ac23ede27d5d53fa9e2bab113008c553131/src/training/trainer.cpp#L6646-L6652)
- [Normal prior loss](https://github.com/MrNeRF/LichtFeld-Studio/blob/8e1a6ac23ede27d5d53fa9e2bab113008c553131/src/training/trainer.cpp#L6774-L6780)
- [Normal/depth consistency loss](https://github.com/MrNeRF/LichtFeld-Studio/blob/8e1a6ac23ede27d5d53fa9e2bab113008c553131/src/training/trainer.cpp#L6950-L6958)
[...]

PR #2047, opened 2026-09-06
Reject unsupported 3DGUT options without breaking existing projects

francescofugazzi:
## Summary

Related to #1910.

Reject unsupported training backend combinations before starting or resuming training, with actionable errors and consistent feedback.

3DGUT rejects IGS+, Mip Filter, Depth Loss, and Normal Loss. Undistort remains supported.

## Changes

- Centralize backend-conflict identification and message context for native callers and the training panel.
- Reject invalid parameter updates through the typed Trainer::setParams result.
- Run start preflight before worker transitions and project preparation, including paused resumes.
- Separate command rejection from training completion and from an accepted request's initialization state.
- Return Resume rejection reasons through Python, MCP and TCP.
- Surface rejected training commands in the GUI without reporting a failed training run.
- Disable incompatible panel controls and refresh displayed values after native rollback, including rollback between UI updates.
- Repair the inactive IGS+ preset when choosing Use MCMC.
- Remove the redundant headless checkpoint load.
- Make TCP tests use loopback destinations and bounded cleanup, without changing production server bindings.

This change does not introduce a separate storage or resume compatibility policy. Invalid persisted combinations follow normal parameter validation.

## Project loading and lazy restoration

Start while the project is still loading reports that loading must finish before retrying.

For direct API Start, a stored session requiring asynchronous trainer restoration is restored without automatically starting training afterward. The caller receives an instruction to retry Start after restoration completes.

The training panel retains its existing restore-and-start flow. Its Resume action requests that flow once, without issuing another Resume while restoration is pending.

## What to test

- Start valid 3DGS and 3DGUT configurations and confirm training advances.
- Try 3DGUT with IGS+, Mip Filter, Depth Loss, and Normal Loss through CLI/configuration and Python. Confirm each request fails with an actionable reason.
- Confirm 3DGUT with Undistort remains accepted.
- Reject an invalid start, correct the settings, start training, then Stop while the viewport is rendering. Confirm no deadlock, false completion notification, or stale error state.
- Issue a second Start while the first is initializing. Confirm only the second request is rejected and the first retains its own initialization outcome.
- Pause training and attempt to resume with incompatible settings through the panel, Python and MCP. Confirm training stays paused, the reason is reported, and MCP does not report success for the rejected command.
- Confirm rejected commands from scripts also produce GUI feedback without a failed-run modal.
- Reject a mid-run parameter edit and confirm the controls return to the effective native settings.
- Choose Use MCMC from the IGS+/3DGUT conflict, switch between strategies, and save the project. Confirm the inactive IGS+ preset no longer prevents saving.
- Attempt Start immediately after opening a project. Confirm the loading/retry message replaces the misleading missing-dataset error.
- Open a project requiring lazy trainer restoration. Confirm direct API Start requests an explicit retry, while the panel retains its restore-and-start behavior.
- Refuse a required project overwrite and confirm the refusal is returned to the caller.
- Resume a valid checkpoint through the headless path and confirm the saved iteration is restored.
- Exercise paused Save As, Stop, project save, and reopen. Confirm checkpoint state and save destination remain correct.
- Confirm Python failures contain the actionable reason without the developer diagnostic dump.
- Run the TCP roundtrip and receive-status tests on Windows. Confirm valid exchanges complete and timeout paths terminate without hanging.


francescofugazzi:
Follow-up: preserve correctable 3DGUT backend conflicts when loading project state.

This keeps legacy `.licht` projects openable when they contain a now-unsupported 3DGUT option combination, while preserving the saved values so users can correct them before starting training.

Runtime validation remains unchanged: CLI and trainer start still reject unsupported 3DGUT combinations, and malformed numeric or structural parameter values are still rejected during project loading.

Validated locally with the focused native regression suite covering parameter validation, PRMS serialization round-trip, and pending project-state restoration.

src/core/parameters.cpp
```diff
@@ -357,6 +357,28 @@ namespace lfs::core {
...
+                return TrainingBackendConflict::DepthSupervision;
+            if (use_normal_loss)
+                return TrainingBackendConflict::NormalSupervision;
+            return TrainingBackendConflict::None;
+        }
+
+        std::string OptimizationParameters::validate_for_storage() const {
+            auto compatible_probe = *this;
+            compatible_probe.gut = false;
+            return compatible_probe.validate();
```
MrNeRF: The `gut=false` copy also waives the IGS+ conflict at load time. Fine as a policy, but none of the storage tests cover `igs+ && gut`, and the next gut-dependent rule inherits this silently. A `validate(Mode::Storage)` flag reads clearer than flipping the backend on a probe.

src/core/parameters.cpp
```diff
@@ -478,8 +500,20 @@ namespace lfs::core {
                                    static_cast<uint64_t>(bilateral_grid_W))
                 return std::format("bilateral grid dimensions are too large ({}x{}x{})",
                                    bilateral_grid_X, bilateral_grid_Y, bilateral_grid_W);
-            if (gut && canonical_strategy_name(strategy) == kStrategyIGSPlus)
+            switch (backend_conflict()) {
+            case TrainingBackendConflict::IGSPlus:
                 return "GUT and igs+ strategy cannot be used together";
+            case TrainingBackendConflict::Undistort:
```
MrNeRF: These strings now go on screen verbatim (see the panel comment), so `gut=true` / `use_depth_loss` jargon and the GUT vs 3DGUT mix matter. Either keep them short and human, or let the panel map `backend_conflict()` to a locale key and leave these for the CLI.

src/io/project/project_chapters.cpp
```diff
@@ -2914,7 +2914,7 @@ namespace lfs::io::project {
                 result.config_file.clear();
                 result.bg_image_path.clear();
                 result.ppisp_sidecar_path.clear();
-                if (const std::string invalid = result.validate(); !invalid.empty()) {
+                if (const std::string invalid = result.validate_for_storage(); !invalid.empty()) {
```
MrNeRF: The leniency stops at the PRMS chapter. The embedded CKPT's parameters still go through the strict `validate()` in `load_checkpoint_params`, so the project opens but its session fails to restore ("Could not read checkpoint parameters"). Details in the summary.

src/python/lfs_plugins/training_panel.py
```diff
@@ -2419,24 +2420,11 @@ def _start_after_consent(self):
...
-                elif button == _gut:
-                    p.gut = False
-                    lf.start_training()
-
             lf.ui.confirm_dialog(
-                tr("training.error.strategy_gut_title"),
-                tr("training.conflict.strategy_gut_start_message"),
-                [btn_mcmc, btn_gut, btn_cancel],
-                _on_conflict,
+                tr("status.error"),
```
MrNeRF: Title and button are `tr()`-ed, the body is the raw C++ string, and we ship ten locales. `lf.ui.message_dialog(title, msg, style="error")` is the one-button API for this. The IGS+ case also lost its one-click Disable GUT here while that state is still reachable through project restore and Python setters.

src/python/lfs_plugins/training_panel.py
```diff
@@ -2419,24 +2420,11 @@ def _start_after_consent(self):
         params = lf.optimization_params()
         error = params.validate() if params and params.has_params() else ""
```
MrNeRF, line 2505: Rather than reporting after the click: `gut_disabled` already greys the GUT box under igs+, the same `disabled_condition_id` on undistort / mip_filter / use_depth_loss / use_normal_loss while gut is on prevents the state instead. Already-enabled options must stay correctable.

tests/test_argument_parser.cpp
```diff
@@ -35,6 +35,43 @@ namespace {
...
+            "--gut",
+        };
+        const auto gut = lfs::core::args::parse_args_and_params(static_cast<int>(std::size(argv)), argv);
+        ASSERT_FALSE(gut.has_value());
+        EXPECT_NE(gut.error().find("3DGUT"), std::string::npos);
+        EXPECT_NE(gut.error().find(test.field), std::string::npos);
+
+        const auto fastgs = lfs::core::args::parse_args_and_params(static_cast<int>(std::size(argv)) - 1, argv);
+        ASSERT_TRUE(fastgs.has_value()) << fastgs.error();
+        EXPECT_FALSE((*fastgs)->optimization.gut);
```
MrNeRF: Also assert the option itself survived (`optimization.undistort` etc. is true). As written a parser that silently drops the flag passes.

tests/test_project_chapters.cpp
```diff
@@ -456,6 +456,22 @@ namespace {
                 .background_image_reference->to_string());
     }
 
+    TEST(ProjectChapterTest, BackendConflictIsPreservedAsCorrectableStoredState) {
```
MrNeRF: PRMS only. The CKPT chapter is where the strict `validate()` still bites, worth the same round-trip with an embedded checkpoint.

tests/test_parameter_manager.cpp
```diff
@@ -242,6 +242,14 @@ namespace {
         EXPECT_EQ(target.getDatasetConfig().images, "images_project");
         EXPECT_FALSE(target.consumeDirty());
 
+        auto backend_conflict = *captured;
+        backend_conflict.mrnf_current.gut = true;
```
MrNeRF: This puts the conflict into the inactive mrnf role while igs+ stays active. The case users hit is the active role, please cover that one too.

MrNeRF (changes requested):
Thanks Francesco, this is the right call for #1910 (please link it) and the core is clean: I built the branch, ran the parameter/CLI/checkpoint/project gtests (178 + 14 green) and the panel pytest (73 green). Two things block it for me, both around what happens to runs that were legal yesterday:

1. **Older 3DGUT checkpoints can no longer be resumed or imported.** The lenient `validate_for_storage()` covers the PRMS chapter only. `load_checkpoint_params` still runs the strict `validate()` (`src/core/checkpoint_format.cpp:222/260`, `src/training/checkpoint.cpp:469`), so a `.resume` trained with `--gut --use-depth-loss` (a no-op loss, the checkpoint itself is fine) now fails in `SceneManager::loadCheckpointForTraining`, in the CLI resume path before the overrides are applied (`application.cpp:192`), and `lf.read_checkpoint_params` returns None so the import panel refuses the file. A project with such an embedded CKPT opens but the session restore fails. Checkpoint reads need the same leniency, and resume needs a way past the flag: either an explicit override or warn-and-clear, since the option never did anything under GUT.

2. **"Parameter updates already reach the shared validation path" is not what happens.** `Trainer::setParams` logs and returns on an invalid update (`trainer.cpp:3651`), and `TrainingManager::startTraining` validates the *previous* trainer params before `applyPendingParams()` runs on the worker. So a script or MCP `training_start` that sets depth loss under GUT starts fine and trains with the stale settings. Pre-existing, but this PR makes it the only guard outside the panel button. Validating the pending candidate in `startTraining` (or propagating the `setParams` failure) closes it.

One correction on scope: `--gut --undistort` is not merely redundant on master, it is broken (the camera keeps its coefficients after `prepare_undistortion`, the GUT kernels apply them on top of the pinhole intrinsics, GT is undistorted: alameda renders a fisheye blob). Fix in #2054; once that is in, `undistort` should come out of the reject list, GUT trains fine on undistorted images.

Smaller items inline. Also the branch conflicts with `dev` in `tests/test_argument_parser.cpp` (two tests added at the same spot, keep both), and the description still says projects are rejected, which the follow-up commit changed.

francescofugazzi:
I addressed all requested changes and merged the latest `dev` into the branch.

- Replaced the backend-mutating storage probe with an explicit runtime/storage validation mode. Storage mode preserves correctable saved settings, while the pre-existing IGS+ / 3DGUT conflict remains invalid.
- Extended storage-safe reads to standalone and embedded checkpoint parameters.
- Checkpoint resume now warns and clears only inherited 3DGUT options that previously had no training effect. Explicit unsupported overrides remain explicit and are rejected.
- Training start now validates and freezes the effective pending ParameterManager candidate before transitioning or launching the worker, covering UI, Python, and MCP callers.
- Backend errors use localized one-button dialogs, while the existing IGS+ recovery actions remain available.
- Incompatible controls prevent new conflicts but remain editable when restoring an already-conflicting project.
- Parser coverage verifies that options survive parsing, and the project tests now cover the active ParameterManager role and embedded CKPT round-trips.
- Both argument-parser tests from the previous `dev` conflict were retained.
- Since #2054 is now included in `dev`, Undistort is no longer treated as a 3DGUT conflict and remains enabled for new runs and checkpoint resume.
- Fixed the embedded-checkpoint fixture so the commit snapshot identity matches the CKPT instance.

The focused C++ tests pass locally, including the pending-start, CLI, checkpoint, project, and 3DGUT/Undistort cases. The Training panel Python suite also passes with 72 tests.

src/training/checkpoint.cpp
```diff
@@ -459,6 +486,7 @@ namespace lfs::training {
                     cli_bg_color_set;
                 lfs::core::param::apply_explicit_training_overrides(
                     loaded_params, loaded_params.overrides);
+                clear_inherited_3dgut_noop_options_for_resume(loaded_params);
```
MrNeRF: This never runs on a real resume: `Trainer::initialize()` validates strictly before `load_checkpoint()` is reached (headless `application.cpp:998`, project `training_setup.cpp:1240`). Verified live with a master-made `--gut --enable-mip` project: exit 1, no warning, no cleared copy.

src/training/checkpoint.cpp
```diff
@@ -58,6 +58,33 @@ namespace lfs::training {
             };
         }
 
+        void clear_inherited_3dgut_noop_options_for_resume(
+            lfs::core::param::TrainingParameters& params) {
+            auto& optimization = params.optimization;
+            if (!optimization.gut) {
```
MrNeRF: Two things once this does run: it keys off the effective `gut` after CLI overrides, so a 3DGS checkpoint resumed with `--gut` gets its depth/normal/mip "cleared as no-ops" although they did train; and the cleared values are committed to the live params (line ~660) and land in the next saved checkpoint with only a log line. Fine if intended, then say so in the description.

src/visualizer/training/training_manager.cpp
```diff
@@ -1098,15 +1102,22 @@ namespace lfs::vis {
 
         // Parameter validation is deliberately synchronous: callers get an
         // immediate rejection without starting a worker or touching the scene.
-        if (auto error = trainer_->getParams().validate(); !error.empty()) {
+        auto start_params = pendingParamsCandidate();
+        if (auto error = start_params.validate(); !error.empty()) {
             return reject_start(std::move(error), lfs::ErrorCode::InvalidArgument);
```
MrNeRF: `reject_start` sets the training error but not `initialization_error_`/`initialization_complete_`, so off-thread `lf.start_training()` waits, sees a clean initialization state and returns `None`. Reproduced on a fresh dataset and on the legacy project. MCP gets `isError` but only the generic text; the reason should travel with the rejection.

tests/test_checkpoint_resume.cpp
```diff
@@ -1230,6 +1230,85 @@ namespace {
...
+        const auto stream_params = lfs::core::load_checkpoint_params(
+            input, std::filesystem::file_size(checkpoint));
+        ASSERT_TRUE(stream_params.has_value()) << stream_params.error();
+        EXPECT_EQ(stream_params->optimization.to_json(),
+                  path_params->optimization.to_json());
+
+        auto target_model = make_checkpoint_test_splat(1);
+        lfs::training::MCMC target_strategy(*target_model);
+        auto resumed_params = make_params_json_test_params(temp_dir);
+        const auto resumed = lfs::training::load_checkpoint(
```
MrNeRF: This bypasses `Trainer::initialize()`, which is where the real resume dies. Worth one case through `Trainer` (or the application resume path) so the test fails the way the CLI does today.

src/python/stubs/lichtfeld/__init__.pyi
```diff
@@ -1988,6 +1988,14 @@ class OptimizationParams:
     def validate(self) -> str:
         """Validate parameter consistency, returns empty string if valid"""
 
+    @property
+    def backend_conflict(self) -> str:
```
MrNeRF, line 1992: Hand-written: stubgen emits the binding docstring ("Stable identifier for the selected backend incompatibility, ...") in multiline form, and `check_python_stubs` fails on a default build. Run `refresh_python_stubs` and commit the output.

MrNeRF (changes requested):
Thanks for the quick turnaround. Round 2 on `d41cc897f`: I built it, ran the parameter/parser/checkpoint/project/manager gtests (185 green), the format suite (108 green) and the panel/property pytests (121 green), then drove the CLI and the GUI over MCP with a legacy project made on master (`--gut --enable-mip`). Most of round 1 is done: storage-mode `validate()` keeps IGS+ invalid and has its test, the panel shows a localized one-button error and keeps the IGS+ quick fixes, the four controls grey out sensibly and stay correctable, the wording is human, Undistort is allowed again, the requested tests exist, and the dev merge dropped nothing. One item is not fixed, and it is the one that matters:

1. **Legacy 3DGUT projects still cannot be resumed.** `--resume legacy/project.licht` on this head fails with `Failed to initialize trainer from project: Invalid training parameters: 3DGUT does not support Mip Filter; disable Mip Filter or select 3DGS`. The warn-and-clear lives in `load_checkpoint()`, but `Trainer::initialize()` (`trainer.cpp:2836`) runs the strict `validate()` first, and both the headless path (`application.cpp:998`) and the project path (`training_setup.cpp:1240` before `load_checkpoint` at 1257) go through it. Same in the GUI: the project opens, the display model shows, the session restore fails in the log, and clearing Mip Filter in the panel does not help because the checkpoint's stored parameters are what gets validated. So the "readable; inherited option is warned about and cleared for resume" rows in the table do not hold. The new test calls `training::load_checkpoint()` directly and passes. Either run the clear (or a Storage-mode check) before `initialize()` on the resume paths, or validate in `initialize()` in Storage mode when a checkpoint is attached. Please add a test that goes through `Trainer::initialize` or the application resume path.

2. **Script callers still do not see the rejection.** `lf.start_training()` returned `None` with no exception on both the legacy project and a fresh dataset with GUT + Mip Filter; `reject_start()` records the training error but never sets the initialization error that `waitForInitialization()` reads. Over MCP the fresh case does error, but with the generic "The training manager rejected the start request"; the reason only shows up in `training_get_last_error`. On the legacy project MCP reported "Training started" while the restore had already failed. The description says the same guard covers UI, Python and MCP; today only the panel's own pre-check gives the user the message.

3. **Stubs again.** The two new docstrings in `__init__.pyi` differ from what stubgen emits, so `check_python_stubs` fails on a default build. `cmake --build build --target refresh_python_stubs` and commit.

Smaller items inline. Two more without a line to hang them on: `VisualizerImpl` still calls `prepareTrainingStartProject()` (`visualizer_impl.cpp:3843`) before the new candidate validation, so a rejected script start can create or bind a project first; and `Trainer::setParams` still logs and returns void, so an invalid mid-session update is dropped silently. Side observation, probably pre-existing: after a corrected run finished 40/40, `runtime_job_wait` still reported the job as failed with the earlier Mip Filter message.

francescofugazzi:
Addressed the remaining review findings in `dc93695d9`.

- Legacy 3DGUT checkpoints now pass through the real `Trainer::initialize()` resume path. Standalone `.resume`, embedded `.licht`, GUI, CLI, and TCP restore paths can read configurations that were accepted by older releases.
- Resume normalization is based on the backend stored in the checkpoint, not merely the effective overridden backend. A 3DGS checkpoint switched explicitly to 3DGUT therefore keeps its supervision settings explicit and is rejected instead of being incorrectly sanitized.
- Only inherited Mip Filter, Depth Loss, and Normal Loss settings from an actual legacy 3DGUT checkpoint are warned about and disabled. Undistort remains supported.
- Explicit unsupported overrides are never silently changed and return the corresponding rejection reason.
- The normalized live parameters are used by subsequent checkpoints, while the original checkpoint remains unchanged.
- Embedded `.licht` restore is covered through the shared project-checkpoint installer and verifies the restored iteration, effective parameters, and subsequent checkpoint serialization.
- Project PRMS remains independent from CKPT state. Restoring a checkpoint no longer replaces unrelated editable project settings.
- Start preflight now validates the effective pending parameters before project binding, state transition, or worker launch.
- Python and MCP callers receive the actual initialization rejection instead of a successful/empty result or only a generic manager error.
- `Trainer::setParams()` and pending mid-run updates propagate rejection reasons instead of silently discarding invalid values.
- A successful run clears stale errors left by an earlier rejected attempt.
- The generated Python stubs and committed stubs are synchronized.

Test results:

- Focused native suite: 63 passed, 0 failed, 1 skipped.
- The skipped `SecondRunCompactsLeftoverProjectFile` test is an existing policy conflict between checkpoint-history retention (#1943) and the old no-growth expectation (#1696), unrelated to this change.
- Training panel Python suite: 77 passed.
- Generated Python stubs match the committed stubs.
- All ten locale JSON files parse successfully.
- Diff whitespace check passes.

src/visualizer/training/training_manager.cpp
```diff
@@ -1116,6 +1108,56 @@ namespace lfs::vis {
...
+            .detection = LFS_SOURCE_SITE_CURRENT(),
+        });
+        {
+            std::lock_guard lock(initialization_mutex_);
+            start_params_candidate_.reset();
+            initialization_complete_ = true;
+            initialization_error_ = typed;
+        }
+        initialization_cv_.notify_all();
+        state::TrainingCompleted{
```
MrNeRF: Emitting `TrainingCompleted{success=false}` for a start that never happened is the root of two live problems: the window switches to the failed-run presentation (Error header, all controls greyed, Reset Training, failed-training modal and error-bus notification) while the manager is still Ready, and the next Stop after a corrected start hangs forever (the GUI reclaims the shared scratch arena in `releaseScratchOnIdle` while the trainer is mid-backward; backtrace in the review). Publish a start-rejection signal instead.

src/core/parameters.cpp
```diff
@@ -478,8 +523,12 @@ namespace lfs::core {
                                    static_cast<uint64_t>(bilateral_grid_W))
                 return std::format("bilateral grid dimensions are too large ({}x{}x{})",
                                    bilateral_grid_X, bilateral_grid_Y, bilateral_grid_W);
-            if (gut && canonical_strategy_name(strategy) == kStrategyIGSPlus)
-                return "GUT and igs+ strategy cannot be used together";
+            const auto conflict = backend_conflict();
+            if (conflict == TrainingBackendConflict::IGSPlus ||
+                (mode == ParameterValidationMode::Runtime &&
```
MrNeRF: Storage mode exempts Mip/Depth/Normal but not IGS+. Combined with per-strategy presets this makes the dialog's own Use MCMC quick fix a trap: the inactive `igs+` preset keeps `gut=true`, and Save As / Start then fail with `PRMS.presets.igs+.current: 3DGUT cannot be used with IGS+` while the panel looks valid. Reproduced live.

src/python/lfs_plugins/training_panel.py
```diff
@@ -2419,24 +2481,65 @@ def _start_after_consent(self):
...
+                )
+                if is_backend_conflict
+                else ""
+            )
+            if is_backend_conflict and conflict == "igs_plus":
+                btn_mcmc = tr("training.conflict.btn_use_mcmc")
+                btn_gut = tr("training.conflict.btn_disable_gut")
+                btn_cancel = tr("training.conflict.btn_cancel")
+
+                def _on_conflict(button, _mcmc=btn_mcmc, _gut=btn_gut):
```
MrNeRF, line 2533: Use MCMC switches the active strategy but leaves GUT on in the IGS+ preset it just left. The next save is refused (see the parameters.cpp comment). Either clear GUT in that preset here, or stop validating inactive presets.

src/training/checkpoint.cpp
```diff
@@ -58,6 +58,46 @@ namespace lfs::training {
             };
         }
 
+        void clear_inherited_3dgut_noop_options_for_resume(
```
MrNeRF: There are no legacy projects (v0.5.3 predates `.licht`). This normalization, the storage validation mode, the PRMS mirroring in project_lifecycle.cpp and the override provenance all exist for files no shipped build can have written. Drop them; an invalid stored combination can fail with the same message as a fresh start.

src/visualizer/visualizer_impl.cpp
```diff
@@ -3837,6 +3837,12 @@ namespace lfs::vis {
                 trainer_manager_->getActionBlockedReason(
                     TrainingAction::Start)));
         }
+        if (auto preflight =
```
MrNeRF, line 3848: Two paths bypass this preflight: the lazy-restore branch above returns success before the trainer exists (MCP reports "Training started", the later refusal is silent, `training_get_last_error` stays empty), and the paused branch resumes without it (the rollback then catches it mid-run, but the panel keeps the stale checkmarks all run).

src/visualizer/training/training_manager.cpp
```diff
@@ -2250,7 +2302,15 @@ namespace lfs::vis {
...
             if (auto* pm = services().paramsOrNull(); pm && pm->consumeDirty()) {
-                applyPendingParams();
+                if (auto applied = applyPendingParams(); !applied) {
+                    auto typed = std::move(applied).error();
+                    last_error_ = typed.user_message().empty()
+                                      ? std::string(typed.detail())
+                                      : std::string(typed.user_message());
+                    last_training_error_.set(typed);
+                    pm->importTrainingParams(trainer_->getParams());
+                    LOG_ERROR("Rejected training parameter update: {}", last_error_);
```
MrNeRF, line 2316: The rollback restores the values (verified) but the panel keeps GUT and Mip checked for the rest of the run, even after switching tabs. Worth refreshing the bound display here.

src/app/application.cpp
```diff
@@ -738,7 +737,24 @@ namespace lfs::app {
...
-                        trainer->setParams(effective_params);
+                        // Restore legacy checkpoints before exposing the TCP
+                        // manager. initialize() auto-loads the checkpoint and
+                        // normalizes inherited no-op flags before start preflight.
+                        if (effective_params.resume_checkpoint) {
+                            if (effective_params.optimization.enable_eval) {
+                                trainer->set_lpips_weights_path(
+                                    prepare_lpips_weights(!effective_params.no_download));
+                            }
+                            if (auto initialized = trainer->initialize(effective_params); !initialized) {
```
MrNeRF: Same shape as the headless resume path further down: `initialize()` now auto-loads the checkpoint, and the headless path still calls `load_checkpoint` again afterwards, so the resume warning appears twice. Worth checking this TCP path does not repeat the load when the manager starts.

MrNeRF (changes requested):
Round 3 on b10935d52. Built, unit-tested, and driven live (CLI, Python, MCP, the real window with screenshots).

The round-2 blocker is fixed and proven: a 3DGUT + Mip checkpoint from an older build resumes headless, from a standalone `.resume`, and in the GUI, with the warning, iteration 20 restored, Mip cleared in the new save and the source untouched. Explicit `--enable-mip` on it is refused, a 3DGS + normal-loss checkpoint is refused under `--gut`, fresh conflicts stop at parse time, `--gut --undistort` trains on fisheye, Python raises with the reason, MCP returns `isError`, no project is created, the dialogs work in EN and DE, and the checkbox greying holds under real clicks in both directions. Stubs match the generator; the three failing `VisualizerImplResetTest` cases fail identically on master.

**First the framing.** There are no legacy projects. v0.5.3 predates the `.licht` format, so no shipped build has ever written one. Most of this PR's source changes exist to keep such files resumable: the storage validation mode, the inherited-flag normalization on resume, the PRMS mirroring, the override provenance, the TCP reorder. Every live problem below sits in or next to that layer. Please shrink the PR to the validation rule, the start preflight that returns the reason, the panel greying, and the typed `setParams`. A file with an invalid combination can fail with the same message.

**Blockers found live**

1. **Stop hangs after a rejected start.** Reject a scripted start (GUT + Mip), fix the setting, start, press Stop: the run freezes at its iteration and never reaches Finished. 4/4 with a prior rejection (3DGS and 3DGUT); 9/9 clean stops without one, including on master. Backtrace: the training thread is blocked on the arena-manager mutex inside the rasterizer backward (`release_forward_context`), the GUI thread is in `releaseScratchOnIdle -> releaseSharedScratchArena -> clear_external_backing` reclaiming the shared scratch arena, the reaper is joining the stuck thread. The fake `TrainingCompleted{success=false}` from `rejectStart` leaves the window believing the trainer is idle.
2. **"Use MCMC" leaves the project unsaveable.** IGS+ with GUT, Start, press the dialog's own Use MCMC: the panel looks valid, but Save As and Start fail with `PRMS.presets.igs+.current: 3DGUT cannot be used with IGS+`. The inactive IGS+ preset keeps `gut=true` and storage validation exempts Mip/Depth/Normal but not IGS+. Only re-selecting IGS+ and unticking GUT recovers. Disable GUT does not have this problem.

**Should fix**

3. A rejected scripted start flips the panel to Error, greys every control, shows the Training Failed modal, replaces Start with Reset Training, and raises a failed-training notification on the error bus, while the manager and `lf.trainer_state()` say ready. A rejection needs its own signal, not a completion event.
4. `training_start` on a project whose trainer is restored lazily returns success before the restore; the later refusal (overwrite consent) is silent and `training_get_last_error` stays empty.
5. Resume from Pause skips the preflight. The mid-run guard rejects the edit and rolls back, which is fine, but the panel keeps GUT and Mip visibly checked for the rest of the run.

Nits: the Python exception carries the whole developer dump with source frames; headless `.resume` loads the checkpoint twice (the warning appears twice).

Not attributed: one SIGSEGV in MCMC relocation about a second after a mid-run rejection, 0/6 in controlled repeats.

francescofugazzi:
I narrowed the change as requested and addressed the third-round findings.

Removed the Storage validation mode, inherited checkpoint-option normalization, PRMS mirroring, and eager TCP restoration. Invalid persisted settings now follow normal parameter validation rather than a separate compatibility policy.

Start preflight rejection now emits TrainingStartRejected, not TrainingCompleted, so rejecting a request no longer sends a false completion notification to lifecycle/rendering listeners.

Paused resumes now run parameter preflight. The Use MCMC action repairs the IGS+ preset being left while preserving 3DGUT on MCMC, and the panel republishes its bindings when native parameter rollback changes the effective settings.

Python receives the user-facing rejection reason rather than the developer diagnostic dump. The redundant headless checkpoint load was also removed.

For lazy project restoration, Start now requests restoration without deferred automatic training and returns an explicit instruction to retry after restoration completes, rather than reporting success before the operation is ready.

The requested native tests passed after rebuilding. An additional automated lifecycle selection passed 31 tests, with one existing checkpoint-retention policy skip. Panel/confirmation coverage passed 87 Python tests.

The exact interactive rejected-start → correction → renderer-active Stop sequence has not been reproduced end-to-end, and the asynchronous restore/retry interaction still needs live review. CUDA primary-context release warnings also remain visible at test-process shutdown; I am not claiming those are resolved by this change.

MrNeRF (changes requested):
Round 4, head 749ce25b4, built and run live on Linux (unit + python + format tests, headless CLI, GUI on Xvfb with MCP and clicks).

Good news first. The round-3 regressions are gone:
- Stop after a rejected-then-corrected start: 3 fresh runs plus a control, all Stopping -> Finished in under 100 ms.
- "Use MCMC" no longer leaves the IGS+ preset broken: Save As, reopen, switch back to IGS+ and Start all work.
- A rejected start no longer produces the Training Failed modal, the Error state or the greyed panel.
- Paused resume runs the preflight; the invalid GUT+Mip resume never runs, training stays paused.
- The compat layer is really gone (no storage mode, no inherited-option normalization, no PRMS mirroring, no eager TCP restore, old locale keys removed).

Two things still need fixing before merge, both reproduced live:

1. Resume rejections are silent. `resumeTraining()` still returns void, so `lf.resume_training()` returns None, the panel Resume button just stays paused with no message, and the MCP job control resume answers with `isError: true` but `success: true` in the same payload. The new `TrainingStartRejected` event has no GUI subscriber, so a start rejected from Python or MCP shows nothing in the panel either (the panel's own Start button is fine because it shows the dialog itself). Please return a typed result from resume and surface it in the panel the same way Start does.

2. A second Start while the first one is initializing overwrites the first request's result. `rejectStart()` resets `start_params_candidate_`, sets `initialization_complete_` and the error, and wakes every waiter. Live: two concurrent `lf.start_training()` calls both raise "Training is starting." while the run happily continues to the end. Keep the rejection of request B out of request A's initialization state.

Smaller:
- Start right after opening a project shell says "No dataset loaded" instead of the new restore/retry message; the second attempt says restoring, the third works. The first message is wrong.
- After a rejected mid-run parameter change the panel shows the requested (rejected) checkmarks, not the values that are actually in effect.
- The PR text says lazy Start only restores and needs an explicit retry; the panel still auto-starts after restore (`_restore_stored_session_if_needed(then_start=True)`). Fine either way, but the text should say so.

Pre-existing, not yours: the 3 VisualizerImplResetTest failures about tests/data/bike.ply fail the same way on master.

francescofugazzi:
Addressed the latest review feedback:

- Resume now returns a typed result, propagated to Python, MCP and TCP instead of silently discarding rejection reasons.
- Training command rejections reach the GUI through a warning notification, without reporting a failed training run.
- A rejected second Start no longer clears the first request's accepted parameters, initialization result or waiters.
- Start during project loading reports the loading/retry requirement instead of a misleading missing-dataset error.
- Backend controls are refreshed even when a rejected edit is rolled back between UI updates.
- The panel's stored-session Resume path no longer issues a duplicate Resume after requesting restore-and-start.

The PR description now distinguishes the existing panel restore-and-start flow from direct API Start, which requires an explicit retry after restoration.

Additional TCP testing exposed a Windows test portability issue: the clients connected to the wildcard bind address. The tests now use loopback with the assigned port, bounded send/receive timeouts and zero linger. Missing responses stop the test before JSON parsing. Production TCP listening behavior is unchanged.

The requested native tests passed after rebuilding. Additional checks passed: 99 Python tests, 23 complementary native tests and all 6 TCP tests. Interactive GUI behavior has not been manually rechecked.

MrNeRF:
Thank you!

PR #2048, opened 2026-09-06
Expose consistent 3DGS and 3DGUT training backend identities

francescofugazzi:
## Summary

Depends on #2047 and must be merged after it.

Expose explicit `3dgs` and `3dgut` training backend identities through JSON,
`--raster-backend`, and Python, while retaining `gut` as the single backing
value and preserving the existing `--gut` option.

3DGS remains the default when no backend is specified. Training and viewer
APIs use the same backend vocabulary.

## Changes

- Centralize backend identifiers, labels, technical descriptions, viewer mapping,
  and high-level capability information.
- Use the named identity for training/evaluation dispatch and the capability table
  for backend conflict checks.
- Expose backend selection and capability information through Python.
- Report the active trainer's backend in the status bar, or the saved checkpoint's
  backend before trainer restoration. Do not assume 3DGS when identity is unavailable.

3DGUT supports Undistort. IGS+, Mip Filter, Depth Loss, and Normal Loss remain
unsupported.

## Parameter behavior

- JSON accepts either `gut` or `raster_backend` alone. New output writes both
  consistently.
- When both fields are present, they must agree. Conflicts are errors in configs,
  checkpoint parameters, and project presets.
- Unknown names and incorrect JSON types are rejected.
- Explicit CLI backend selection overrides a valid configuration/checkpoint
  selection and records both aliases for checkpoint overrides.
- Invalid configurations fail before CLI overrides are applied.
- `--gut` remains an alias for 3DGUT. Contradictory CLI options fail independently
  of argument order.
- Python `set("raster_backend", value)` rejects unknown names and non-string
  values with ValueError without changing the selection. None retains the
  existing binding-level TypeError.

Strict parameter validation and start/resume behavior come from #2047.
There is no separate storage-validation mode, inherited-option normalization,
or older-writer precedence policy.

## Scope

The Training panel layout remains unchanged.

Viewer hand-off and existing panel consumers still use `gut`. Adding another
backend requires implementation, storage and consumer migration, and explicit
dispatch; adding a descriptor alone does not install a rasterizer.

Next-run parameter changes do not replace an active trainer or viewer.

## What to test

- Train and evaluate with 3DGS and 3DGUT, including 3DGUT with Undistort.
- Confirm 3DGS is selected when no backend is specified.
- Compare `--gut` with `--raster-backend 3dgut`.
- Reject contradictory CLI flags in both orders and unknown backend names.
- Load matching JSON aliases and either field alone. Reject disagreeing aliases
  and non-string backend values.
- Override a valid 3DGUT config/checkpoint with `--raster-backend 3dgs`, then
  resume without flags and confirm the saved selection persists.
- Save and reopen valid projects/checkpoints, preserving current and session
  parameter selections.
- Reject unsupported 3DGUT combinations, correct the settings, and start training.
- Exercise Python properties and get/set, including invalid values, and confirm
  rejected assignments leave the selection unchanged.
- After a 3DGUT run, confirm the bottom status bar reports 3DGUT.
- Reopen that project and confirm the backend is correct before trainer restoration.
- Change next-run or viewer settings and confirm they do not relabel the saved
  or active training session.
- Exercise pause, resume, Stop, Save As, masks, segmentation, background, and
  appearance settings.

### DIFF
--- a/docs/docs/development/training-backend-contract.md
+++ b/docs/docs/development/training-backend-contract.md
@@ -52,16 +52,13 @@ result instead of duplicating UI text in the descriptor.
   and Python writers of `gut` therefore remain effective.
 - JSON accepts `raster_backend: "3dgs"` or `"3dgut"`. Files containing only
   `gut` retain their original meaning. Missing both selects the existing 3DGS default.
-- New JSON writes both names consistently. While `gut` remains the compatibility
-  storage, its value takes precedence if a previous application preserved a stale
-  known `raster_backend` field while changing `gut`. Unknown identifiers and
-  incorrect JSON types are still errors.
-- This precedence also applies to `--config` JSON. If both fields disagree,
-  `gut` wins and a warning is logged. When editing a config, update both fields
-  consistently or remove `gut` to select solely through `raster_backend`.
-  For example, `{"gut": true, "raster_backend": "3dgs"}` still selects 3DGUT;
-  `{"raster_backend": "3dgs"}` selects 3DGS. An explicit CLI backend selection
-  overrides either valid config after loading it.
+- New JSON writes both names consistently. If both fields are present, they
+  must agree. A disagreement, unknown identifier, or incorrect JSON type is an
+  error, including in configs, checkpoint parameters, and project presets.
+- For example, `{"gut": true, "raster_backend": "3dgs"}` is rejected;
+  `{"raster_backend": "3dgs"}` selects 3DGS. Config authors must update both
+  fields consistently or specify only one. Explicit CLI selection overrides a
+  valid config, not an invalid file that already failed loading.
 - `--raster-backend 3dgs|3dgut` is additive. `--gut` remains an alias for 3DGUT.
   `--gut --raster-backend 3dgs` is an error, independent of argument order.
 - Explicit CLI selection overrides a valid configuration's backend. Its captured
@@ -70,14 +67,10 @@ result instead of duplicating UI text in the descriptor.
   fail their existing load-time validation before CLI overrides are applied.
 - Checkpoint parameter JSON and `.licht` parameter presets use the same adapters.
   The binary checkpoint format and `.licht` chapter schema are unchanged.
-- A stored `.licht` preset may retain an unsupported 3DGUT option so legacy
-  projects can still open without silently changing user data. Storage validation
-  continues to reject malformed values; trainer validation reports the backend
-  conflict when Start is requested.
-- Older applications ignore the new field and continue to read `gut`. In `.licht`,
-  they can preserve the unknown new field while changing `gut`; the precedence
-  rule above ensures that the file still reopens with the legacy application's
-  selected backend.
+- Stored parameters follow the strict validation established by PR #2047.
+  Unsupported 3DGUT combinations are rejected by validating load paths, as
+  well as start/resume preflight. There is no separate storage-validation mode,
+  inherited-option normalization, or older `.licht` writer precedence rule.
 
 ## Python and viewer lifecycle
 
@@ -87,6 +80,9 @@ notification path. `backend_capabilities` returns the complete high-level map fo
 the selected backend using the strings `supported` and `unsupported`.
 `lf.training_backends()` lists IDs, labels, descriptions, viewer IDs, and the
 same capability map without needing to change the current selection.
+`set("raster_backend", value)` rejects non-string values and unknown names with
+an actionable `ValueError`, without changing the selected backend. `None` retains
+the existing binding-level `TypeError` before setter dispatch.
 
 These are next-run parameter APIs, not an atomic command to replace an active
 trainer and viewer. The existing panel change path still updates its viewer
@@ -96,8 +92,11 @@ and active trainer state. The future RmlUi selector should use the descriptors
 and the existing scene/lifecycle commands rather than mutating active training
 from a parameter setter.
 
-Training and evaluation dispatch use the named identity. The existing viewer
-hand-off in `application.cpp` and `visualizer_impl.cpp`, status-bar label, and
+Training and evaluation dispatch use the named identity. The status bar uses
+the active trainer's backend or, before restoration, the saved checkpoint's
+backend. Missing identity is omitted rather than presented as 3DGS. Labels come
+from the backend descriptors, independently of viewer and next-run choices.
+The existing viewer hand-off in `application.cpp` and `visualizer_impl.cpp` and
 training panel still consume the compatibility `gut` boolean. These consumers
 must also migrate before a backend beyond the current two can be installed.
 Training and viewer identifiers use the same `3dgs`/`3dgut` vocabulary.
@@ -107,7 +106,7 @@ Training and viewer identifiers use the same `3dgs`/`3dgut` vocabulary.
 This is a compatibility step, not runtime backend registration. An additional
 backend needs its implementation, descriptor, an explicit state for every
 high-level capability, explicit dispatch, and a replacement for boolean `gut`
-storage with legacy accessors. Unknown enum values are rejected by the
+storage with accessors for the existing API. Unknown enum values are rejected by the
 compatibility setter. Do not map an additional backend to either boolean value or
 assume that adding a descriptor installs a rasterizer.
 
@@ -120,7 +119,7 @@ runtime libraries on the library search path. The chapter test belongs to
 ```powershell
 .\build\tests\lichtfeld_tests.exe '--gtest_filter=TrainingParametersTest.*:ArgumentParserTest.*Backend*:ArgumentParserTest.Gut*:TrainerConstructionTest.*:CheckpointParamsJsonTest.*:ParameterManagerTest.PendingProjectRestoreChangesOnlyRoleQualifiedManagerState' --gtest_color=no
 .\build\tests\lichtfeld_format_tests.exe '--gtest_filter=ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility' --gtest_color=no
-python -m pytest tests/python/test_property_system.py -q -p no:cacheprovider
+.\build\vcpkg_installed\x64-windows\tools\python3\python.exe -m pytest tests/python/test_property_system.py -q -p no:cacheprovider
 ```
 
 On Linux, with the build's Python module and libraries available:
@@ -128,15 +127,18 @@ On Linux, with the build's Python module and libraries available:
 ```sh
 ./build/tests/lichtfeld_tests --gtest_filter='TrainingParametersTest.*:ArgumentParserTest.*Backend*:ArgumentParserTest.Gut*:TrainerConstructionTest.*:CheckpointParamsJsonTest.*:ParameterManagerTest.PendingProjectRestoreChangesOnlyRoleQualifiedManagerState'
 ./build/tests/lichtfeld_format_tests --gtest_filter='ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility'
-python3 -m pytest tests/python/test_property_system.py -q -p no:cacheprovider
+./build/vcpkg_installed/x64-linux/tools/python3/python3.12 -m pytest tests/python/test_property_system.py -q -p no:cacheprovider
 ```
 
-Use the Python interpreter matching the built extension. Local shell helpers
+These paths assume the standard vcpkg triplets. Use the Python interpreter
+matching the built extension, with pytest installed and the built module and
+its runtime libraries available; a system Python is not interchangeable.
+Local shell helpers
 are not repository prerequisites. After changing native Python bindings,
 generate the committed stubs from the rebuilt module via `refresh_python_stubs`
 and run `check_python_stubs`; do not maintain binding stubs by hand.
 
-Manual checks: open legacy 3DGS and 3DGUT projects/checkpoints, save and reopen
+Manual checks: open valid 3DGS and 3DGUT projects/checkpoints, save and reopen
 them, change the backend through Python and the existing panel, and verify a
 valid training run with each backend. Exercise CLI selection, config overrides,
 viewer startup, and invalid aliases. Confirm that changing next-run settings
--- a/docs/docs/development/training-backend-contract.md
+++ b/docs/docs/development/training-backend-contract.md
@@ -1,0 +1,118 @@
+# Training backend identity and compatibility
+
+Training strategy and raster backend are independent choices. `3dgs` is the
+default training backend; `3dgut` selects Gaussian Unscented Transform rendering.
+No additional backend alias is introduced for 3DGS.
+
+`core/training_backend.hpp` owns the backend identifiers, labels, descriptions,
+viewer mapping, and high-level training capabilities. Its descriptions provide
+technical tooltip text for later UI consumers. The capability map is exhaustive
+for the high-level feature groups exposed by the training panel; a consumer does
+not need to infer support from a missing key.
+
+Each capability has one of two states:
+
+- `supported`: the backend path is established and may be presented normally;
+- `unsupported`: the combination is a known hard incompatibility and may prevent
+  training from starting.
+
+Only `unsupported` participates in central backend validation. The current
+matrix is:
+
+| Feature group | 3DGS | 3DGUT |
+| --- | --- | --- |
+| MCMC strategy | supported | supported |
+| MRNF strategy | supported | supported |
+| IGS+ strategy | supported | unsupported |
+| Undistortion | supported | supported |
+| Mip Filter | supported | unsupported |
+| Depth supervision | supported | unsupported |
+| Normal supervision | supported | unsupported |
+| Masking | supported | supported |
+| Segmentation | supported | supported |
+| Background modes | supported | supported |
+| Background Improvements | supported | supported |
+| Exposure correction | supported | supported |
+| Bilateral grid | supported | supported |
+| PPISP | supported | supported |
+| Sparsity | supported | supported |
+
+This matrix describes backend compatibility, not complete option availability.
+Detailed numeric controls, strategy-specific applicability, mutually exclusive
+option groups, and dataset requirements still belong to the property registry
+and the later resolved UI model. For example, Background Improvements is an MRNF
+option even though the backend matrix also reports its backend support state.
+Localized conflict reasons continue to come from the existing backend-conflict
+result instead of duplicating UI text in the descriptor.
+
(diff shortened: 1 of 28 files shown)


src/core/include/core/training_backend.hpp
```diff
@@ -0,0 +1,51 @@
...
+        std::string_view label;
+        std::string_view viewer_name;
+        TrainingBackendCapabilities capabilities;
+    };
+
+    // Only verified training restrictions belong here. Mask/appearance availability
+    // must not be inferred from the supervision capabilities.
+    inline constexpr std::array kTrainingBackends{
+        TrainingBackendDescriptor{RasterBackendId::FastGS, "fastgs", "FastGS", "3dgs", {true, true, true, true, true}},
+        TrainingBackendDescriptor{RasterBackendId::ThreeDGUT, "3dgut", "3DGUT", "3dgut", {false, false, false, false, false}},
```
MrNeRF: Undistort is supported on the current #2047 head (and on master since #2054). This row and the docs need `undistort: true`; `TrainingBackendConflict::Undistort` no longer exists there, so `backend_conflict()` conflicts on rebase.

docs/docs/development/training-backend-contract.md
```diff
@@ -0,0 +1,83 @@
...
+
+Training strategy and raster backend are independent choices. `fastgs` is the
+default training backend; `3dgut` selects Gaussian Unscented Transform rendering.
+The viewer calls its standard rasterizer `3dgs`: that viewer identifier is not a
+third training backend.
+
+`core/training_backend.hpp` owns the backend identifiers, labels, viewer mapping,
+and verified training capabilities. Central parameter validation consumes these
+capabilities. The current capability set covers IGS+, undistortion, Mip Filter,
+depth supervision, and normal supervision. All five are unsupported with 3DGUT.
```
MrNeRF: Out of date: 3DGUT trains on undistorted images now. Four remain unsupported (IGS+, Mip Filter, depth, normal).

src/python/stubs/lichtfeld/__init__.pyi
```diff
@@ -430,6 +430,10 @@ def trainer_eta_seconds() -> float:
 def trainer_strategy_type() -> str:
     """Get training strategy type (mcmc, default, etc.)"""
 
+def training_backends() -> list[dict[str, object]]:
```
MrNeRF: Hand-edited: stubgen emits `-> list` with the binding docstring, places this after `BackgroundMode`, and writes `raster_backend(self, arg: str, /)` and `backend_capabilities(self) -> dict: ...`. The default build fails at `python_stub_check` on this. Run `refresh_python_stubs` and commit the output.

src/core/parameters.cpp
```diff
@@ -163,6 +163,25 @@ namespace lfs::core {
...
+                    if (!backend)
+                        throw std::invalid_argument("Unknown training raster_backend: " + name);
+                    if (json.contains("gut")) {
+                        const bool legacy_gut = json.at("gut").get<bool>();
+                        const auto legacy_backend = legacy_gut
+                                                        ? RasterBackendId::ThreeDGUT
+                                                        : RasterBackendId::FastGS;
+                        if (*backend != legacy_backend) {
+                            LOG_WARN(
+                                "Conflicting raster_backend '{}' and legacy gut={}; using the legacy value for backward compatibility",
```
MrNeRF: Legacy-wins is justified for PRMS: `set_snapshot` goes through `merge_known`, so an older build really can flip `gut` and leave a stale `raster_backend` behind. For a `--config` file it means editing only `raster_backend` in a file that also carries `gut` silently keeps the old backend (warn only, I verified at default log level). Either say so in the contract doc, or make a disagreement an error for config files and keep the lenient rule for chapters/checkpoints.

src/python/lfs/py_params.cpp
```diff
@@ -991,6 +1024,20 @@ namespace lfs::python {
                 },
                 nb::arg("image_count"),
                 "Auto-scale steps for all strategies based on image count")
+            .def_prop_rw(
+                "raster_backend",
```
MrNeRF, line 1072: `rs.get("raster_backend")` returns `3dgs`/`3dgut`; feeding that into this setter raises. Same property name, different value set. See the summary, I would rename the training-side key or accept both spellings.

docs/docs/development/training-backend-contract.md
```diff
@@ -0,0 +1,83 @@
...
+either boolean value or assume that adding a descriptor installs a rasterizer.
+
+## Verification
+
+After updating the native binaries, run from the repository in the normal
+PowerShell development environment:
+
+```powershell
+lfsdev
+.\build\tests\lichtfeld_tests.exe --gtest_filter="TrainingParametersTest.BackendIdentityCompatibility:TrainingParametersTest.StoredBackendConflictPreservesSettingsButStillRejectsInvalidNumbers:ArgumentParserTest.ExplicitBackendSelectionAndLegacyAlias:ArgumentParserTest.BackendCliOverrideReplacesConfigAliasesTogether:ArgumentParserTest.ViewerBackendSelectionSurvivesParameterDefaults:ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility:ParameterManagerTest.PendingProjectRestoreChangesOnlyRoleQualifiedManagerState" --gtest_color=no
```
MrNeRF: `ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility` lives in `lichtfeld_format_tests`, so this filter runs six of the seven names and still exits 0. `lfsdev`/`lfspytest` are not defined anywhere in the repo; a plain Linux invocation would help.

tests/test_project_chapters.cpp
```diff
@@ -414,6 +414,38 @@ namespace {
...
+            for (const auto* role : {"session", "current"})
+                legacy_json["presets"][strategy][role].erase("raster_backend");
+        }
+        auto legacy_chapter = ParametersChapter::parse(legacy_json.dump());
+        ASSERT_TRUE(legacy_chapter) << lfs::format_for_developer(legacy_chapter.error());
+        auto legacy_snapshot = legacy_chapter->snapshot();
+        ASSERT_TRUE(legacy_snapshot);
+        EXPECT_TRUE(legacy_snapshot->mrnf_current.gut);
+        EXPECT_FALSE(legacy_snapshot->mrnf_session.gut);
+
```
MrNeRF: Stray blank line.

MrNeRF (changes requested):
Thanks Francesco, the shape is right: a named identity with `gut` kept as the storage, and the capability table feeding the conflict checks is where I want this to go. I built the branch, ran the parameter/CLI/checkpoint/project gtests (204 + 108 green), the property/panel pytests (126 green), and the CLI end to end: both backends train and eval, `--gut --raster-backend fastgs` fails in either order, an unknown name fails from CLI, JSON and Python, a 3DGUT project resumed with `--raster-backend fastgs` continues as FastGS and without the flag stays 3DGUT. All of that matches the description. Two things block it, one needs a decision:

1. **The branch sits on the old #2047.** #2047 now allows 3DGUT with undistortion (`d41cc897f`, after #2054 landed), and `TrainingBackendConflict::Undistort` is gone. This PR still hardcodes undistort as unsupported in `training_backend.hpp` and in the docs, so on top of the current #2047 it conflicts in `parameters.cpp` and `tests/test_argument_parser.cpp` and references a removed enum value. I did the rebase locally to check the rest survives: with the 3DGUT undistort capability set to true it builds, the gtests/pytests stay green, and `--raster-backend 3dgut --undistort` trains. Please rebase onto the current #2047 head and flip that capability.

2. **The committed Python stubs are hand-written.** `src/python/stubs/lichtfeld/__init__.pyi` differs from what stubgen emits in docstrings, annotations and placement of `training_backends()`. A default build (`BUILD_PYTHON_STUBS=ON`) stops at `python_stub_check` with "Committed stubs are out of date"; CI only passes because every workflow sets it OFF. `cmake --build build --target refresh_python_stubs` and commit the result; if you want the richer docstrings, put them in the bindings.

3. **Two `raster_backend` vocabularies.** The viewer already has a `raster_backend` setting with `3dgs`/`3dgut` (render settings, VIEW chapter, MCP schema, rendering panel). This PR adds a training `raster_backend` with `fastgs`/`3dgut`. Same key, disjoint values: `3dgs` is rejected as a training backend, `fastgs` is not a viewer backend, and the panel already translates between the two by hand. Before the selector PR consumes this, I would rather pick one name for the training side (`training_backend`?) or accept both spellings, than ship two keys called the same thing.

Smaller items inline. On scope: "training and evaluation routing use the identity" holds for `trainer.cpp`/`metrics.cpp`, but the viewer hand-off (`application.cpp:1341`, `visualizer_impl.cpp:289`), the status-bar label and `training_panel.py:1674/1740` still read the raw bool; fine today, worth a sentence in the doc so the third-backend story is honest. Tests: no case for the reversed CLI order, for `3dgs` as a training name, or for a non-string JSON value other than null, and the chapter round trip would pass even if `raster_backend` were never written.

francescofugazzi:
I rebased the branch onto the current #2047 head
and addressed the review points.

3DGUT now advertises Undistort as supported, while IGS+, Mip Filter, Depth, and
Normal remain unsupported. The training identity now uses `3dgs` / `3dgut`, the
same vocabulary as the viewer, so the separate FastGS spelling is gone.

The Python stubs were regenerated from the rebuilt module and the committed
output now matches stubgen.

I also added coverage for reversed CLI argument order, explicit `3dgs`, 3DGUT
with Undistort, non-string JSON values, config alias precedence, and the actual
serialized backend fields in the project chapter. The documentation now explains
the config precedence rule, identifies the remaining `gut` consumers, and runs
the chapter test from `lichtfeld_format_tests`.

The requested native and Python tests pass, as do the generated-stub, error-debt,
and I/O checks.

src/core/parameters.cpp
```diff
@@ -163,6 +163,25 @@ namespace lfs::core {
...
+                    backend = parse_training_backend(name);
+                    if (!backend)
+                        throw std::invalid_argument("Unknown training raster_backend: " + name);
+                    if (json.contains("gut")) {
+                        const bool legacy_gut = json.at("gut").get<bool>();
+                        const auto legacy_backend = legacy_gut
+                                                        ? RasterBackendId::ThreeDGUT
+                                                        : RasterBackendId::ThreeDGS;
+                        if (*backend != legacy_backend) {
+                            LOG_WARN(
```
MrNeRF: No legacy writers exist (v0.5.3 predates `.licht`), so there is nothing for legacy-wins to protect. Make this an error; a config author who edited only `raster_backend` should not silently get the old backend.

src/python/lfs/py_params.cpp
```diff
@@ -280,6 +305,15 @@ namespace lfs::python {
     }
 
     void PyOptimizationParams::set(const std::string& prop_id, nb::object value) {
+        if (prop_id == "raster_backend") {
+            const auto name = nb::cast<std::string>(value);
```
MrNeRF, line 311: `nb::cast<std::string>` on a non-string surfaces as `RuntimeError: std::bad_cast`. Check the type and raise the same clear `ValueError` as for an unknown name.

docs/docs/development/training-backend-contract.md
```diff
@@ -0,0 +1,144 @@
...
+  overrides either valid config after loading it.
+- `--raster-backend 3dgs|3dgut` is additive. `--gut` remains an alias for 3DGUT.
+  `--gut --raster-backend 3dgs` is an error, independent of argument order.
+- Explicit CLI selection overrides a valid configuration's backend. Its captured
+  overrides contain both aliases so subsequent checkpoint restoration cannot
+  combine a stale config alias with the CLI choice. Invalid configurations still
+  fail their existing load-time validation before CLI overrides are applied.
+- Checkpoint parameter JSON and `.licht` parameter presets use the same adapters.
+  The binary checkpoint format and `.licht` chapter schema are unchanged.
+- A stored `.licht` preset may retain an unsupported 3DGUT option so legacy
```
MrNeRF: Reads as if every unsupported combination survives loading. IGS+ with 3DGUT is rejected in both storage and runtime validation (parameters.cpp).

MrNeRF (commented):
Round 2 on 3f72859cc, on top of the current #2047 head. All seven points from round 1 are fixed. Built and verified live: with breakpoints on the two forward kernels, `--gut` and `--raster-backend 3dgut` hit the gsplat path and `--raster-backend 3dgs` hits FastGS, on pinhole and on fisheye with undistortion; conflicting flags are refused in both orders; unknown names and non-string JSON are refused; a config with both fields disagreeing warns at the default level and takes `gut`; an explicit CLI backend wins over a config; a checkpoint resumed with `--raster-backend 3dgs` stays 3DGS on the next flag-less resume and keeps its Mip Filter; old projects open with the right name and gain the field on save; the Python property, `get`/`set` and the legacy flag stay in sync both ways, the viewer string round-trips, notifications fire. Stubs match the generator.

Fine to merge after #2047, with two changes:

- There are no legacy writers (v0.5.3 predates `.licht`), so the "legacy `gut` wins" rule has no case to serve. Make a disagreement between `gut` and `raster_backend` an error.
- `p.set('raster_backend', 5)` raises a bare `RuntimeError: std::bad_cast`; a bad name raises a clear `ValueError`. Same treatment for both.

Doc nits: the storage-validation paragraph reads as if every unsupported combination survives loading, but IGS+ with 3DGUT does not; the plain `python3 -m pytest` line only works with the build's interpreter. Observation, not counted against this PR: after a proven 3DGUT run the status bar reads "Complete (MCMC/3DGS)" from the viewer setting you already list as a remaining `gut` consumer.

docs/docs/development/training-backend-contract.md
```diff
@@ -0,0 +1,144 @@
...
+.\build\tests\lichtfeld_tests.exe '--gtest_filter=TrainingParametersTest.*:ArgumentParserTest.*Backend*:ArgumentParserTest.Gut*:TrainerConstructionTest.*:CheckpointParamsJsonTest.*:ParameterManagerTest.PendingProjectRestoreChangesOnlyRoleQualifiedManagerState' --gtest_color=no
+.\build\tests\lichtfeld_format_tests.exe '--gtest_filter=ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility' --gtest_color=no
+python -m pytest tests/python/test_property_system.py -q -p no:cacheprovider
+```
+
+On Linux, with the build's Python module and libraries available:
+
+```sh
+./build/tests/lichtfeld_tests --gtest_filter='TrainingParametersTest.*:ArgumentParserTest.*Backend*:ArgumentParserTest.Gut*:TrainerConstructionTest.*:CheckpointParamsJsonTest.*:ParameterManagerTest.PendingProjectRestoreChangesOnlyRoleQualifiedManagerState'
+./build/tests/lichtfeld_format_tests --gtest_filter='ProjectChapterTest.TrainingBackendIdentityRoundTripAndCompatibility'
```
MrNeRF, line 129: Only works with the build's interpreter (system `python3` has no pytest and no module). Point at `build/vcpkg_installed/x64-linux/tools/python3/python3.12` or say so.

francescofugazzi:
Addressed the second review round and aligned this branch with the latest local changes from #2047.

- Disagreeing gut/raster_backend JSON fields now produce an error instead of silently choosing gut.
- Non-string values passed to Python set("raster_backend", ...) produce an actionable ValueError without changing the selection. None retains the existing binding-level TypeError.
- Removed the older-writer precedence and storage-validation claims from the documentation.
- Updated Python test instructions to identify the build-matching interpreter.
- Also addressed the status-bar observation: the label now uses the active trainer's backend or the saved checkpoint's backend before trainer restoration. Missing identity is omitted rather than displayed as 3DGS.

Added regression coverage for these cases, including backend-label precedence and checkpoint identity before trainer restoration.

The requested native tests, including the additional status-bar tests, passed after rebuilding. The extra Python property/panel selection passed 121 tests, and committed stubs match generated output.

MrNeRF (approved):
Round 3, head 998ad4051 (stacked on #2047 749ce25b4), built and verified live on Linux.

All round-2 points are fixed and hold up at runtime:
- Disagreeing `gut` / `raster_backend` in JSON is rejected with the "Conflicting raster_backend" error; `--gut --raster-backend 3dgs` is rejected in both orders; `--raster-backend 3dgut` alone trains and hits the 3DGUT forward (checked with gdb breakpoints on both rasterizer entry points; plain runs hit the fast path).
- `p.set('raster_backend', 5)` now raises ValueError naming raster_backend and leaves the selection unchanged; both spellings round-trip.
- Status bar: a completed 3DGUT project shows "Complete (MCMC/3DGUT)" and keeps it when the viewer backend and the next-run preset are changed and after a restore, resume and stop. An explicit `--resume ... --raster-backend 3dgs` is accepted and the saved run then reads 3DGS, which matches the override contract.
- Docs use the build interpreter. Fresh stub generation matches the committed stubs.

Tests: 58 C++ (new StatusBarBackendTest and the checkpoint-before-restore lifecycle test listed and executed), 1 format, 121 python, all green.

Nothing new to flag in this PR's own diff. It has to wait for #2047.

MrNeRF:
needs a sync but can be merged. Thank  you!

francescofugazzi:
Rebased the branch onto the current `dev` after #2047 was merged and force-updated the PR branch.

`git range-diff` confirmed that the three reviewed backend-identity commits were preserved unchanged while removing the old stacked #2047 history.

The Windows rebuild also exposed two test integration problems inherited from the updated `dev` base:

- `CrashHandlerTest.HandledGpuFailureIsSavedOutsideRotatingLogs` placed platform-specific preprocessor directives inside an `EXPECT_EXIT` macro argument. This is undefined preprocessing behavior and was rejected by conforming MSVC preprocessing. The process-ID selection now lives in a small platform-specific helper outside the macro.
- The new Vulkan terminal-state and VkSplat scratch-release tests referenced implementation members through the Windows shared `lfs_visualizer` library, but the required symbols were absent from its import library. Only the eight members exercised by those tests are now exported; the implementation classes were not exported wholesale and the tests were not disabled.

These corrections improve Windows test and build portability without changing runtime training or rendering behavior.

The requested backend identity, project-format and Python tests passed after rebuilding. Additional Windows coverage also passed:

- 6 `CrashHandlerTest` tests
- 1 `VulkanContextTerminalTest`
- 1 `VksplatScratchReleaseTest`

Error-debt, I/O discipline, locale completeness, hardcoded UI text and localization contract checks pass. The three additional files also pass the Visual Studio clang-format 19.1.5 check.

PR #2057, opened 2026-09-07
Fix normal loss degrading large scenes late in training (wrong gradient index)

MrNeRF:
Training with the normal loss on large scenes looked fine for a long time and then slowly turned into fog, while the same dataset without normals trained cleanly (Discord report, 16M splats, SH 0, MRNF).

Cause: in the fastgs backward the per-Gaussian normal gradient helper only has `n_visible` rows and is filled at the compact visible index, but the preprocess backward read it at the full-model index. Every other helper in that kernel already uses the compact index. So as soon as an invisible Gaussian sits ahead of a visible one, the rotation gradient from normal supervision comes from another Gaussian, and Gaussians past the visible count read unrelated arena memory. With millions of splats and most of them invisible per view, that is a small garbage rotation update on almost every visible Gaussian at every step, which is exactly the slow late decay people saw.

Fix is a one-token index change. A regression test puts two culled primitives ahead of the visible one so the two indices differ: it fails on the old kernel on all four quaternion components and passes with the fix. The existing finite-difference test only had a single splat, where both indices are 0. fastgs, normal-loss and Morton suites pass.

A full before/after on the reported drone dataset (pier90, 3.7k images, 16M cap, checkpoints every 2500 iterations) is running and will be posted here.

### DIFF
--- a/src/training/rasterization/fastgs/rasterization/include/kernels_backward.cuh
+++ b/src/training/rasterization/fastgs/rasterization/include/kernels_backward.cuh
@@ -383,7 +383,7 @@ namespace fast_lfs::rasterization::kernels::backward {
                 2.0f * (rotation_scaled.m13 * dL_dcov3d.m13 + rotation_scaled.m23 * dL_dcov3d.m23 + rotation_scaled.m33 * dL_dcov3d.m33)};
 
             if (grad_normal != nullptr) {
-                const float3 g_cam = grad_normal[primitive_idx];
+                const float3 g_cam = grad_normal[work_idx];
                 const float3 g_world = make_float3(
                     w2c_r1.x * g_cam.x + w2c_r2.x * g_cam.y + w2c_r3.x * g_cam.z,
                     w2c_r1.y * g_cam.x + w2c_r2.y * g_cam.y + w2c_r3.y * g_cam.z,
--- a/tests/test_fastgs_kernels.cpp
+++ b/tests/test_fastgs_kernels.cpp
@@ -554,13 +554,14 @@ namespace {
         std::vector<float> t_data{0.0f, 0.0f, 4.0f};
 
         SplatData make_splat(const std::vector<float>& rotation_data) const {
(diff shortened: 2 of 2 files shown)


MrNeRF:
Before/after on the reported dataset (pier90: 3,692 DJI frames at 2009x1493, MRNF, SH 0, 16M cap, normal loss with the current defaults, MoGe-2 auto-generated normal maps, 30k iterations, same config file for both runs).

**Unfixed master (5c72b16f):** MRNF starts with 7.6M Gaussians from the COLMAP points, so almost every Gaussian's model index is beyond the per-view visible count. Normal supervision switches on at iteration 2400 (8% of 30k) and the run died at iteration 2403 with `cudaErrorIllegalAddress`: the wrong-index read goes past the helper buffer and here hit unmapped memory rather than the stale arena bytes that produce the slow fog on other machines.

**With the fix:** the same run completed all 30k iterations with zero error lines. Held-out eval (every 100th image), normal error is rendered vs. prior:

| iter | PSNR | SSIM | LPIPS | Gaussians | normal err |
|---:|---:|---:|---:|---:|---:|
| 5000 | 17.47 | 0.528 | 0.587 | 15.5M | 28.0° |
| 10000 | 17.08 | 0.542 | 0.558 | 15.5M | 28.7° |
| 15000 | 17.64 | 0.568 | 0.531 | 16.0M | 28.8° |
| 20000 | 18.39 | 0.596 | 0.505 | 16.0M | 27.7° |
| 25000 | 19.13 | 0.611 | 0.487 | 16.0M | 25.7° |
| 30000 | 19.69 | 0.630 | 0.468 | 16.0M | 22.8° |

Timelapse renders of three training views every 500 iterations: Laplacian variance and mean gradient magnitude keep rising through the last 5k iterations (+12% to +20% from 25k to 30k), i.e. the renders get sharper, not foggier, in exactly the phase where the report showed the decay. Resumable checkpoints were written every 2500 iterations into the run's project.licht for later experiments.

PR #2058, opened 2026-09-07
Per-panel depth windows: split viewports get independent depth windows, panel-addressed cameras, and a focused-panel depth toolbar

alexmgee:
Implements step 3 of the three-PR sequence proposed in #1845, following #1860 and #1932.

### The problem

Steps 1 and 2 gave the depth window a stated screen footprint, per-axis sizing and direct
manipulation, but there was still only one window. In a split view both panels showed the same
rectangle over different cameras, a stroke in the right panel filtered by whatever the shared
window said, undo restored state into whichever panel happened to be focused, and the toolbar
could only talk about "the" window. Toolbar camera actions did not consistently address the intended panel: Home used the
primary-camera command path, while Eye followed the focused viewport.

This PR gives each panel of a split view its own depth window. A sync toggle restores the old
shared behaviour, strokes filter by the panel they were made in, undo entries know which panel
they belong to, the toolbar shows and controls the focused panel, and Home/Eye can address a
specific panel. Outside independent split view, depth edits use the existing global window.
The GT retention and explicit-panel refusal rules are described below.

### Where this is going

This is the last of the three steps:

1. #1860 established the screen-space window and projection contract.
2. #1932 added per-axis size and direct manipulation.
3. **This PR** adds per-panel windows in split view with a sync toggle, panel-addressed
   cameras, and the focused-panel toolbar.

### What is in this PR

- Per-panel depth-window state owned by the rendering manager: two slots, a sync flag, and one
  shared transition helper at every mode-change site so entering or leaving a split can never
  leave a panel unseeded.
- Press ownership decided at the input event. Mouse DOWNs retained for dispatch carry
  GUI/viewport ownership in the ordered stream introduced upstream in #1966/#1993.
  Available, still-owned UPs follow their DOWN's owner, including releases outside the
  original control; earlier event-loss limits are disclosed below.
- Panel-addressed camera actions (Home/Eye), exposed through RML, Python and the input
  controller; `main` and the legacy panel-less form keep their old meaning.
- Python: optional `panel=` on the depth-window and camera surface, a focused-panel getter,
  regenerated stubs.
- The depth toolbar shows the focused panel (chip + controls), with new icons, per-panel Size
  references, and all ten locales extended (+9 keys, every locale complete).
- Both panels' windows survive GT comparison untouched (see below).

### Design notes

Press ownership rides upstream's event stream rather than a parallel record. This consumer
preserves retained canonical records and delivers owned events in their original order;
it may skip events it does not own. Earlier frame-loss paths remain disclosed below.
Two deliberate divergences from the old overlay behaviour follow from that rule:

- The aggregate button fallback runs only when the frame carried no real button events at
  all. Running it merely because nothing was *delivered* would fabricate a press at the
  cursor after every real event had been rejected as unowned.
- An UP whose DOWN this overlay did not deliver is rejected, even while another button's drag
  capture is active. An unmatched release must not borrow ownership.

Focus follows presses in order. If a frame carries several left presses, each applies its own
focus decision, so a later press on toolbar chrome cannot erase the focus an earlier viewport
press legitimately set.

Sync fans out on write, so a synced pair stays byte-identical to the single-window behaviour.
During GT comparison, a parked panel pair makes `set_depth_window_sync` silently refuse
both ON and OFF changes, extending the existing drag-ownership refusal. It returns the
actual flag. During the retained Disabled interval, an actual sync change first discards
restoration eligibility and then applies; a same-value request preserves it.

Size references are per-panel and re-base on the panel a commit names. Valid GT excursions
preserve both references, including when polling observes only part of the mode sequence.
Destructive lineage changes tell the Python consumer to recover its references from the
current windows; they can lose the earlier percentage baseline.

The drag operator no longer opens an operator-wide history transaction. A changed,
successfully committed drag pushes its own panel snapshot entry; cancelled and unchanged
drags add no entry. Sync changes retain their separate undo entry. An enclosing transaction
can still collect a direct push.

`reset_camera()` and `reset_camera(panel=None)` retain the legacy primary-camera target.
Use `reset_camera(panel="main")` for the focused panel. Both `focus_selection()` and
`focus_selection(panel="main")` target the focused camera. Explicit left/right camera
actions do not change focus.

The state identities, drag rollback, lock ordering, thread entry points and consumer
reconciliation are documented in `docs/docs/development/depth-window-state.md`.

Removed unused `restoreDepthWindowForPanelIfEpoch`, duplicate generation updates,
temporary lineage fields/helpers and a redundant epoch read. Ownership, rollback and
thread-ordering mechanisms remain; broader redesigns were investigated separately.

Off-focus Home/Eye moves the addressed camera without re-anchoring the shared depth box.
A blank Hide result after off-focus Eye was reported during review and has not been
reproduced here. In the Vulkan containment path inspected, visibility uses the displayed
camera and the panel screen window and depth range; it does not use the shared transform
or XY extents for that test. The reported blank result remains undiagnosed.

### GT comparison

GT suspends depth-window rendering, selection filtering and drag creation. Entering from
independent split parks both windows. Direct return restores them, and the GT on, GT off,
then independent-split sequence also restores the original pair. Repeated GT entry during
that retained Disabled interval keeps the original pair; the next independent entry consumes
it. Focus and camera-toggle behavior retain their existing rules.

Project/scene reset, another comparison mode, changed global window geometry or a changed
global drag commit while Disabled, and an actual Disabled sync change discard retention.
Geometry equality compares all six fields exactly after the writing path's existing
normalization; drag commit compares with its pre-drag baseline. A handle released within
0.001 screen pixel of its press restores the exact pre-drag state without an undo entry.
Enable/viz-only changes, equal normalized writes, same-value sync, and
cancelled/subthreshold drags preserve it.
Global GT-time writes keep the existing compatibility behavior and preserve the parked pair.

An explicit `panel=` window request while a pair is parked in GT is refused atomically:
`enabled=True` raises `RuntimeError`; `enabled=False` returns without changing geometry,
the global enabled flag, projection, generation or history. This intentionally differs from
the sync API's silent refusal and actual-flag return. `panel=None` retains the global route,
and GT entered without a parked independent pair remains writable. Explicit panel writes
during retained Disabled act as global edits under the invalidation rules above.

Boundary tests cover drag rollback on entry, combined settings writes entering GT,
Disabled sync invalidation before refusal, observed/coalesced Size-reference continuity,
and the compiled binding's full-request refusal.

### Defects found and fixed before submission

**The GT round trip destroyed differing panels.** The generic mode-transition code stamped
the focused window over both slots on GT entry and re-seeded both on return, so two different
windows never survived a round trip. Every targeted suite was green when this was found by
review against the contract above; the existing GT test only looked at state after the
destructive step. Fixed before submission with direct GT-to-independent park/restore,
a round-trip test and boundary tests. Each of those five fix predicates was mutation-checked: forcing
it false failed exactly its matching test. This review round extends retention through
Disabled and tests that journey separately; those earlier mutations were not rerun.

**Home always addressed the primary camera.** The panel-addressed path fixes the pre-existing
defect where Home from the toolbar moved the primary camera regardless of the panel meant.

**The Size/X/Y row overflowed.** The row previously overflowed its container at whole-dp
widths 701-940 (the base had the same defect class at 651-888dp). The control now hides
through 960dp instead of stacking; per-viewport controls sit 12dp lower from 961-1480dp.

**A first-apply tooltip repaint latch.** The viewport overlay carries the first apply's
change edge into `render()`, fixing five reproducible cases where a tooltip's first
appearance was not repainted.

**A horizontal-wheel loss avoided.** Re-expressing press ownership on the upstream stream
deletes a dispatcher that would have discarded horizontal wheel input; #1966's two-axis
wheel delivery is preserved.

### Interface changes and migrations

1. **`panel=` addressing.** Depth-window and camera calls accept keyword-only `panel=`
   with `"left"`, `"right"` or `"main"`. Omitted or `None` retains the legacy route.
   The sync getter/setter manage one global flag and take no panel. **Migration:** none.
2. **Depth-window call compatibility.** The getter retains its seven-value tuple and the
   setter retains its positional parameters; the seventh remains `scale_y`. Select a
   panel by keyword, for example `set_depth_filter_window(True, panel="right")`.
3. **GT refusal.** Explicit-panel window requests follow the enable-raises/disable-noops
   rule above. Sync changes are silently refused during a drag or a parked GT session;
   an actual Disabled sync change invalidates retention and applies. **Migration:**
   handle `RuntimeError` for refused enable requests and use the sync setter's returned
   flag as authoritative. The binding docs and generated stubs describe these rules.
4. **Middle-button policy retired.** The old dispatcher intentionally never delivered the
   middle button; the canonical producer and generic replay now support it. The bit-only
   aggregate fallback remains left/right. **Migration:** none.
5. **Undo entries are panel-addressed** and carry an epoch; a stale entry is inert and
   renders an expired label (three new history keys, localised). **Migration:** none.

### Verification

After merging `dev` at `ae08a9ab3`, the Windows rebuild passes **370 targeted C++** and **302 Python controller**
cases, including both reported regressions. Source and loaded-library identities match;
formatting, whitespace, fresh stubs, locale/UI checks and 36 preflight tests pass.
Earlier RmlUi package/link checks remain recorded separately.

Before the dev merge, live duplicate-DOWN injection on `8f93ba209` cleared the slider press; subsequent
unheld movement left Near fixed. Manual slider release, GT round trip and
Sync/Undo/Redo across a focus change also passed. Exact handle return is covered
by automated tests; the latest manual attempt could not assess the subpixel tolerance.

Built on Windows with `dev` at `ae08a9ab3` merged, Release, Ninja. C++ is compiled
with MSVC toolset **14.51.36231**; CUDA **12.9 (V12.9.41)** is compiled by nvcc using MSVC
toolset **14.44.35207** as its host compiler, pinned by our local build scripts, not the
project's CMake, with `--allow-unsupported-compiler` permitting the pairing. CUDA
architectures: `native`.

To separate this PR's effect from upstream motion, a pristine `a61330d26` was built in a
second worktree with the same toolchain, and every comparison below is against that
baseline's results by test name, not by count.

Validation recorded through c5b8e5664 rebuilt the test executable, visualizer DLL and
Python binding and passed **340 targeted C++ names** and **418 Python names**, with no missing, extra,
failed or skipped names against the declared sets. All 227 original C++ and 300 original
Python names remain covered. Source and binary identities were checked before and after
the runs. The C++ gate includes 20 new compiled-overlay cases and real-binding refusal
checks; Python includes 118 added Size-reference cases.

Forced fresh stub generation and an explicit generated-versus-committed comparison both
passed. The same UI stub also reconciles existing video signature/doc drift with unchanged
bindings. Locale completeness passes for all 2255 English keys across ten locales;
the hardcoded-UI-text scan reports zero candidates. Changed C++ passes clang-format
22.1.3 and the diff whitespace check passes.

The submitted branch's earlier differential gate passed 227 targeted C++ names and 300
Python names. Its five GT source mutants were killed by matching regressions. The earlier
full fast battery had zero regressions or dropped tests against the pristine baseline
(89 additions), with the same recorded failure sets. Those full-battery and mutation
results apply to that submitted revision; they are not a new full-battery run of this round.

Earlier, two app-driven checks were run against both binaries. The scripted
no-panel A/B sequence is byte-identical between this branch and pure upstream. The step-1
parity lock is red at one of its three cameras, with the same 55 mismatches, indices and
values on pure upstream and green at the other two cameras on both binaries; that makes it a
current-upstream condition rather than something this PR introduces (documented below). The
same both-binaries differential holds for one sub-assertion of the step-2 split-write lock.

This is targeted plus differential coverage, not a claim that every repository suite was run
clean; the pre-existing upstream failures named above still fail on both binaries.

Review note: the first two commits form one logical unit. The first forward-references two
types the second defines, so the intermediate commit is intentionally not independently
buildable and should not be bisected as a green point.

### Known limitations

- Explicit-panel refusal is exercised through the compiled Python binding with a live
  manager. This does not establish active-tool GUI coverage or a complete RmlUi camera click.
- Automated pointer tests drive compiled overlay routing with real
  RmlUi button/range/text controls, six explicit blocker states, independent Left/Right
  releases and an unmatched Middle UP. No positive Middle DOWN/UP or autoscroll
  cancellation regression is claimed. Live duplicate-DOWN coverage is described above;
  suppression of synthetic clicks is checked by compiled tests, not a live click counter.
- The input repair handles observed, still-owned releases and cancels a superseded primary-button
  RmlUi press without a synthetic release or click. Cancellation uses an interim RmlUi 6.2 patch;
  middle-button autoscroll cancellation is outside its scope. Unobserved/unusable releases
  and earlier resize/export/no-consumer frame loss remain unresolved.
- Per-panel windows and the sync flag are session-only. Project restore seeds both panels
  from the global depth state present after restore and turns sync off. Existing global
  depth-filter fields remain serialized; per-axis screen size and offsets are not.
  Saving per-panel windows and the sync flag is a follow-up to this PR.
- A pre-existing Size-reference gap remains when equal-slot sync ON and GT both occur
  between polls: no geometry copy means no sync-source stamp, so the consumer can lose
  which panel supplied the reference. The displayed percentage can differ from a sequence
  where sync is observed separately. Current and pre-round code reproduce the same gap.
- Live layout, hit targets, DPI and IME are outside this round's automated GUI coverage.
  New dock-ownership tests use native components; IME checks use a composition-state
  helper. Earlier hands-on observations apply to their recorded revisions and states.
- The ten translations were not native-speaker reviewed; localized axis labels can still
  paint outside their fixed 24dp boxes (pre-existing).
- Crossing range-slider sub-elements can briefly blink a tooltip via the 90ms reshow delay;
  deliberately accepted.
- The new camera entry points keep the existing viewer-thread assumption; worker-thread
  plugin calls remain unsupported, as with the legacy `focus_selection()`.

### Upstream work already incorporated

This branch is built directly on current `dev` and is consistent with the recent input rework
(#1966, #1993, #2004, whose ordered event stream this PR's press ownership extends rather
than replaces), the per-half render sizing (#1968), the selection-service rewrite and Windows
fixes (#1992, #2007, #2009), the GT comparison rework (#1963), the toolbar/dock motion
(#1983), the viewport chrome and rail re-flow (#1956, #2012), the icon migration (#1985,
including retargeting two of our test assertions to the renamed icons), and the deferred
async loads (#2015).

### Observations outside this PR

Three upstream conditions were observed while validating against the pristine baseline, and
the evidence for each is preserved: the step-1 parity lock's mid-camera regression on current
`dev`; a stale committed `lichtfeld/ui/__init__.pyi` that makes a plain `ninja` build fail on
clean `dev`; and a TCP test family that blocks on a socket after an assertion failure instead
of failing. They can be filed as separate issues if useful.


src/visualizer/rendering/rendering_manager.cpp
```diff
@@ -815,6 +1084,612 @@ namespace lfs::vis {
...
+        }
+
+        // Dormancy is scoped to ONE GT session entered directly from
+        // independent-dual, so every OTHER boundary this helper runs for drops
+        // the park: a GT entered from a global mode can never restore a pair it
+        // did not park, a GT left for a global mode leaves that mode's single
+        // window governing, and a pair just restored above is consumed. (A
+        // transition that touches neither boundary returned at the top, and one
+        // of those cannot be leaving GT.)
+        if (!park_dormant_panels) {
```
MrNeRF: This is the one that bit me live. Split view with two different windows, press the GT button, press it again to leave, re-enter the split: both panels now show the same window. Leaving GT lands in Disabled, that drops the park here, and the split entry re-seeds both slots. The direct GT → split jump your test uses does restore the pair, but the GT button never takes that path. The values are still in the slots when GT is left, so keeping the park until the next split entry (or seeding from it) would fix this cheaply.

src/visualizer/rendering/rendering_manager.cpp
```diff
@@ -815,6 +1084,612 @@ namespace lfs::vis {
...
+        }
+        return fan_out;
+    }
+
+    // Every depth-window write in this family marks DirtyFlag::ALL, not just
+    // SELECTION: the selection-only fast path re-rasters from cached per-splat
+    // containment, so a window change that marks only SELECTION keeps stale
+    // classifications on screen until an unrelated full render. Upstream's
+    // write path (updateSettings, one-arg) marked ALL for the same reason.
+    void RenderingManager::setDepthWindowForPanel(const SplitViewPanelId panel, const DepthWindowState& state) {
```
MrNeRF: Second live one: during GT with a parked pair, `set_depth_filter_window(..., panel="right")` fans into both live slots, the getters show the new values, and the direct return then throws them away. Sync gets a refusal in exactly this state; panel writes should get the same refusal, or write into the parked pair.

src/visualizer/rendering/rendering_manager.cpp
```diff
@@ -640,14 +730,79 @@ namespace lfs::vis {
...
+                if (depth_window_projection_changed) {
+                    const auto projection_window = depthWindowFromProjection(settings_);
+                    // Every branch below is a deliberate NON-drag write, so it
+                    // supersedes the pre-drag backup AND the pin of every slot
+                    // it writes (see releaseDepthWindowBackupsLocked).
+                    if (depth_window_sync_) {
+                        panel_depth_windows_ = {projection_window, projection_window};
+                        releaseDepthWindowBackupsLocked(split_view_service_.focusedPanel(),
+                                                        /*fan_out=*/true);
+                    } else if ((dirty_flags & DirtyFlag::SELECTION) != 0) {
```
MrNeRF, line 749: `settings_` is assigned wholesale above, but the slot only follows when SELECTION is set. A stale snapshot written without that flag leaves the projection on the old values while the focused slot keeps the new ones; the selection tool then sees the generation bump and stamps the stale values into the focused slot on its next tick. `StaleProxyWithoutSelectionMaskDoesNotMoveOtherPanel` only asserts the slot, so it doesn't catch this — worth asserting the projection there too.

src/visualizer/window/window_manager.cpp
```diff
@@ -1054,6 +1056,17 @@ namespace lfs::vis {
...
             const int mouse_y = static_cast<int>(std::round(event.button.y));
             const bool titlebar_point = isTitlebarDragPoint(mouse_x, mouse_y);
+            // Press ownership is decided HERE, at the event, and carried to the
+            // GUI frame on the frame buffer. Rectangle containment computed in
+            // the frame cannot tell a viewport press from one on a GUI-owned
+            // edge that overlaps the viewport (the left-dock resize strip), and
+            // the layout it compares against may have moved in between.
+            if (event.type == SDL_EVENT_MOUSE_BUTTON_DOWN) {
+                auto* const gui = services().guiOrNull();
+                frame_input_.notePressOwner(
```
MrNeRF, line 1066: The verdict recorded here is only read by the overlay focus rule. `InputRouter::beginMouseButton` still uses `hitTestPointer`, so a press on the left-dock resize strip (the case this exists for) is GUI-owned for focus but viewport-owned for the camera: it can start an orbit and refocus the panel while the dock resizes. The router should consume the same verdict.

src/visualizer/gui/gui_manager.cpp
```diff
@@ -7315,15 +7347,57 @@ namespace lfs::vis::gui {
             if (!block_underlay_input)
                 rml_viewport_overlay_.processInput(viewport_overlay_input);
```
MrNeRF: When `block_underlay_input` skips this call, `leftPressClassifications()` still holds last frame's list, and the focus loop below pairs this frame's presses with it. A click while a modal or the context menu is up can move panel focus on a stale classification. Clearing the list on the skip path closes it.

src/visualizer/gui/rmlui/rml_pointer_dispatch.hpp
```diff
@@ -0,0 +1,126 @@
...
+            // same-button press, or from a press this host never delivered.
+            const bool deliver = event.down ? (capture_active || owns_element(event_element))
+                                            : down_delivered[slot];
+            if (!deliver) {
+                // A press this host refused supersedes whatever that button was
+                // doing before: the flag is cleared so a stale DOWN (one whose
+                // UP was eaten by a focus loss) cannot lend its delivery right
+                // to this press's release.
+                if (event.down)
+                    down_delivered[slot] = false;
```
MrNeRF, line 81: A refused DOWN clears the delivered flag but never sends RmlUi an UP. With two left DOWNs and no UP in between (your intake tests allow that stream), an owned slider press followed by a refused press leaves RmlUi armed and the real UP is then rejected — stuck slider. Sending the UP when a delivered press is superseded would keep both sides in step.

src/python/lfs/py_selection.cpp
```diff
@@ -513,17 +647,34 @@ namespace lfs::python {
...
-                                                                                                                                                                                                                "requests (enabled=True) raise RuntimeError because they cannot be applied\n"
-                                                                                                                                                                                                                "atomically; disable requests (enabled=False) are silent atomic no-ops,\n"
-                                                                                                                                                                                                                "matching the legacy calls' contract.");
+            nb::arg("enabled"), nb::arg("depth_near") = 0.0f, nb::arg("depth_far") = 100.0f, nb::arg("scale") = 0.35f, nb::arg("offset_x") = 0.0f, nb::arg("offset_y") = 0.0f, nb::arg("scale_y") = nb::none(), nb::kw_only(), nb::arg("panel") = nb::none(), "Set the screen-space selection depth window.\n"
```
MrNeRF, line 650: Small wording thing: `panel` is keyword-only here (correctly — the 7th positional is `scale_y` from #1932), but the PR text says it is "appended as a final optional argument". Worth fixing the description so nobody tries `set_depth_filter_window(..., "right")`.

src/python/lfs/module.cpp
```diff
@@ -2497,14 +2549,44 @@ NB_MODULE(lichtfeld, m) {
...
+            const auto panel_id = resolveGizmoPanelArg(*panel);
+            if (auto* const controller = lfs::vis::InputController::instance())
+                controller->resetCameraForPanel(panel_id);
+        },
+        nb::kw_only(), nb::arg("panel") = nb::none(), "Reset camera to default position and orientation.\n"
+                                                      "\n"
+                                                      "panel (keyword-only) selects which split panel's camera is reset:\n"
+                                                      "- None (default): the command path, addressed at the primary viewport\n"
+                                                      "  exactly as before.\n"
+                                                      "- 'main': the panel that currently has focus. This is NOT the same as\n"
```
MrNeRF: This asymmetry will trip people: `reset_camera()` moves the primary camera, `reset_camera(panel="main")` moves the focused one, while for `focus_selection` both mean the focused panel. I'd rather have `None` and `"main"` agree for both, or at least call the difference out loudly in the stub.

src/visualizer/operator/ops/depth_window_ops.cpp
```diff
@@ -336,7 +391,7 @@ namespace lfs::vis::op {
             .description = "Draw, resize, or move the selection depth window",
             .icon = "selection",
             .shortcut = "",
-            .flags = OperatorFlags::REGISTER | OperatorFlags::UNDO,
+            .flags = OperatorFlags::REGISTER,
```
MrNeRF, line 378: `UNDO` went away here, so the drag no longer runs inside a history transaction and the entry is pushed raw. Probably what you want given the new entries, but it isn't mentioned anywhere — a line in the description would help.

src/python/lfs/py_ui.cpp
```diff
@@ -5864,6 +5864,116 @@ namespace lfs::python {
...
+                    return false;
+                rm->setDepthWindowSync(sync);
+                // The manager silently ignores the change while a depth-window
+                // drag is in flight, so report the ACTUAL post-call state rather
+                // than the requested one.
+                return rm->getDepthWindowSync();
+            },
+            nb::arg("sync"), "Set the per-panel depth-window sync flag. Turning it on with\n"
+                             "differing panels copies the focused panel's window to the other as\n"
+                             "one undo step. The call is silently ignored while a depth-window\n"
```
MrNeRF: Docstring still only mentions the mid-drag refusal; the parked-GT refusal is the one people will actually hit from a script. Same in the stub and the plugin comment.

src/visualizer/input/input_controller.cpp
```diff
@@ -3314,10 +3348,28 @@ namespace lfs::vis {
             .emit();
     }
 
-    void InputController::publishCameraMove(Viewport* target_viewport) {
+    // The shared depth anchor is global (DepthWindowState carries no
+    // transform or x/y extents), so re-anchoring it from a panel the user is not
+    // looking through would move the focused panel's box. Guard the side effect
+    // only -- the viewport was already chosen by panelViewport(), focus-free.
+    bool InputController::shouldSkipDepthAnchorSync(
```
MrNeRF, line 3368: Confirmed the consequence live: after an off-focus Eye on the right, the right panel shows a slice cut by the left camera's anchor, and in Hide mode it goes empty. Skipping the re-anchor is the right call, but with two panels the shared anchor is now user-visible. Short of per-panel anchors, not culling with the shared box in independent-dual Hide (the per-panel screen window is already correct) would avoid the blank panel.

src/visualizer/gui/assets/icon/src/depth-dim.svg
```diff
@@ -0,0 +1,24 @@
+<!-- SPDX-FileCopyrightText: 2025 LichtFeld Studio Authors
+
+     SPDX-License-Identifier: GPL-3.0-or-later -->
+<svg xmlns="http://www.w3.org/2000/svg"
```
MrNeRF, line 4: These five icons live next to the PNGs at 24 px with hard-coded `#fff`, outside the `#1985` pipeline (`icon/src/`, `currentColor`, exported at 40 px). The next regeneration won't emit them and the PNGs won't track the SVGs. Please route them through `export_icons.py`.

src/visualizer/rendering/rendering_manager.hpp
```diff
@@ -955,10 +1186,101 @@ namespace lfs::vis {
...
+        // the gap). The acquisition is unreachable on the re-entrant chains
+        // that already hold this mutex on the SAME thread - the mode-change
+        // event sites' cancel hook and the drag release both reach
+        // updateSettings through finishLatch ->
+        // SelectionTool::setDepthWindowDragInProgress ->
+        // applySelectionFilterSettings, which builds its RenderSettings from a
+        // fresh getSettings() and therefore always carries the CURRENT
+        // split_view_mode. An equal-mode write never touches this mutex, so
+        // those chains cannot self-deadlock on a non-recursive mutex.
+        mutable std::mutex depth_window_transition_mutex_;
```
MrNeRF, line 1125: This is where the complexity worry from the summary lives. A second mutex whose only job is to keep a drag's release sequence away from a mode change, acquired by releasing and re-taking the settings lock and re-checking the mode, is the kind of thing that is correct today and fragile tomorrow. If the release sequence and the mode changes both ran on the viewer thread in one place, none of this would be needed — worth a look before this lands.

MrNeRF (changes requested):
@alexmgee — built this on Linux on top of current `dev` (merge is clean) and spent a good while in it: two panels with different windows, the sync toggle, drags, Home/Eye on each side, GT comparison, undo, and the Python surface over MCP. The core promise holds: each panel really has its own window, the chip follows the panel you click, chrome clicks don't steal focus, a drag in the unfocused panel edits only that panel, and undo puts things back. Your targeted C++ and Python suites are green here too, the stubs match the generator, and the locale check passes.

I found two things that break the round trip in practice and a few input holes; those are inline. Two bigger notes that don't fit a single line:

- This is a lot of machinery for two slots: ownership tokens, a lineage record with generation and kind, a second mutex with a release-and-retake loop in `updateSettings`, a parked snapshot, and a 100 ms poller in Python reconciling "lineage deltas". The comments alone are hundreds of lines and reference a plan file that isn't in the repo. Whoever touches this next has to hold all of it in their head. I'd ask for a simplification pass, or at minimum trimming the comments to the invariants.
- Nothing per-panel is saved. Save with two different windows, reopen, and the second window and the sync flag are gone. Fine as a stated limit, but it should be stated.

One thing I hit that is not yours: Hide mode culls every splat in my garden scene even inside the rectangle, and pristine `dev` does the same, so that's a separate issue.

MrNeRF:
An additional request: Could you shorten the comments, please? They are very verbose and the code becomes hard to follow.

alexmgee:
Sorry about the comments, I'll do better about proofreading and shortening them. 

Also currently working to try to simplify and trim things overall as you requested, my apologies.

MrNeRF:
@alexmgee Just ping me when it is done. Sorry about closing this pr. This was an accident.

alexmgee:
@MrNeRF No worries. I shortened the comments across the depth-window code and tests, and kept the invariants and examples beside the code. State machinery and toolbar behavior are unchanged. All 340 C++ and 418 Python regression tests pass.

src/visualizer/rendering/rendering_manager.cpp
alexmgee: Fixed. Both windows now survive turning GT on, turning it off and returning to split view, including repeated toggles. Resetting the scene or project, switching to another comparison mode, or changing geometry or sync while back in single view discards the saved pair. No-op edits, visibility changes and cancelled drags leave it intact.
Tests cover the toggle commands and toolbar Size references, including changes between toolbar updates. They don’t simulate clicks in the live GUI.
One separate, pre-existing Size-percentage issue remains if you enable sync on identical windows and enter GT between toolbar updates, which is documented in the PR.

src/visualizer/rendering/rendering_manager.cpp
alexmgee: Fixed. Panel-specific writes are now refused while GT is holding a saved pair: enabled=True raises RuntimeError, and enabled=False returns without making changes. Neither request changes the windows or global enable flag.
Tests cover both requests for left, right and main, with the filter initially on or off. Calls without a panel keep their old behavior, and panel writes still work in GT when no pair is saved. These tests use the real Python binding without an active selection tool and do not exercise the live GUI.

src/visualizer/rendering/rendering_manager.cpp
alexmgee: Fixed. With sync off in independent split view, updates without SELECTION now preserve the current window geometry, including any drag preview, so long as the view mode stays the same.
The tests now check the projection, both stored windows and the generation counter through two selection-tool updates. There’s also a separate active-drag check.

src/visualizer/window/window_manager.cpp
alexmgee: Fixed. The overlay and camera input router now use the same mouse-press decision, including the dock’s resize edge. I also checked this in the GUI by resizing the Asset Manager, including quick grabs of its edge, left both cameras still and focus on the right. Middle-button input on that edge did nothing and worked normally inside the viewport afterward.

src/visualizer/gui/gui_manager.cpp
alexmgee: The overlay now clears the list before focus handling, even when a modal or menu blocks input. New presses stay blocked, while real releases are still handled for presses it owns.
Automated tests cover all six blocking states. I checked both cases in the GUI: clicking outside the context menu dismissed it without changing focus while clicking behind the modal left it open with focus still on the right. Normal focus switching worked after closing the dialog.

src/visualizer/gui/rmlui/rml_pointer_dispatch.hpp
alexmgee: Confirmed with an injected event sequence using real RmlUi. The second rejected DOWN can leave the original press armed. This remains unresolved.
Real releases reaching the overlay are now delivered for presses it still owns, including through input blockers. I tested a synthetic UP for the remaining case, but it can click the original control. Mouse-leave alternatives can leave a drag armed or trigger a drop, so I haven’t added an unverified cancellation workaround. Superseded presses and releases lost before reaching the overlay remain disclosed.

src/python/lfs/py_selection.cpp
alexmgee: I updated the PR description to say panel is keyword-only and show panel="right". The seventh positional argument remains scale_y. It also now states that sync is a global flag with no panel argument.

src/python/lfs/module.cpp
alexmgee: I’ve moved the warning to the top of the stub and shortened the parameter descriptions. It now leads with the primary-camera default, shows panel="main" for the focused camera, and explicitly contrasts this with focus_selection(). I kept the existing default so older scripts continue targeting the same camera.

src/visualizer/operator/ops/depth_window_ops.cpp
alexmgee: Added this to the PR description, the drag no longer opens its own history transaction. A changed drag adds an undo entry when successfully committed but cancelled or unchanged drags add none. An enclosing transaction can still collect that entry.

src/python/lfs/py_ui.cpp
alexmgee: Updated the binding docstring, generated stub and plugin comment. They now explain that both ON and OFF sync changes are silently ignored while GT holds a saved pair, and that the setter returns the actual sync state.

src/visualizer/input/input_controller.cpp
alexmgee: I checked the culling change you suggested and in the Vulkan path, Hide uses each panel’s camera, screen rectangle and Near/Far range. The shared anchor transform is passed into the request, but that containment test doesn’t read it. Clearing the suggested flag would disable the panel’s Hide filtering, so I haven’t made that change.
I haven’t isolated the anchor-dependent slice you describe. My GUI test went blank because Focus Selection placed the model way beyond its Far limit. I haven't resolved your issue.

src/visualizer/gui/assets/icon/src/depth-dim.svg
alexmgee: Moved all five SVGs into icon/src/, switched them to currentColor, and regenerated the 40 px PNGs with export_icons.py. All five icon tests pass.

src/visualizer/rendering/rendering_manager.hpp
alexmgee: I tried this in an isolated prototype and the approach looks viable, but moving only the rendering handlers caused ordering problems with focus changes and background undo. It needs to cover complete commands and the scope is broader than removing the mutex and retry loop.
Would you be comfortable handling the viewer thread issue in a focused follow-up PR, using these probes as regression tests? Also found an existing edge case: a GT toggle requested from an undo callback during drag release can be dropped. Should that be fixed before merge or can be worked on separately?

MrNeRF (changes requested):
Thanks for the round-2 work, this is in much better shape. Head c5b8e5664 merged onto dev, built with tests, and checked live in the GUI.

Fixed and verified:
- The left/right pair survives the GT round trip now (natural toggle-off, direct return, resize, dock and divider gestures).
- A right-panel write while parked in GT is refused cleanly, nothing leaks into the other slot or the history.
- Stale settings desync, dock-strip press orbiting, stale focus under a modal: all good.
- `panel` keyword-only, `reset_camera` docs, icons through the SVG pipeline: good.
- Tests: 532 C++, 107 format, 1925 python pass (the one python failure is an orthographic vksplat test outside this PR), locales complete, stubs match a fresh generation.

Two things left before merge:

1. A refused second mouse DOWN still strands the RmlUi press (`rml_pointer_dispatch.hpp:71`). Press the Near slider, get a second DOWN/UP on bare viewport, then move the mouse with nothing held: Near keeps changing (6.45 to 15.0 live). The test at `test_depth_window_panels_interaction.cpp:3594` currently asserts that stranded "active" state; it should assert the opposite once the press is cancelled. Please cancel the RmlUi side without synthesizing a click.

2. A Shift+Alt handle drag that ends where it started still changes the geometry by about 1e-4, which counts as an edit, drops the retained pair and adds a "Drag Depth Window" undo entry. A drag back to its origin should be a no-op, a small tolerance is enough.

Two smaller notes:
- Per-panel windows and the sync flag are not saved in the project. Fine as a follow-up since it is documented now, just say so in the PR text.
- The comment trimming is real (about 1,000 lines), but the PR itself grew to +14,375/-410. The ownership tokens, drag backups, generations, lineage reconciliation and second mutex are all still there, and `restoreDepthWindowForPanelIfEpoch` has no caller. If any of that can go, please cut it.

alexmgee:
Feels like I have been letting this drag on, my apologies. Both remaining fixes are in this update:

1. A replacement DOWN cancels the old RmlUi press without synthesizing UP, click or drop. The regression now requires cancellation. Live duplicate-DOWN injection cleared the press, and subsequent unheld movement left Near fixed. The required RmlUi patch is included in the vcpkg overlay.

2. Returning an existing handle within 0.001 screen pixel of its start restores the exact pre-drag state. Tests verify no Undo entry or retained-pair loss, while preserving genuine small edits and newer writes.

The PR now explicitly lists saving per-panel windows and sync as follow-up work.

I removed `restoreDepthWindowForPanelIfEpoch`, duplicate generation updates, temporary lineage fields/helpers and a redundant epoch read. I didn't do further trimming/optimization because of the scope of the work involved and wanting to get this submitted. I attached an assessment below that explains what remains and the alternatives investigated/proposed by Codex.

Validation: 358 targeted C++ and 302 Python controller tests pass. Manual slider release, GT round trip and Sync/Undo/Redo across a focus change also pass. Exact handle return is covered by automated tests.

<details>
<summary>Simplification assessment</summary>

The remaining mechanisms have distinct jobs:

| Mechanism | What a replacement must preserve |
| --- | --- |
| Ownership tokens | An old drag cannot overwrite a replacement drag or newer edit. |
| Drag backups | Cancel restores the state the replacement found; Undo can restore an earlier baseline. |
| Generations | Mode epochs expire operations; projection generations notify selection of changes. These are different events. |
| Lineage reconciliation | Preserve Size references through GT/sync transitions the toolbar poller may miss. Current values alone do not describe that history. |
| Second mutex | Keep drag release and mode changes ordered, including changes arriving off the viewer thread. |

The backup distinction is observable:

```mermaid
flowchart LR
    O[Original window] -->|Drag A| P[Preview A]
    P -->|Replacement drag B| Q[Preview B]
    Q -->|Cancel B| P
    Q -->|Commit B, then Undo| O
```

One backup cannot represent both return points. Existing replacement tests require both.

**What could change next**

- **Combine preview read/write/read:** three settings-lock acquisitions become one. This changes which concurrent history operations can interleave. Define that behavior and measure the real drag path before adopting it; fewer locks do not establish fewer lines or faster frames.
- **Replace the second mutex with viewer-thread execution:** route every relevant transition and release through the existing dispatcher. Worker calls would queue and wait; shutdown can cancel queued work. Completion, reentry and shutdown behavior need validation first.
- **Remove the duplicate toolbar comparison cache:** a smaller, independent deletion, estimated at about 12 source lines. Preserve existing comparison semantics. This remains an unimplemented candidate.

The investigation supports these narrower proposals, not deleting the current safeguards without replacements. No larger redesign has yet demonstrated equivalent behavior and performance.

**Evidence:** `DepthWindowDragOperator::captureBaselineIfNeeded` and `applyRect` show the separate backups and preview calls. In `tests/test_depth_window_panels_interaction.cpp`, `CancelOfReplacementDragKeepsInvokeTimeScreenStateSynced`, `UndoOfReplacementDragNeverResurrectsReplacedPreviewSynced` and `NewerSettingsWriteBeatsCancelTeardownOnBothSlots` pin the replacement behavior. `selection_controls.py` contains the polling reconciliation; `py_ui.cpp::invoke_on_viewer` exposes worker dispatch and shutdown fallback.

</details>

PR #2072, opened 2026-09-09
Reorganize training settings and improve toolbar feedback

francescofugazzi:
Depends on #2048, which in turn depends on #2047.

## Summary

Reorganize the RmlUi Training panel while retaining existing parameter bindings, training actions, edit restrictions and project settings.

### Main settings

- Keep Training Method, Camera & Rasterization, Masking & Segmentation, and Dataset before Advanced.
- Retain strategy, backend, iterations and its automatic-scaling lock, capacity, BG Improvements, Undistort, Mip Filter, Exposure Correction activation and masks in the main settings.
- Place Dataset immediately before Advanced, expanded by default while respecting saved collapse preferences.

### Advanced settings

- Provide one real activation control per optional feature, without duplicate enable checkboxes.
- Give Depth, Normal, PPISP, Bilateral Grid, Exposure Correction tuning, Evaluation, Random Initialization and Sparsity separate conditional sections.
- Keep background modes and specialist tuning in Advanced.
- Preserve tuning values when features are disabled and re-enabled.

### Search and editing

- Keep matching controls reachable through collapsed ancestors without overwriting saved expansion preferences.
- Align search ownership with the displayed sections.
- Keep background mode selection and image browsing available during image searches.
- Preserve numeric drafts during unrelated refreshes, correctly handle Enter, blur and Escape, and reconcile authoritative changes or rejected updates without a recurring refresh loop.

### Toolbar and feedback

- Use compact icon-and-text actions with tooltips and accessible labels.
- Derive icon-only mode from measured RmlUi widths rather than fixed panel-width thresholds.
- Keep the status badge aligned with the actions and place multiline errors outside the action row.
- Represent restoration, stopping, saving and failure states.
- Keep project saving directly available while Running and Paused.
- Show descriptor-backed capability notices and block Start when the selected configuration is invalid.

### Localization

- Remove duplicate and unused locale entries.
- Reject duplicate JSON keys in the locale completeness checker.

### Search and layout

- With Advanced collapsed, search `means_lr` and `use_normal_loss`. Confirm the matching controls become reachable and clearing search restores the previous expansion state.
- Search `use_exposure_correction`; confirm its activation remains visible in Camera & Rasterization.
- In Background Color mode, search `bg_image`; confirm Mode and Browse remain available.
- Confirm Dataset is before Advanced and expanded with fresh panel preferences. Restore saved collapsed preferences and confirm they remain respected.
- Search Dataset fields and confirm Advanced does not open unnecessarily.

### Numeric editing

- In Ready, slowly replace Max Gaussians, including leaving the input empty briefly. Confirm the previous value is not restored during typing.
- Test Enter, clicking outside, Escape and +/- independently.
- Repeat with Iterations and a learning rate.
- After committing with Enter, start another draft without leaving the field; Escape should return to the last committed value.

### Configuration and persistence

- Switch between 3DGS and 3DGUT with supported and conflicting options. Confirm selected conflicts block Start and removing them restores availability.
- Enable PPISP and its controller, then Exposure Correction. Confirm conflicting standalone flags are cleared and tuning values remain preserved.
- Test standalone PPISP, Bilateral Grid and their combination.
- Enable Depth and Normal together and confirm their parameters remain separate.
- Toggle each Advanced feature, save and reopen a project, and check flags and tuning values.
- Exercise the iterations lock, mask modes, background modes and dataset edit restrictions.

### Toolbar and training lifecycle

- Test Ready, Starting, Running, Paused, Completed, Stopped, Error, Restoring and Stopping/Saving transitions.
- Check German Completed around 360dp and Paused around 450dp at 100%, 150% and 200% UI scale.
- Narrow and widen the panel in English, German and Japanese. Confirm captions disappear when needed and return when space allows, without wrapping or clipping.
- Save a .licht project while Running and Paused, reopen it and resume training.
- Cancel Reset/Clear confirmation dialogs and confirm no work is discarded.
- Reproduce a failed trainer restore and confirm persistent error feedback. A later trainer failure must show its current error rather than stale restore details.
- Confirm final model saving displays SAVING and transitions to the final trainer state.


MrNeRF (changes requested):
First review of head bf2df0fed (stacked on #2048 -> #2047, so the PR-only diff is 18 files, +1815/-396). Built with tests and run live on Linux: C++/python suites, then the whole manual test plan plus the review checklist in the GUI on Xvfb with real clicks, in English, German and Japanese and at 100/150/200% scale.

What works: backend switch with IGS+/Mip/Depth/Normal blocking Start and clearing restoring it, every optional activation opening only its own section (Depth and Normal stay separate), reopening a project with seven optional features on keeps all 25 checked values, the six ordinary toolbar states, project save while Running/Paused, padlocks and edit locks, mask modes and background modes, locales render.

Please fix before merge (all reproduced live, screenshots on request):

1. Searching `use_exposure_correction` hides the only Exposure activation. The schema puts it in Training Method, the widget lives in the Camera block, so the search shows an empty "Training Method" header. Clearing the search brings it back.
2. Search does not open Advanced. `means_lr` or `use_normal_loss` with Advanced collapsed leaves the match behind the closed header (`property_view.py` skips `advanced_params` on purpose). Same family: searching `bg_image` in Color mode hides the Mode dropdown and the Browse button, so the result cannot be used.
3. Status badges drop the restoration states. A project whose trainer restore fails ends up as "PAUSED" with Resume/Reset/Stop/Save offered and no persistent error text once the toast is gone. Related: after such a failed restore, a later real training failure still shows the old "Project trainer restore failed" message instead of "Training failed."
4. All ten locale files now contain `section.masking` twice; the later "Mask Settings" value wins and the completeness check does not notice.
5. The new appearance-preservation test asserts binding ids that do not exist. Real clicks (PPISP on, its controller on, then Exposure on) leave PPISP off with the controller still on, which native validation then rejects. Inherited behaviour, but the test hides it instead of catching it.
6. German does not fit the fixed compact thresholds: Completed at 360 dp puts "Löschen" on a second row with the badge spanning both rows, Paused at 450 dp puts "Licht speichern" on row two, at every UI scale. Derive the compact switch from measured width, not fixed dp.
7. Leftovers: the standalone Bilateral block carries an exposure-start template that can never show, and six new locale keys have no consumer.

Nit: Escape in the iterations field keeps the typed value instead of cancelling the edit.

This has to wait for #2047 and #2048 anyway.

francescofugazzi:
Rebased this PR onto the updated #2048 and addressed the review findings:

- Aligned Exposure Correction search ownership with its Camera & Rasterization location.
- Made searches expand the Advanced ancestor when necessary, without changing saved collapse preferences.
- Kept Background Mode and Browse reachable when searching for `bg_image` from Color mode.
- Added restoration/error badges and persistent restore-failure details. Current trainer errors take precedence over stale restoration feedback.
- Removed duplicate `section.masking` entries from all ten locales and made the completeness checker reject duplicate JSON keys.
- Corrected the PPISP controller and sidecar-freeze property IDs used when enabling Exposure Correction, preserving tuning values.
- Replaced fixed compact thresholds with measured RmlUi button, gap and badge widths, with remeasurement after layout changes.
- Removed the unreachable bilateral exposure-start template and unused locale keys.
- Fixed Escape cancellation being overwritten by the subsequent blur event.

Additional fixes found during follow-up testing:
- Removed a refresh/publication loop that restored Max Gaussians while typing.
- Preserved focused numeric drafts during unrelated refreshes while still accepting authoritative changes, including rollback arriving after the first post-edit refresh.
- Added a distinct SAVING badge during final model saving.

As discussed, Dataset now sits in the main panel immediately before Advanced. It starts expanded, while existing saved collapse preferences and dataset edit restrictions remain respected. Dataset searches no longer open Advanced.

The targeted Python suite now passes 266 tests. Locale completeness, hardcoded UI, error-debt and I/O checks also pass.

GUI testing so far is partial; the full language/scaling matrix and all runtime scenarios have not been repeated on this revision. This remains dependent on #2048.

francescofugazzi:
Addressed the additional training-panel layout feedback.

- Moved runtime status out of the action toolbar and changed it to lightweight informational text rather than a button-like badge.
- Balanced the visible training actions to a consistent width while retaining the automatic icon-only fallback for narrow panels and long translations.
- Renamed “Training Method” to “General Parameters” and simplified “Advanced Training Parameters” to “Advanced Parameters”.
- Promoted Background and Exposure & Appearance to the main panel.
- Kept Dataset in the main panel immediately before Advanced.
- Reorganized Advanced settings by purpose, placing every optional feature’s single activation checkbox next to its own conditional parameter section.
- Kept Advanced and Initialization collapsed by default while preserving explicitly saved panel expansion preferences.
- Preserved the existing parameter IDs, editing locks, search behavior, backend capability restrictions and training action flows.
- Updated all bundled locales, documentation and regression coverage.

The targeted Python suites pass with 196 tests, the confirmation/property-system suites pass with 52 tests, and the standalone localization, locale-completeness and hardcoded-UI checks all pass.

PR #2085, opened 2026-09-10
Masks: compare segment_and_ignore bands on the normalized mask scale

rkskmt:
## Problem

`--mask-mode segment_and_ignore` drops every pixel from the photometric loss and never applies the opacity penalty. Training runs to completion with the loss pinned at ~0 from the first logged iteration, and nothing reports an error.

A three-band mask is authored on the eight-bit scale — `< 128` ignore, `128..250` segment, `> 250` keep — and the fused preprocess kernel still compares against those raw levels (`v > 250.0f`, `128 <= v <= 250`). The pipelined loader stopped delivering them: `io::process_mask` returns `mask.to(Float32) / 255.0f`, so a keep sample of 255 arrives as `1.0`, the band comparisons are false for every pixel, and the mode degenerates to "ignore everything".

Sidecar masks under `masks/` take that path, so the mode is dead for the common case. In-memory masks still work, because they arrive through `Camera::load_and_get_mask`, which still hands over `UInt8` 0..255.

## Root cause

`9169a2a0` (#1588) fused the trainer's `masked_fill` band remap into `mask_preprocess.cu`, carrying the `250` / `128` bounds over unchanged — the kernel comments still describe the remap it replaced (`// After trainer remap: >250 kept as 255`), and `tests/test_mask_preprocess_fusion.cpp` still checks against that "old trainer chain". The same commit rewrote `process_mask` to normalize to `[0,1]`. The consumers kept the eight-bit domain; the producer left it.

The existing tests do not catch this because they feed `UInt8` samples straight into the kernel, bypassing the loader.

## Fix

Express the band bounds on the `[0,1]` scale in one place (`kMaskKeepMin`, `kMaskSegmentMin` in `mask_preprocess.hpp`), and scale `UInt8` samples the same way when reading a band, so both kernel entry points share one value domain.

The bounds sit on the midpoints between neighbouring eight-bit levels (`250.5/255`, `127.5/255`) rather than on the levels themselves, so the rounding in the normalization cannot move an authored level across a band edge. Classification is unchanged for raw eight-bit input — `v/255 > 250.5/255` iff `v > 250` — which is what keeps the existing `UInt8` fusion tests passing untouched.

The tile error mask in `trainer.cpp` picks its bound by dtype for the same reason: masks that come through `Camera` still arrive as raw levels, and the eval mask in `metrics.cpp` reads those, so both stay correct without a change.

## Tests

`tests/test_mask_preprocess_fusion.cpp` gains coverage of the shape masks actually arrive in:

- `SegmentAndIgnorePhotometricMatchesReferenceForNormalizedMask` and `SegmentAndIgnoreOpacityPenaltyMatchesReferenceForNormalizedMask` — a `Float32` `[0,1]` mask must classify identically to the raw eight-bit reference
- `SegmentAndIgnoreBandLevelsMatchForBothMaskTypes` — sweeps all 256 levels through both entry points and pins the `127/128` and `250/251` band edges

All three fail on `master` and pass with this change; the seven existing fusion tests are untouched and still pass.

Beyond the unit tests, I drove the pipelined loader over an eight-bit three-band PNG — as a sidecar mask and as an RGBA alpha channel, on the first load and on the cached reload — and checked the delivered mask through the fused entry points: every band classifies correctly with this change and collapses to Ignore without it.

## Reproducing

Point any COLMAP dataset at a `masks/` directory of eight-bit three-band PNGs (255 keep, 200 background, 0 ignore) and train with `--mask-mode segment_and_ignore`. Every sample lands outside both bands, so the photometric weight map is identically zero and the opacity penalty never fires: the reported loss sits at ~0.0000 from the first logged iteration, while `--mask-mode none` on the same data trains normally.

### DIFF
--- a/src/training/kernels/mask_preprocess.cu
+++ b/src/training/kernels/mask_preprocess.cu
@@ -27,10 +27,11 @@ namespace lfs::training::kernels {
             return std::min(num_blocks_1d(total), kMaxReduceBlocks);
         }
 
+        /// Float32 band masks are normalized; UInt8 masks retain their 0..255 scale.
         template <typename MaskT>
-        __device__ __forceinline__ float load_mask_raw(const MaskT* mask, const int idx) {
+        __device__ __forceinline__ float load_mask_normalized(const MaskT* mask, const int idx) {
             if constexpr (std::is_same_v<std::remove_cv_t<MaskT>, uint8_t>) {
-                return static_cast<float>(mask[idx]);
+                return static_cast<float>(mask[idx]) / 255.0f;
             } else {
                 return mask[idx];
             }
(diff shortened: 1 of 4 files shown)


MrNeRF:
Thanks, looks good!

PR #2086, opened 2026-09-10
Fix duplicate selection brush on Wayland

constantin906:
Fix duplicate selection brush rings under Wayland by using only the existing Vulkan brush overlay

SDL clients can run through XWayland even in a Wayland session, so the check covers the SDL video backend plus `XDG_SESSION_TYPE` and `WAYLAND_DISPLAY`.

### Testing
- `cmake --build build --target LichtFeld-Studio -j2`
- Tested on Fedora 44, Niri/Noctalia Wayland; import  a PLY, press `1` to use the selection tool, hold left click; only one brush is visible and selection works

### Tradeoff
The Vulkan brush follows rendered frames, so it is less responsive than
the compositor cursor it replaces

### DIFF
--- a/src/visualizer/gui/gui_manager.cpp
+++ b/src/visualizer/gui/gui_manager.cpp
@@ -91,6 +91,7 @@
 #include <chrono>
 #include <cmath>
 #include <condition_variable>
+#include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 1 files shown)


MrNeRF:
Thank you!

PR #2090, opened 2026-09-11
fix: add Pascal viewer compatibility and avoid scratch cleanup deadlock

parsamahdavi48:
On a Windows GTX 1060 6 GB (SM 6.1, driver 571.96), master `da2fb64` could complete headless training but crashed in `nvoglv64.dll` when displaying Gaussians. After addressing that compatibility path, full-resolution GUI training exposed a separate shared-scratch cleanup deadlock. This draft fixes both reproduced failures while preserving the existing supported-device paths.

Related to #1681. This does not propose changing the project's official SM75+ support policy. Feedback on the maintenance suitability of these fallbacks is welcome.

### Changes

- Guard WMMA convolution/SH-assignment device code and retain scalar/SIMT work below SM70. Select SIMT before benchmarking so an empty architecture-guarded WMMA kernel cannot be chosen.
- Disable Vulkan conditional rendering by default on the selected pre-Volta CUDA device, retaining an explicit diagnostic override and the existing fallback renderer.
- Select FP32 SH when Vulkan lacks `shaderFloat16` or `storageBuffer16BitAccess`. Decode q16/IEEE-half data correctly and preserve exportable storage capacity/layout through loading, hydration and rebinding.
- Release the global arena initialization mutex before waiting in `clear_external_backing()`. An active training frame may need `get_arena()` to finish and release itself; holding the manager mutex across that wait deadlocked the viewer and trainer. The change follows the existing runtime-owned arena lifetime contract and does not change training mathematics or quality settings.
- Document the reproduction, executed validation and limits in `docs/pascal_compatibility.md`.

### Evidence

Executed on GTX 1060 with CUDA 12.8.61, MSVC 14.44.35207 and native Release/SM61:

- Conditional-rendering A/B: disabling it allowed GUI training/rendering; re-enabling it with q16 still disabled reproduced the access violation on an independent 513-Gaussian SH2 fixture. This isolates the failing path without claiming knowledge of the driver's internal defect.
- Deadlock: native stacks showed the renderer holding the manager mutex inside scratch cleanup while the trainer waited for that mutex in `get_arena()` during FastGS frame release. The five decisive machine-code ranges matched the diagnostic symbol mapping; no whole-binary identity is claimed.
- A standalone compiled concurrency regression passed: manager lookup completed while clear awaited an active frame, then clear completed after the frame was released. Independent review found no blocking issue.
- The old isolated GUI reproduction stalled at iteration 641. The repaired GUI completed 21,425 iterations with 713 full-resolution images, MRNF, SH3 and a 1,000,000-splat cap before the user intentionally stopped it. Repeated storage/scratch growth, camera-view changes, checkpoint saving, clean stopping and reopening the saved state passed. The repaired run was configured for 71,300 iterations; the old reproduction used 30,000 total steps with the scaled early refinement schedule, so this was not an exact full-schedule A/B test.
- Four 3,000-step headless runs compared q16 and FP32 using the same patched executable, resize factor 2, SH2, 623 training and 90 held-out images. Mean q16/FP32 scores: PSNR 20.697513/20.778174 dB, SSIM 0.719228/0.719667, LPIPS 0.387395/0.386986. No mean quality regression was detected; the differences were smaller than the larger within-mode repeat gap for each metric. This does not establish equivalence to an untouched upstream build or a causal quality improvement.
- Earlier CUDA checks: six LPIPS convolution GPU cases (41,344 outputs) matched an independent CPU reference; eight SH-assignment GPU cases matched an independent FP32/FMA reference; full LPIPS identical/distinct-image checks passed.
- Saved q16 conversion preserved 12,312 coefficients exactly. The earlier 600-step GUI save/reopen test passed but did not expose the later deadlock.
- Native Release rebuild and installed-binary hash verification passed. Retained SM75 kernel paths compiled but were not executed. The original compatibility commit passed clang-format 19.1.5; the additional focused change passed `git diff --check` and independent review.

No build or training workload was repeated solely for this PR update. Private datasets, profiles, local paths, captures and binary artifacts are excluded.

### Limits and tradeoffs

The user stopped validation at 21,425 of 71,300 iterations; full-run completion and converged-quality parity remain unverified. The upstream full test suite, other GPU models and all formats have not been tested. Vulkan validation layers were unavailable.

FP32 SH uses more VRAM; the existing non-conditional renderer uses 16 depth waves rather than 64. Automatic quantization selection is process-global, so future in-process GPU switching needs reassessment. Arena lifetime still requires shutdown after active users stop; concurrent manager reset/destruction is not newly supported.

### DIFF
--- a/docs/pascal_compatibility.md
+++ b/docs/pascal_compatibility.md
@@ -12,7 +12,7 @@ The same device also lacked the Vulkan16-bit capabilities required for q16 SH. W
 
 ## Executed validation
 
-Environment: Windows, GTX1060 6GB/SM6.1, driver571.96, CUDA12.8.61, VS2022 17.14.37/MSVC14.44.35207, clang-format19.1.5, native Release with `CMAKE_CUDA_ARCHITECTURES=61`, Python3.12.13. All results below precede a final whitespace-only clang-format pass.
+Environment: Windows, GTX1060 6GB/SM6.1, driver571.96, CUDA12.8.61, VS2022 17.14.37/MSVC14.44.35207, clang-format19.1.5, native Release with `CMAKE_CUDA_ARCHITECTURES=61`, Python3.12.13. The initial validation table below preceded a final whitespace-only clang-format pass. Later validation is recorded separately below.
 
 | Check | Observed result |
 |---|---|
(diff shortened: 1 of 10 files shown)


MrNeRF:
Thanks! Let's merge it!

PR #2093, opened 2026-09-11
Override checkpoint data [Volinga]

Ediolot:
Allow for override of optimizer checkpoint data from params for easier experiments. Allow zeroing of means_lr and scaling_lr.


MrNeRF:
Thanks for this pr. 

I am preparing a release soon and this will unfortunately introduce subtle bugs. This treats every resume to  reset the learning rates. This cannot be merged.

I am not 100% sure what the intention is. The training is very fast and everything that's important happens pretty much in the first part of the training. It is win or lose in the beginning. So you should be able to iterate very quickly.

If you let me know what your intention or you goal is, this can be scheduled.

PR #2123, opened 2026-09-13
Fix MSVC C4273 for exported tensor and Vulkan UI friends

francescofugazzi:
## Problem

MSVC emits C4273 (inconsistent DLL linkage) in downstream translation units because three friend declarations omit the DLL annotations already present on their exported namespace declarations. The affected functions are `gpu_backend_of`, `tensor_vulkan_buffer`, and `setVulkanUiTextureContext`.

## Change

Apply `LFS_CORE_API` to the two tensor friends and `LFS_VIS_API` to the Vulkan UI texture friend. The patch changes only declarations (three lines), without changing function behavior.

## Validation

- The warning was observed in a Windows build of upstream `dev` at `6083a3b09`.
- `clang-format --dry-run --Werror` passed for both changed headers.
- `git diff --check` passed.
- A Windows rebuild with this patch has not yet been reported; this PR is a draft until that confirmation.


MrNeRF:
thank you!

PR #2124, opened 2026-09-13
Add label tool to HTML viewer export

jacobvanbeets:
Adds a label annotation tool to the self-contained HTML viewer, following the existing measure-tool pattern (toolbar button + vendored tool file, like the gizmo/measure tools).

## Features
- **Tag button** toggles label mode; click the model to place a named label (same depth-based `Picker` as the measure tool)
- Placed points get a **TranslateGizmo attached**, so a mis-picked point can be nudged precisely before saving
- Click to select/move, double-click to rename (empty text deletes), right-click to delete
- **Save** button writes e.g. `Sample.lbl` (suggested name derived from the scene file name, `Sample.html` -> `Sample.lbl`) via the File System Access API, with an anchor-download fallback
- **Load** button reads a `.lbl` file back into the model (replaces current labels after confirm)

## .lbl format (simple JSON)
```
{
  "format": "lfs-labels",
  "version": 1,
  "labels": [
    { "text": "Door", "position": [1.25, -3.4, 2.1] }
  ]
}
```

## Implementation
- New `label-tool.js` is concatenated into the export IIFE by `src/io/formats/html.cpp` alongside `gizmo.js` and `measure-tool.js` (its trailing `export` is stripped there, same as the other tools)
- The 100k-line `index.js` bundle is **untouched**
- Measure and label modes are mutually exclusive (both attach a left-button gizmo)
- `tests/test_python_io.cpp` asserts the label-tool hook is embedded and its export statement stripped

## Testing
- Full export bundle (index.js + IIFE) verified with `node --check` (this also caught a real `CLICK_DEADZONE` name collision with measure-tool.js in the shared IIFE scope, fixed)
- Build + in-browser verification to follow


src/visualizer/gui/resources/viewer/label-tool.js
```diff
@@ -0,0 +1,483 @@
...
+        }
+        const trimmed = text.trim();
+        if (!trimmed) {
+            return;
+        }
+        addLabel(trimmed, result);
+    };
+
+    const onDoubleClick = (e) => {
+        if (!active || !isPrimary(e)) {
```
MrNeRF: Double-click renaming doesn't work in Chromium. `dblclick` is a `MouseEvent`, so it has neither `pointerType` nor `isPrimary`, and this guard always returns. Please check `e.button` here.

The viewer also handles the same gesture as a camera pick on `pointerdown`. In the browser test, double-clicking a label opened an add-label prompt and created another label instead. Please consume the editing gesture before camera navigation handles it, and add a browser regression test.

src/visualizer/gui/resources/viewer/label-tool.js
```diff
@@ -0,0 +1,483 @@
...
+        }
+        app.renderNextFrame = true;
+        refreshVisuals();
+    };
+
+    // ---- visuals --------------------------------------------------------
+    const refreshVisuals = () => {
+        svg.classList.toggle('hidden', labels.length === 0);
+        for (let i = 0; i < labels.length; i++) {
+            camera.camera.worldToScreen(labels[i].position, _screen);
```
MrNeRF: Labels behind the camera are still drawn and hit-testable. I reproduced this with a point two units behind the camera: its text appeared in the middle of the model. This also happens when flying past an annotated point. Please check camera-space depth before drawing or hit testing. The bundled `Annotation` code already uses `viewMatrix.transformPoint()` and hides points with `z >= 0`.

src/visualizer/gui/resources/viewer/label-tool.js
```diff
@@ -0,0 +1,483 @@
...
+            position: [label.position.x, label.position.y, label.position.z]
+        }))
+    }, null, 2);
+
+    const suggestedName = () => {
+        try {
+            const name = decodeURIComponent(new URL(location.href).pathname.split('/').pop() || '');
+            const base = name.replace(/\.[^.]*$/, '');
+            if (base) {
+                return base + '.lbl';
```
MrNeRF: Optional: could we use `.labels.json` instead of `.lbl`? The contents are ordinary JSON and the loader already accepts `.json`. That would make the format obvious when sharing or editing these files.

MrNeRF (changes requested):
Tested the viewer in Chromium with a generated scene. Placement, JSON save/load, invalid-file handling, right-click deletion, and mode switching worked. Double-click renaming and labels behind the camera need fixing before merge.

These were browser checks; I did not rebuild Studio locally.

PR #2144, opened 2026-09-15
Fix PLY comparison alignment, black panels and crash

francescofugazzi:
PLY comparison can show black panels because a single owned model inherits the entire scene's SH limits while using only one transform slot. When panel rendering fails, the full-frame fallback can then dereference the intentionally absent combined model and crash.

Scope each owned model's scene bindings, filters and selection state to its comparison panel, and guard the fallback against an absent model. Models continue to render directly without creating a combined copy for comparison.

Use consistent pixel-center sampling in the Vulkan and CPU compositors so both models remain aligned while dragging the divider.


MrNeRF:
The alignment fix looks good, but please keep rendering each model directly and fix the crash there. The combined copy uses more GPU memory and may slow comparisons down, especially with several large models loaded.

francescofugazzi:
Updated to keep rendering each model directly. The failure was in the per-panel request: it replaced the transform array with one node transform but retained the scene-wide SH limits. VkSplat rejected that mismatch, then the full-frame fallback dereferenced the null combined model.

The request now scopes scene bindings together with filters and selection state to the owned node. The fallback also checks that a full-frame model exists. The combined-copy workaround has been removed, and the pixel-center alignment fix remains.

MrNeRF (approved):
The per-model settings and fallback guards look correct. With current master and the PPISP fix, 151 focused tests pass; the eight trained models survive 12 matched-pair switching checks and six camera-change captures across EWA and GUT.

Both Windows failures are the missing alignment export fixed in #2147, whose Windows builds pass.

PR #2152, opened 2026-09-16
Fix spacing in wayland and xkbcommon module versions

Alatar98:
Proper syntax per https://linux.die.net/man/1/pkg-config. Otherwise pkg-config will always return 1 i.e. fail.


MrNeRF:
thx!

PR #2164, opened 2026-09-17
Fix Windows window bounds and Asset Manager resize cursors

francescofugazzi:
## Summary

Fixes two UI regressions affecting window placement and Asset Manager resize feedback.

### Respect the usable display area

Borderless maximization now uses the complete SDL usable display bounds, including the display origin. This prevents the application from being cropped by a Windows taskbar positioned at the top, PowerToys docks, or other reserved desktop areas.

### Refresh Asset Manager resize cursors

Hosted RmlUi panels now re-hit-test the pointer before the final layout update. Resize cursors and hover styling therefore remain synchronized when a data-model update moves a resize handle without physical mouse movement.

Cached pointer modifiers are also cleared on document reload and pointer leave.

### Preserve native text-input behavior

Backspace and Delete remain on RmlUi's native key path. Focused real-context coverage confirms that native deletion updates the input value and dispatches the `change` event without introducing a platform-wide interception.

The native-key regression test lives in its own test source rather than the status-bar fit suite.


src/visualizer/gui/rmlui/rml_text_input_handler.cpp
```diff
@@ -47,11 +47,34 @@ namespace lfs::vis::gui {
...
+            if (start == end) {
+                if (key_identifier == Rml::Input::KI_BACK) {
+                    if (start == 0)
+                        return true;
+                    --start;
+                } else {
+                    ++end;
+                }
+            }
+            input_context_->SetText(Rml::StringView(), start, end);
```
MrNeRF: This is the blocker. Returning `true` here means the callers skip `rml_context_->ProcessKeyDown`, and that is the only path that fires the `change` event (RmlUi does it from `DeleteCharacters()`).

Going through `SetText` instead ends up in `WidgetTextInput::SetValue`, which only fires `change` when its sanitiser shortens the string, so a normal delete fires nothing. Typing still fires it, so it is deletions only.

Two ways to fix it:

1. Keep this and fire the event yourself after `SetText`, with the new value in the `value` parameter, the way `DispatchChangeEvent` does.
2. Or drop the interception. Backspace and Delete do arrive in `keys_pressed` on Windows as well (`frame_input_buffer.hpp` pushes every `SDL_EVENT_KEY_DOWN`, and the scancodes map fine in `sdl_rml_key_mapping.hpp`), so something else is eating them there. I would rather find that than route two keys around RmlUi on every platform. If you can tell me what the Windows symptom looked like exactly, I am happy to help dig.

src/visualizer/gui/rmlui/rml_text_input_handler.cpp
```diff
@@ -47,11 +47,34 @@ namespace lfs::vis::gui {
...
+            input_context_->SetSelectionRange(0, std::numeric_limits<int>::max());
+            return true;
+        }
 
-        input_context_->SetSelectionRange(0, std::numeric_limits<int>::max());
-        return true;
+        // SDL text events only carry inserted text. Route the two editing keys
+        // through the active text context as well so platform text-input mode
+        // cannot leave them to an unrelated application shortcut path.
+        if ((key_identifier == Rml::Input::KI_BACK || key_identifier == Rml::Input::KI_DELETE) &&
```
MrNeRF: Smaller thing, same block. There is one shared handler for the whole app, and `input_context_` is whatever text field was focused last, anywhere. Each panel has its own RmlUi context with its own focus, and blurring only clears the local one.

With Ctrl+A that was harmless. With Delete it is not: a Delete meant for one panel can now get swallowed and applied to a field in another. Every caller already has the focused element handy, so adding a check that it is actually a text field (`rml_input::isTextEditableElement`) would close it.

src/visualizer/gui/rmlui/rml_panel_host.cpp
```diff
@@ -593,6 +593,13 @@ namespace lfs::vis::gui {
         if (needs_post_layout_update)
             rml_context_->Update();
 
+        // A data-model update can move a resize handle without any physical
+        // mouse motion. Re-hit-test the settled layout so RmlUi republishes the
+        // cursor belonging to the element now under the pointer.
+        if (last_hovered_ && last_forwarded_mx_ >= 0 && last_forwarded_my_ >= 0)
+            rml_context_->ProcessMouseMove(last_forwarded_mx_, last_forwarded_my_,
```
MrNeRF, line 598: The idea is right and I am not arguing with it, `Update()` only refreshes the hover chain at its start so a settled layout does need one more pass, and the wheel handler further down does the same trick.

But this sits right before `content_dirty_ = false` and `render_needed_ = false`. So if the re-test changes which element is hovered, the `:hover` styling that comes with it is never laid out or drawn this frame, and `drawDirectCached()` then happily reuses the old picture because both flags are clear. End result: the cursor updates but the hover highlight can stay wrong until something else dirties the panel.

Either move this above the `needs_post_layout_update` update, or set `render_needed_ = true` when the hovered element actually changed.

src/visualizer/gui/rmlui/rml_panel_host.hpp
```diff
@@ -159,6 +159,7 @@ namespace lfs::vis::gui {
         int last_layout_padding_ = -1;
         int last_forwarded_mx_ = -1;
         int last_forwarded_my_ = -1;
+        int last_forwarded_mods_ = 0;
```
MrNeRF, line 162: Nit: `last_forwarded_mx_` and `last_forwarded_my_` get reset to -1 when the pointer leaves the panel and on document reload, but this one never does. Worth resetting it in the same two places so the replay cannot use modifiers from an old hover.

tests/test_rml_text_input_handler.cpp
```diff
@@ -71,3 +78,43 @@ TEST(RmlTextInputHandlerTest, IgnoresPlainAWithoutModifier) {
     EXPECT_EQ(context.selection_start, -1);
     EXPECT_EQ(context.selection_end, -1);
 }
+
+TEST(RmlTextInputHandlerTest, BackspaceDeletesSelectionOrPreviousCharacter) {
```
MrNeRF: These are fine as far as they go, but they cannot catch the problem above: the stub's `SetText` just records the numbers, so there is no real text field and no event to miss.

A real test is cheap here. `tests/test_status_bar_fit.cpp` already stands up a real `Rml::Context` with a font and a stub render interface, so you can drop in an `<input type="text" value="hello">`, focus it, hang a `change` listener on it, and assert both the new value and that the event fired. Two gotchas I hit doing it: pick the loaded font in the inline style or RmlUi leaves its line list empty and crashes, and set the caret position explicitly since it starts at 0.

src/visualizer/window/window_manager.cpp
```diff
@@ -1643,13 +1643,10 @@ namespace lfs::vis {
...
         }
 
-        SDL_Rect target_bounds = usable_bounds;
-#if defined(_WIN32)
-        // Windows-only: keep work-area dimensions so the taskbar stays visible.
-        // Ask for the display top; the WM may still clamp managed windows to work-area y.
-        target_bounds.y = display_bounds.y;
-        target_bounds.h = std::min(usable_bounds.h, display_bounds.h);
-#endif
+        // Borderless windows are not constrained by the native non-client area.
```
MrNeRF, line 1646: This one is good. Just noting for whoever reads the history later: this looks like a revert of "fix gap on maximize" but it is not really, that commit applied to every platform and was about the GNOME top bar. The `_WIN32` fence came later and the comment on it was a guess. No test covered the old behaviour, so nothing to update.

MrNeRF (changes requested):
Thanks @francescofugazzi. Built this on Linux (Release, current master merges clean) and went through all three fixes, plus dug through the git history on the window one.

Two of the three look good:

- The window bounds change is right, and I think it is more clearly right than the PR text says. The `display_bounds.y` line came in as `e383364a9` "fix gap on maximize" and was not Windows-specific at all back then; #1668 later wrapped it in `#if defined(_WIN32)` only to avoid touching Windows, on a guess rather than a measurement. On Windows that line only does anything when the taskbar is at the top, which is the exact case it breaks. Linux and macOS come out identical.
- The resize cursor fix works on the right mechanism, and it matches what the file already does after a mouse wheel. One placement note inline.

The Backspace/Delete change has a blocker: it deletes the text but no longer fires RmlUi's `change` event. I wrote a throwaway test against a real `Rml::Context` to be sure:

| | value after | change events |
|---|---|---|
| RmlUi's own path (today) | `hell` | 1 |
| this PR, Backspace | `hell` | 0 |
| this PR, Delete | `ello` | 0 |

That matters because `data-value` is how most text fields in the app talk to their panel, and RmlUi only writes the field back into the data model on `change`. There are 80 text inputs and textareas using it. Easiest one to see: open the Gallery publish panel (the title field opens focused and selected), press Backspace to clear it, hit Publish. It publishes the old title, and Submit never greys out, because both the submit and the validation read the model, not the box. Plugin text/number fields and the VRAM HUD filters do the same. Rename, package search, the sequencer fields and the video extractor get away with it because they re-read the box on blur or submit.

Also worth flagging: this change is not inside `#ifdef _WIN32`, so it lands on Linux and macOS too.

Rest is inline. Everything else checks out: clean build, no new warnings, clang-format clean, 4/4 text input tests, 44/44 window tests, 26/26 Rml tests. The full suite shows 37 failures but all 37 fail on master sources too (missing `data/bicycle` fixtures), so none are yours. Small correction to the PR text: the Asset Manager run is 156 passing and 1 failing, not 154 passing, and that one failure is also already on master.

I could not test the Windows side, no Windows machine here, so I am taking your manual checks for that.

francescofugazzi:
Addressed all notes`:

- Backspace/Delete now use RmlUi’s native key path, preserving `change` events and data bindings.
- Replaced the stub deletion tests with a real `Rml::Context` test covering value updates and `change` dispatch.
- Moved pointer re-hit-testing before the final `Update()` so hover styling is rendered correctly.
- Reset cached modifiers on document reload and pointer leave.

I’ll also revalidate the Windows-specific behavior before updating the PR.

tests/test_status_bar_fit.cpp
```diff
@@ -368,6 +381,38 @@ namespace {
...
+            rmlui_dynamic_cast<Rml::ElementFormControlInput*>(input_element.get());
+        ASSERT_NE(input, nullptr);
+        ChangeListener listener;
+        input->AddEventListener(Rml::EventId::Change, &listener);
+        document_->AppendChild(std::move(input_element));
+        input->Focus();
+        context_->Update();
+
+        input->SetSelectionRange(5, 5);
+        EXPECT_TRUE(context_->ProcessKeyDown(Rml::Input::KI_BACK, 0));
```
MrNeRF: This fails, and so does the same assertion on line 409.

`Context::ProcessKeyDown` returns whatever `DispatchEvent` returns, which is `false` once a listener stops propagation (`Context.cpp:530-531`). RmlUi's text input does stop it for these keys: `WidgetTextInput.cpp:610-611` runs `if (shift || ctrl || !out_of_bounds || selection_changed) event.StopPropagation();`, and for Backspace `out_of_bounds` stays false, so it always stops.

So `false` here means "RmlUi consumed the key", which is the behaviour you want — it is how the app knows not to treat it as a shortcut. `EXPECT_TRUE` asserts the opposite.

Easiest fix is to drop both `EXPECT_TRUE` wrappers and just call `context_->ProcessKeyDown(...)`, since the return value is not what this test is about. If you do want to pin it, `EXPECT_FALSE` is the correct expectation. Everything below these two lines already passes.

tests/test_status_bar_fit.cpp
```diff
@@ -368,6 +381,38 @@ namespace {
         assertFlexSiblingsDoNotOverlap(document_);
     }
 
+    TEST_F(StatusBarFitTest, NativeTextDeletionUpdatesValueAndDispatchesChange) {
```
MrNeRF: Nit, take it or leave it: a text input key test in the status-bar fit suite is a surprising place to find it. Reusing this file's `Rml::Context` harness is the pragmatic move and I would have done the same, but nothing here touches the status bar.

If it is cheap, lifting the harness (the stub render interface, `Rml::Initialise`, the font load) into its own `test_rml_text_input_native_keys.cpp` would put it where someone debugging text input will look. Not worth blocking on.

MrNeRF (changes requested):
Thanks, that is the right call on all three. Built d8ddf34ae on Linux and checked each one.

Good:

- The handler is fully back to master (`git diff master -- rml_text_input_handler.cpp` is empty), so `change` and the data bindings work again through RmlUi's own path.
- The new test proves it. I ran it and the substance passes: the value becomes `hell`, the listener fires once with `hell`, then `ello` and twice. That is exactly the regression guard that was missing.
- Re-hit-test moved above the post-layout `Update()`, so the hover styling now lands in the frame that gets drawn.
- `last_forwarded_mods_` is reset in both clearing paths.
- Window bounds unchanged and still right.

One thing to fix: **the new test fails as written.** Two assertions, details inline. The value and `change` assertions all pass, it is only the `EXPECT_TRUE` around `ProcessKeyDown` that is wrong. My guess is it slipped past because `test_status_bar_fit.cpp` already has two tests failing on master (`GalleryChipHidesBeforeAccountChip`, `GalleryChipHiddenWhenSignedOut`), so the file is red either way and a third failure is easy to miss. I have a PR out (#2178) that deletes those two, which will also conflict lightly with this file — happy to rebase around whichever of us lands first.

One question rather than a request: now that the interception is gone, this PR does not change Backspace/Delete behaviour at all any more, so whatever you were seeing on Windows is not fixed by it. Was that symptom maybe caused by something else, or does it still reproduce on your machine? Backspace and Delete do reach `keys_pressed` on Windows (`frame_input_buffer.hpp` pushes every `SDL_EVENT_KEY_DOWN`, and the scancodes map fine in `sdl_rml_key_mapping.hpp`), so if it still happens I would like to help find the real cause. Either way the PR summary should drop the "Restore Backspace and Delete" claim, since the change now is the regression test.

francescofugazzi:
Rebased onto current master, resolved the #2178 conflict, corrected the ProcessKeyDown expectations, moved the native-key test to a dedicated file, and updated the PR scope.

MrNeRF:
Thank you!

Issue #2175, opened 2026-09-17
"Move file to Trash" returns error on windows


kmeirlaen:
To simulate:

* open LFS
* open project manager
* right click a project
* choose "Move file to trash"
* On windows - error:

```[19:28:33.797] [error] module.cpp:3031  [Python] Failed to move project to trash C:\Users\Gebruiker\.lichtfeld\projects\untitled.licht: [WinError 2] Het systeem kan het opgegeven bestand niet vinden
```

Even though the file exists:

<img width="1641" height="310" alt="Image" src="https://github.com/user-attachments/assets/c92cac15-5630-44f8-97d3-7fd876832cdb" />

kmeirlaen:
verified in f87d90893 - move file to trash works.

Issue #2173, opened 2026-09-17
Scrolling bottom panels also activates resize


kmeirlaen:
Scrolling the bottom panels of the project manager also activates resize
(also some layout issues visible)

MrNeRF:
The fix from #2231 is already on master. The resize handle now sits above the Inspector, outside the scrolling area, so scrolling and resizing have separate targets. The layout and resize regression tests pass on current master.

I checked the current Inspector layout in the running Linux app. The original Windows mouse-wheel interaction still needs confirmation; I am leaving the issue open for that check.

francescofugazzi:
I added also a correction to #2255. Scrolling with mouse drag was overlapping with resize handle

kmeirlaen:
verified in d2fc60df7 - closed

Issue #2172, opened 2026-09-17
Gallery Thumbnail size inconsistent across views


kmeirlaen:
When setting a size for the thumbnail, the setting is not retained when resizing the panel - it looks like there are 3 different sizes (depending on the panels that are shown)

Expected: same size is used in all panels.

https://github.com/user-attachments/assets/5792647d-2bb3-4b1d-9729-d384423b212f

<img width="788" height="394" alt="Image" src="https://github.com/user-attachments/assets/02afb7fe-3515-40a0-ac16-45efd4fa9cea" />

kmeirlaen:
Verified in f87d90893 - OK

Issue #2170, opened 2026-09-17
Not sure how to proceed here


kmeirlaen:
After applying a sync operation (https://github.com/MrNeRF/LichtFeld-Studio/issues/2169), a conflict has occured. But i dont know how to resolve it...

https://github.com/user-attachments/assets/db8c88ac-5d8e-420f-a0a4-1111941db875

MrNeRF:
The clearer conflict choices are already on master through #2248: the review shows Local and Gallery values, with separate “Apply locally” and “Apply and update Gallery” actions. Tests confirm that choosing the local-only action or Cancel does not publish.

#2211 is receiving a separate improvement to explain what changed before you open the project. The deleted-online-copy publishing problem is tracked in #2208.

PR #2177, opened 2026-09-17
Improve Project Manager responsiveness, layout, and Gallery UX

francescofugazzi:
## Summary

This PR improves the Project Manager layout, interaction model, and runtime responsiveness across different panel sizes.

## Changes

- Make the project details Inspector an on-demand overlay instead of permanently reserving layout space.
- Automatically close the Inspector when selecting another project.
- Keep the Inspector visually separated and usable across responsive breakpoints.
- Preserve a single thumbnail size across all panel widths and layouts.
- Improve project-list scrolling by updating virtualized rows only when the visible record window changes.
- Batch the virtualized window to reduce repeated row reconstruction during scrolling.
- Use native flex layout for the Name column during horizontal panel resizing, avoiding Python model updates for every resized pixel.
- Keep navigator and column resizing continuous, without visible size quantization.
- Avoid unnecessary full responsive-layout invalidation while the panel remains within the same breakpoint.
- Prevent width-only list changes from rebuilding project rows.
- Use the Windows Recycle Bin implementation when moving projects to trash instead of relying on `gio`.
- Keep Gallery relink/account notices visible at account level without repeating warnings on every project.
- Treat the expected “account details are still loading” response as a recoverable Gallery state instead of logging an error traceback.
- Allow explicit Gallery refresh retries while avoiding repeated automatic failures.

## Related issues

Related to #2172 - Gallery thumbnail size was inconsistent across responsive layouts.

Related to #2173 - Project Manager scrolling and resizing interactions caused unnecessary layout work and undesirable resize behavior.

Related to #2175 - “Move file to Trash” failed on Windows because the implementation used the Linux `gio trash` command.

The Gallery notice and relink improvements also refine the surrounding synchronization UX, but do not claim to fully resolve the separate conflict and metadata presentation problems described in #2169 and #2170.


src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -78,6 +79,36 @@
...
+            ("fFlags", wintypes.WORD),
+            ("fAnyOperationsAborted", wintypes.BOOL),
+            ("hNameMappings", ctypes.c_void_p),
+            ("lpszProgressTitle", wintypes.LPCWSTR),
+        ]
+
+    operation = SHFILEOPSTRUCTW()
+    operation.wFunc = 3  # FO_DELETE
+    operation.pFrom = str(Path(path).resolve()) + "\0\0"
+    operation.fFlags = 0x0004 | 0x0010 | 0x0040 | 0x0400  # silent, no confirm, recycle, no error UI
```
MrNeRF: This can permanently delete a file that the user asked to move to the Trash.

The comment says "recycle", but `FOF_ALLOWUNDO` (0x0040) is documented as "Preserve undo information, **if possible**". It is a request, not a guarantee. Combined with `FOF_SILENT`, `FOF_NOCONFIRMATION` and `FOF_NOERRORUI`, a file on a network share or any volume without a Recycle Bin is deleted outright, `SHFileOperationW` returns 0, `fAnyOperationsAborted` stays false, and we treat it as a successful trash and remove it from the catalog.

The failure handling below is fine — a real error raises and `_set_catalog_notice("trash_failed")` shows it. The problem is that permanent deletion is not an error here.

Two ways out. Either drop `FOF_NOCONFIRMATION` and add `FOF_WANTNUKEWARNING` (0x4000), which exists precisely so the shell warns when a delete is about to become permanent, or check up front that the target can be recycled and refuse with a clear message if it cannot. I would lean to refusing: the "Move to Trash" confirmation the user clicked does not authorize a permanent delete.

Also worth using the documented flag names rather than the raw hex — this is exactly the kind of line where a wrong constant is invisible in review.

src/python/lfs_plugins/gallery_sync.py
```diff
@@ -479,19 +486,28 @@ def checked():
...
-    def refresh(self):
+    def refresh(self, *, force=False):
         if self.busy and self._operation == "refresh":
             return
+        if not force and self._relink_identity == self.identity():
+            return
         if self._unsupported_identity == self.identity():
             return
+        snap = self.account.snapshot()
+        if snap.signed_in and (not snap.email or not snap.connected_since):
```
MrNeRF, line 497: This turns a clear message into a silent dead end.

The comment says the account subscription will request a fresh sync when these fields arrive. It does not, and I checked the whole chain. The controller does subscribe (`gallery_controller.py:73`) and does call `refresh()` when the identity changes, but the account state that gets published only carries `signed_in`, `linking`, `disconnecting`, `error`, `membership_required`, `label`, `tier`, `tooltip` — not `email` and not `connected_since`. The store drops updates that compare equal (`ui/store.py:113-114`, and the native path does the same in `app_store.hpp`), so filling in only `connected_since` publishes an identical dictionary and no callback fires. Polling has already stopped by then, because it only continues while work or linking is pending.

So: signed in with a cached profile that has no `connected_since`, the refresh returns here, nothing ever retries, and the user sees no message at all. Before this change they at least got "Your account details are still loading. Refresh your account, then retry."

If you want the silent wait, the published account state needs to include the fields the wait depends on. Otherwise keep a user-visible loading state so there is a way out.

src/python/lfs_plugins/gallery_controller.py
```diff
@@ -824,11 +825,17 @@ def refresh(self):
             self._refresh_model()
             return
         self._refresh_requested = True
+        self._refresh_force_requested = self._refresh_force_requested or force
```
MrNeRF, line 827: The force intent is recorded here, but two early returns above it can discard it first.

`if self._refresh_pending and self.service.busy: return` fires before this line, so pressing the Gallery Refresh button while a refresh is already in flight throws the `force=True` away. The button has no busy guard, so this is easy to hit: open a Gallery scope (which kicks off an automatic refresh) and press Refresh while it is still running.

And it is worse than just losing the force flag. By that point `_refresh_requested` is already False, set a few lines down on the previous pass, so the poll handler's `if self._refresh_requested and not self.service.busy` does not retry at all. The click is simply swallowed. If the in-flight refresh then fails with `gallery_relink_required`, the relink latch in `gallery_sync.refresh` blocks automatic retries too, so the user is stuck with a Refresh button that appears to do nothing.

Record both the retry request and the force flag before those early returns.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -3534,8 +3569,64 @@ def _sync_asset_window_viewport(self, doc=None) -> bool:
...
+            signature = (
+                self._view_mode,
+                round(values[2], 1) if self._view_mode == "gallery" else None,
+                columns["gallery"], columns["size"],
+                columns["modified"], columns["folder"],
+            )
+            if signature != self._responsive_width_signature:
+                self._responsive_width_signature = signature
+                if self._view_mode == "gallery":
+                    self._dirty_fields("asset_card_slot_width", "asset_card_thumbnail_height")
```
MrNeRF: This marks the card geometry dirty but nothing recalculates it, so the Gallery keeps the old card size.

Both fields read the cached `_asset_card_slot_width`, and that cache is only recomputed inside `_window_assets`. The new signature returns `("gallery", columns, start, visible)`, so a width change that keeps the same column count and the same visible records does not rebuild the window, and `_sync_asset_window_viewport()` returns False.

Concretely: Gallery at 400dp high, default 168dp thumbnails, results width 570 to 580dp. Same three columns, same visible records, so the cache stays at 174.0dp when it should be 177.3dp. The dirty flag just re-publishes the stale number until some unrelated event rebuilds the records.

Card geometry needs recalculating independently of row reconstruction — and if the row height changes with it, the spacers need updating too.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -2160,6 +2182,9 @@ def _select_asset_id(
         self._selection_cursor_id = (
             asset_id if asset_id in self._selected_asset_ids else next(iter(self._selected_asset_ids), None)
         )
+        if self._selected_asset_ids != previous_selection and self._inspector_expanded:
+            self._inspector_expanded = False
```
MrNeRF: Auto-close only happens on this path. Keyboard navigation assigns `_selected_asset_ids = {asset_id}` directly in `_navigate_selection` without closing, so selecting A, opening the Inspector, then arrowing to B leaves the Inspector open showing B — or showing A's details, depending on what re-reads the selection.

Scope changes clear the selection without closing too, which leaves the overlay up over an empty folder, and the selection-repair path has the same gap.

Worth moving this into one place that every selection change goes through, rather than three call sites that each have to remember.

src/visualizer/gui/rmlui/resources/asset_manager.rcss
```diff
@@ -1177,11 +1179,10 @@
...
 .asset-shell.is-narrow .asset-resize-handle--vertical, .asset-shell.is-compact .asset-resize-handle--vertical { display: none; }
-.asset-shell.is-narrow .asset-inspector, .asset-shell.is-compact .asset-inspector { width: auto; height: 32dp; flex: 0 0 32dp; overflow: visible; }
+.asset-shell.is-narrow .asset-inspector, .asset-shell.is-compact .asset-inspector { left: 0; right: 0; bottom: 0; width: auto; overflow: hidden; }
 .asset-shell.is-narrow .inspector-resize-handle, .asset-shell.is-compact .inspector-resize-handle { display: none; }
-.asset-shell.is-narrow .inspector-strip, .asset-shell.is-compact .inspector-strip { display: flex; }
-.asset-shell.is-narrow .asset-inspector-content, .asset-shell.is-compact .asset-inspector-content { position: absolute; left: 0; right: 0; bottom: 32dp; height: 200dp; display: none; z-index: 3; }
-.asset-shell.is-narrow.inspector-expanded .asset-inspector-content, .asset-shell.is-compact.inspector-expanded .asset-inspector-content { display: block; }
+.asset-shell.is-narrow .inspector-strip, .asset-shell.is-compact .inspector-strip { display: none; }
```
MrNeRF: Nit: with the strip now hidden in narrow and compact too, and already `display: none` by default, `.inspector-strip` is never visible in any breakpoint — but the markup and Python wiring are all still there (`toggle_inspector` and `on_strip_action` bindings in the RML, the handler registration and dispatch in the panel). Nothing queries or measures it, so it is just unreachable code now.

Same for `inspector_reserved_height`: still bound and still dirtied in two places, but the RML no longer consumes it.

Worth removing both while the context is fresh.

MrNeRF (changes requested):
Went through this on Linux against current master (merges clean). The layout work is good and the responsiveness idea is right — the Inspector as an overlay makes the medium breakpoint much less cramped, and the RCSS side is done properly, no hard-coded colours, everything through theme tokens. The relink notice reordering is also correct: moving it below the identity reset is exactly why the notice used to vanish.

Four things to fix, all inline. The first one is the one I would not merge without:

**The Windows trash can permanently delete files.** The flags ask for silent, no-confirmation, no-error-UI deletion, and `FOF_ALLOWUNDO` is documented as "preserve undo information, *if possible*" — it is not a promise that the file goes to the Recycle Bin. On a network share, or any volume without a bin, the file is gone for good, `SHFileOperationW` returns success, and we then drop it from the catalog. "Move to Trash" should never be able to do that.

The other three are a silent Gallery hang, a Refresh button that can do nothing, and stale Gallery card sizes on a width-only resize.

One broken test, and it is not in your diff so I could not comment on the line: `tests/python/test_gallery_manual.py:75` stubs the controller as `SimpleNamespace(refresh=lambda: ...)`, but the Refresh action now calls `refresh(force=True)`, so it dies with `TypeError: got an unexpected keyword argument 'force'`. Production is fine — it is just a stale test double. Python tier on this branch is 14 failed / 2409 passed against master's 13 / 2401, and that one test is the only new failure.

On the tests generally: the new coverage is real but it mostly asserts internal plumbing. Nothing exercises the Windows branch (the test asserts `"SHFileOperationW" in source`, which would pass even if the flags were wrong), nothing covers the forced-refresh-while-busy path, and nothing covers either account field actually arriving. Those are the three places that bit here.

No conflict with my #2178 — I trial-merged them and it is clean, banner rename and your new Inspector rules both survive. Worth knowing that this branch alone fails #2178's new no-literals gate because of the pre-existing `#20242b` it inherits; once #2178 lands that goes away, nothing for you to do.

francescofugazzi:
- preserved forced Gallery refresh requests while a refresh is already running
- published the complete account identity through the native runtime store
- kept Gallery card geometry updated during same-column resizes
- centralized selection changes so stale Inspector details close consistently
- added an explicit warning before any permanent-delete fallback on Windows
- updated the stale Gallery refresh test double and added focused coverage for the reviewed paths

Issue #2179, opened 2026-09-17
Add triple dots menu indicator in list view


kmeirlaen:
The grid view has a triple dot menu making it easy to access the context sensitive menu.
In the list view, this menu indicator is not present.  Add a menu indicator so it is clear to the user that actions can be performed on the project file

<img width="460" height="583" alt="Image" src="https://github.com/user-attachments/assets/59e63826-ab26-4387-a706-38a88bc4979a" />

<img width="921" height="354" alt="Image" src="https://github.com/user-attachments/assets/54e10c46-0257-421b-902d-4e0231b711f2" />

kmeirlaen:
verified in f87d90893
The list is visible in mouse over + for the active project.  I would have preferred it to be always visible, but maybe that would have been too much clutter - accepting and closing

Issue #2169, opened 2026-09-17
When syncing gallery changes, mouse over displays native json


kmeirlaen:
How to simulate:

* in LFS, publish a splat as private
* in portal, change splat to public
* in LFS, there is an indication "changes in gallery"
* right click the project and choose "apply gallery changes"
* popup shows
* when hovering mouse over "Mine: private / Gallery: public" -> hover popup -> OK
* when hovering mouse over "Different settings: Camera" -> JSON is displayed

<img width="881" height="446" alt="Image" src="https://github.com/user-attachments/assets/d59092e3-638d-44f9-b824-1ac99e05c7a4" />

kmeirlaen:
Tested in f87d90893 - resolved

PR #2181, opened 2026-09-18
Fix project list actions and gallery conflict tooltips

francescofugazzi:
## Summary

- add the existing project action menu to list-view rows
- keep the action menu in a dedicated rightmost column at every responsive width
- reserve its width in the layout calculation to prevent clipping or data columns overflowing the panel
- replace raw gallery conflict payload tooltips with the existing human-readable difference summary
- add focused regression coverage for markup, responsive geometry, and conflict tooltips

## Related issues

Related to #2169
Related to #2179


MrNeRF (commented):
Looked at this on Linux. It still merges clean now that #2177 has landed, and the suite is green on that merge (331 passed), so no rebase work needed.

This is a good small PR. Both issues are real, the fixes are the obvious ones, and the tooltip change is exactly right — `part.values` was the raw payload, `part.difference` is the summary that is already built for the inline text, and I checked that `conflict_groups` sets it as a non-empty string on every branch it generates (text, visibility, view settings, tracks, content). No blocker from me.

Three things, none of them merge-blocking:

**Keyboard activation of the new button does the wrong thing.** Tab to the More button on a list row and press Enter and you open the project instead of the menu; Space opens Quick Look. `_on_asset_results_keydown` is bound on the scroll container, the keydown bubbles up from the focused button, and the handler acts on the selection and stops propagation before RmlUi can turn it into the button's click. Driving the production handler with the new markup:

```
MENU ENTER: {'loaded': ['A'], 'menus': 0, 'propagation_stopped': True}
MENU SPACE: {'quick_look': ['A'], 'menus': 0, 'propagation_stopped': True}
```

This is not your bug — the grid card menu has the same problem today — but this PR is what puts the control in front of keyboard users in list view, so it is worth fixing here. Inline.

**The width budget has no slack at the bottom end.** Detail inline; it is an edge case and I could not show it is reachable, so treat it as a note rather than a request.

**One colour literal landed in the layout sheet.** `border-color: transparent` at `asset_manager.rcss:1069`, duplicating the rule you correctly put in the theme file. The gate in `test_menubar_resources.py` only bans hex and `rgb()/rgba()`, so it does not fail, and there are ten pre-existing `transparent` declarations in that file already — but the new one is redundant, the theme rule covers it.

One thing I noticed while reading the menu path, definitely pre-existing and not something you introduced: `_handle_asset_context_action` ignores the return value of `_select_asset_id`. Open a row's menu, let a catalog refresh remove that project, then pick an action — the selection never moves and the command runs against whatever was selected before:

```
REFRESH REMOVES A: {'clicked_menu': 'A', 'selection_before': 'B', 'dispatched': [('publish', 'B')]}
```

Row recycling itself is handled correctly — the callback captures the id, not the element, so a recycled row still acts on the right project. It is only the removed-project case.

The new tests are genuinely change-sensitive; I reverted the production files in a scratch copy and all ten cases fail, so they are not just restating the implementation.

src/visualizer/gui/rmlui/resources/asset_manager.rml
```diff
@@ -259,6 +260,10 @@
                       <span class="asset-col asset-col-size" data-class-hidden="!asset_list_show_size" data-style-width="asset_list_size_width" data-attr-title="asset.size_label">{{asset.size_label}}</span>
                       <span class="asset-col asset-col-modified" data-class-hidden="!asset_list_wide" data-style-width="asset_list_modified_width" data-attr-title="asset.saved_label">{{asset.saved_label}}</span>
                       <span class="asset-col asset-col-folder" data-class-hidden="!asset_list_show_folder" data-style-width="asset_list_folder_width" data-attr-title="asset.folder_name">{{asset.folder_name}}</span>
+                      <button class="asset-button asset-button--small-icon asset-list-menu" data-attr-data-asset-id="asset.id"
```
MrNeRF, line 263: Adding the button here makes it focusable (`components.rcss` makes buttons tabbable), and that runs into the results-list keyboard handler.

`_on_asset_results_keydown` is bound on `#asset-gallery-scroll`, so a keydown on the focused More button bubbles up to it. `KI_SPACE` calls `open_quick_look()` and `KI_RETURN` calls `_load_asset(...)`, both followed by `_stop_event(event)`. RmlUi generates a focused button's click from Enter/Space as a *default action*, and `EventDispatcher` skips default actions once propagation has been stopped — so the menu can never open from the keyboard, and the key instead does something entirely different to the row.

Executed against the production handler with this markup shape:

```
MENU ENTER: {'loaded': ['A'], 'menus': 0, 'propagation_stopped': True}
MENU SPACE: {'quick_look': ['A'], 'menus': 0, 'propagation_stopped': True}
```

The fix is for that handler to leave keydowns alone when the event target is an action button rather than the row, so the button gets its normal activation. If you do that, note the menu will also need anchoring to the button — the shared menu helper positions at the current mouse position, which is meaningless for a keyboard-opened menu.

The same weakness already affects the gallery card menu, so a fix in the handler covers both.

src/python/lfs_plugins/asset_layout.py
```diff
@@ -183,8 +184,11 @@ def list_column_widths(width, overrides=None, measured=None):
     for key, threshold in (("size", 360), ("modified", 560), ("folder", 700)):
         if width < threshold:
             widths[key] = 0.0
-    # 24 dp shell inset, 16 dp row inset, thumbnail 32 dp, and its 8 dp gap.
-    available = max(0.0, float(width) - 24.0 - 16.0 - 32.0 - 8.0)
+    # Reserve shell/row insets, the thumbnail and gap, and the row action column.
+    available = max(
+        0.0,
+        float(width) - 24.0 - 16.0 - 32.0 - 8.0 - LIST_ACTION_COLUMN_WIDTH,
```
MrNeRF, line 190: Note rather than a request, since I could not show this is reachable.

Reserving the 32dp is the right call and the numbers line up with the RCSS (24dp button plus 4dp margins, matching the 32dp header spacer). But clamping at zero only hides the shortfall — the fixed items do not shrink, so the row still needs 24 + 16 + 32 + 8 + 32 = 112dp of chrome plus the compact Gallery column's 32dp before any data column exists. Running the production function:

```
width   name  gallery  total+chrome
   80    0.0     32.0         144.0
  128    0.0     32.0         144.0
  144    0.0     32.0         144.0
  150    6.0     32.0         150.0
  260  116.0     32.0         260.0
```

Below 144dp the model claims 144 whatever the width is, and since the menu is the rightmost item under `overflow-x: hidden` it is the thing that gets clipped. Before this PR the same row fit at 128dp.

The floating popup has `min-width: 260dp`, so I do not think you can actually get there that way, and I could not find a docked minimum to check against — hence a note. If you want to close it properly, the narrow fallback needs something that actually fits (collapse the Gallery column first, or enforce a minimum viewport width). The new test only samples 260dp and up and re-derives the implementation's own arithmetic, so it would not catch a change here either way.

src/visualizer/gui/rmlui/resources/asset_manager.rcss
```diff
@@ -1065,6 +1065,10 @@
 .asset-list-row { position: relative; height: 40dp; min-height: 40dp; padding: 0 8dp; border-width: 0; border-bottom-width: 1dp; }
 .asset-list-header .asset-col { white-space: nowrap; overflow: hidden; text-overflow: ellipsis; }
 .asset-list-header-spacer { visibility: hidden; }
+.asset-list-menu-spacer { flex: 0 0 32dp; width: 32dp; min-width: 32dp; visibility: hidden; }
+.asset-list-menu { flex: 0 0 24dp; width: 24dp; min-width: 24dp; visibility: hidden; margin: 0 4dp; border-color: transparent; }
```
MrNeRF: `border-color: transparent` is a colour literal in the layout sheet, and you already have the identical declaration in `asset_manager.theme.rcss` where it belongs. Drop it from here.

The palette gate in `test_menubar_resources.py::test_asset_manager_palette_is_fully_theme_driven` only rejects hex and `rgb()/rgba()`, so this passes — and there are already ten `transparent` declarations in this file — but no reason to add an eleventh when the theme rule covers it.

francescofugazzi:
Action-button key events are now ignored by both delegated row handlers, preserving native Enter/Space activation without triggering project open or Quick Look. Also included the requested menu anchoring, stale-selection guard, and style cleanup.

Issue #2171, opened 2026-09-17
Error when restoring to save point


kmeirlaen:
When restoring to certain points of a save step, an error occurs.

<img width="2544" height="1344" alt="Image" src="https://github.com/user-attachments/assets/c6c5de19-cfd8-4807-831a-77a64af82186" />

[error_restore.log](https://github.com/user-attachments/files/32346687/error_restore.log)

PR #2182, opened 2026-09-18
Fix restoring saves with retained project previews

francescofugazzi:
## Summary

- restore only the thumbnail referenced by the selected save's preview locator
- preserve additional retained `THMB` chunks without resolving them repeatedly through the same writer key
- support both in-place restores and restores to a new project file
- prevent `The project chunk was resolved more than once` when restoring affected save points
- add native regression coverage for projects containing multiple retained thumbnail chunks
- make the project operations Python tests use the checked-in portable fixture
- handle Windows file-locking and symlink capability differences explicitly
- retain strict expected-failure coverage for the existing Windows CJK-path limitation

## Related issue

Related to #2171

## Test results

- Project operations Python suite on Windows: 10 passed, 34 skipped, 2 expected failures
- Added native regression coverage for in-place and copied restore paths


francescofugazzi:
Fixed the Ubuntu Debug failure in 2bc5b647a. Historical generations do not expose a preview locator, so restore now resolves exactly one preview from the retained logical/canonical key while preserving the other THMB chunks verbatim. The regression now covers restoring an older generation after a newer preview was saved.

tests/python/test_project_operations.py
```diff
@@ -235,7 +253,7 @@ def _fixture() -> Path:
         candidate = parent / ".codex_tmp/am_concept/testhome/projects/bonsai.licht"
         if candidate.is_file():
             return candidate
-    return Path(__file__).parents[2] / "missing-project.licht"
+    return Path(__file__).parents[1] / "data" / "portable-sog.licht"
```
MrNeRF, line 269: This line is why two tests are now red.

The old value pointed at a path that never exists, so everything using `_fixture()` skipped. Pointing it at the checked-in fixture un-skips eight test functions, and `test_contents_kill_rolls_back_on_restart` does not survive it:

```
FAILED tests/python/test_project_operations.py::test_contents_kill_rolls_back_on_restart[running]
FAILED tests/python/test_project_operations.py::test_contents_kill_rolls_back_on_restart[completed]
=================== 2 failed, 42 passed, 2 skipped in 1.05s ====================
```

with

```
ModuleNotFoundError: No module named 'lichtfeld'
assert 1 == 37
```

That test spawns `subprocess.run([sys.executable, '-c', code, ...])`, and the child needs both `lichtfeld` and `lfs_plugins`. The parent only has them because `tests/python/conftest.py` does `sys.path.insert` in-process; nothing sets `PYTHONPATH`, and `tests/CMakeLists.txt:1059-1069` invokes pytest without one either. So the child cannot import and exits 1 instead of the deliberate 37. Master is 35 passed / 11 skipped; this branch is 42 / 2 / **2 failed**. Reproduced in an ordinary worktree with its own build directory, no cross-worktree setup.

Passing the import paths to the child explicitly fixes it (I confirmed both cases pass that way).

Un-skipping the other six is a good change and they all pass, so please keep the fixture — just make that child process work.

Separately, the two new `strict=True` Windows xfails below accept *any* failure as the expected CJK-path bug, including a broken identity guard or a bad rollback. If the Windows failure really is CJK path handling, a narrow test for that is better than marking two safety tests as expected-to-fail. And `_skip_locked_identity_swap_on_windows` removes fifteen identity-refusal assertions on Windows on the grounds that an identity handle is held open, but `ProjectPathIdentity::capture` reads through a local `NativeFile` and `NativeFile::open_read` uses `FILE_SHARE_DELETE` — so the justification does not obviously hold.

src/io/project/project_operations.cpp
```diff
@@ -1236,6 +1236,55 @@ namespace lfs::io::project {
...
+            const auto& locator = *reader->preview();
+            const auto current_preview = std::ranges::find_if(
+                reader->chunks(), [&locator](const ChunkInfo& row) {
+                    return row.key.fourcc == FOURCC_THMB &&
+                           row.payload_offset == locator.offset &&
+                           row.stored_bytes == locator.bytes;
+                });
+            if (current_preview != reader->chunks().end()) {
+                append_preview_key = current_preview->key;
+                if (generation == reader->commit().generation) {
```
MrNeRF: This generation guard means branch (a) is skipped for exactly the case the PR is about.

Restoring an *older* save has `generation != reader->commit().generation` by definition, so the locator-span match never runs and selection falls through to (b) "the current preview's key", (c) the canonical project-UUID key, or (d) "the only `THMB`". None of those is evidence that the same chunk was the selected preview back then — `ProjectReader::open_generation` clears the historical preview locator, so the old selection is simply not consulted.

Built this branch and ran three inputs that all pass `verify_project_file`, each restoring generation 2 whose `read_preview` is `OLD SELECTED`:

```
wrong:      COPY -> b'RETAINED NOT SELECTED'          INPLACE -> The project chunk was resolved more than once.
missing:    COPY -> This project generation has no preview.  INPLACE -> The project chunk was resolved more than once.
unresolved: COPY -> b'OLD SELECTED'                   INPLACE -> The save has unresolved project chapters.
```

So on legal containers this can restore a thumbnail that was never the one selected, or drop the preview entirely — branch (d) at line 1271 deliberately nulls the selection when there is more than one candidate, and then no `set_preview` is called at all and the restored project silently has no preview even though the save it came from had one.

Two of those three still fail in place with the original error, which is the path the reporter used.

The honest fix is to preserve or recover the historical selection rather than guess from current state. If that is not reachable from the container today, an explicit ambiguity error beats silently publishing the wrong image or none.

src/io/project/project_operations.cpp
```diff
@@ -1309,7 +1358,9 @@ namespace lfs::io::project {
                 }
             }
             for (const auto& row : reader->chunks()) {
-                if (row.is_live() && !selected_rows.contains(row.key)) {
+                if (row.is_live() && !selected_rows.contains(row.key) &&
```
MrNeRF: This exemption is right for the common case and incomplete for one shape.

When `set_preview` rewrites the current preview key, erasing it afterwards would re-trigger the very "resolved more than once" error, so skipping it is correct.

But consider: generation 2 holds only preview A; generation 3 keeps A live and selects a new preview P. Restoring generation 2 in place picks A through fallback (d), and `writer.set_preview(A_bytes)` writes it under **P**, because `ProjectWriter::set_preview` follows the current head's key. A itself is then never copied — and it is not erased either, because `selected_rows.contains(A)` is true and the exemption only covers P. A stays in the append writer's unresolved set and `commit()` rejects the save:

```
unresolved: INPLACE -> The save has unresolved project chapters.
```

The copy destination has no prior unresolved set, so it does not hit this. When the selected preview's key differs from the writer's destination key, the source row needs resolving too.

tests/test_project_operations.cpp
```diff
@@ -66,6 +66,81 @@ namespace {
...
+            for (const auto& row : reader.chunks()) {
+                if (row.is_live() && !is_current_preview(row))
+                    planned_bytes += row.stored_bytes;
+            }
+            require_status(writer.preflight(planned_bytes));
+            for (const auto& row : reader.chunks()) {
+                if (row.is_live() && !is_current_preview(row))
+                    require_status(writer.copy_chunk_verbatim(reader, row));
+            }
+            require_status(writer.set_preview(newer_preview));
```
MrNeRF: This block is where I would push back on the second commit.

The first version of this test just called `set_project_preview(source, newer_preview)`, and it failed in setup with `The save has unresolved project chapters.` — Ubuntu CI caught it on `2bc5b647a` and I reproduced it locally under gdb, the throw comes from `ProjectDocument::save` inside `set_project_preview`, before either restore is reached.

Replacing the call with a hand-rolled append that resolves the retained row does make the test green, but it removes the symptom rather than the cause. On a build of this branch:

```
  set_project_preview (Update Thumbnail): FAILED -> The save has unresolved project chapters.
  set_project_title (ordinary contents edit): OK
  set_project_license: OK
  compact_project_file: OK
```

`ProjectDocument::save` resolves the current preview through `writer.set_preview` and then `continue`s past every `THMB` source row, so the retained one is never resolved and `ProjectWriter::commit` refuses. Which means the projects affected by #2171 also cannot have their thumbnail updated — and after this commit nothing in the suite says so.

Either fix that alongside the restore fix, or keep the original `set_project_preview` call in a separate test marked as the known failure it is, so it is on the record.

MrNeRF (changes requested):
Spent a while on this one because it writes to people's project files. Built it on Linux and ran the native suite plus the Python operations suite. Short version: **the production change is right and it does fix #2171** — but the PR is not mergeable as it stands, and it exposes a second bug next to the one it fixes.

**The fix works, and I can prove it.** I built a project with two live `THMB` rows and reproduced the user's exact error on master, both destinations:

```
master-restore-proof.licht verify VERIFIED generation 3 preview b'new'
master restore copy FAILED: The project chunk was resolved more than once.
master restore in-place FAILED: The project chunk was resolved more than once.
```

Same fixture against a build of this branch: both succeed. The new gtest is honest red/green too — reverting only `project_operations.cpp` to master makes it fail with `AlreadyExists ... The project chunk was resolved more than once`, which is the user's error verbatim. That is good coverage.

**Two tests are red.** `tests/python/test_project_operations.py` goes from 35 passed / 11 skipped on master to 42 passed / 2 skipped / **2 failed** on this branch. That is a direct consequence of the `_fixture()` change and it reproduces in a plain single-worktree build with no environment tricks. Comment inline.

**The bug next to the bug.** Your first push had the test calling `set_project_preview` to create the newer preview, and it failed in setup with `The save has unresolved project chapters.` — Ubuntu CI caught it on `2bc5b647a`. The new commit replaces that call with a hand-rolled `ProjectWriter::append` block, which makes the test pass. But the reason the original line failed is a real production defect: `set_project_preview` cannot write to a project that carries a second retained `THMB`. On a build of this branch:

```
verify: VERIFIED gen 2 preview b'png'
  set_project_preview (Update Thumbnail): FAILED -> The save has unresolved project chapters.
  set_project_title (ordinary contents edit): OK
  set_project_license: OK
  compact_project_file: OK
  reduce_size drop_thumbnail: OK
```

So the users who hit #2171 also cannot set a thumbnail, and the PR now routes around the only place that surfaced it. `ProjectDocument::save` resolves the current preview through `writer.set_preview` and then skips every `THMB` source row, leaving the retained one unresolved for `ProjectWriter::commit`. Worth fixing here or filing as a follow-up, but not worth hiding.

**In-place restore still fails on other legal shapes.** Against a build of this branch, restoring generation 2 (whose preview was `OLD SELECTED`):

```
wrong:      COPY -> b'RETAINED NOT SELECTED'          INPLACE -> The project chunk was resolved more than once.
missing:    COPY -> This project generation has no preview.  INPLACE -> The project chunk was resolved more than once.
unresolved: COPY -> b'OLD SELECTED'                   INPLACE -> The save has unresolved project chapters.
```

All three inputs pass `verify_project_file` and `read_preview` returns the right bytes before the restore. In-place is the path the user actually took — the log in #2171 is `restore_save(path, row["generation"], path)` — so the same error is still reachable, and the copy path can now publish the *wrong* thumbnail or refuse outright. Details inline on the fallback chain.

**One question that I think should be answered before this merges:** how does a project end up with two live `THMB` rows in the first place? `ProjectWriter::set_preview` deliberately reuses the prior preview's key, and I could not produce a second one through any public API sequence I tried — preview changes, title edits, `reduce_size` with `drop_thumbnail`, `undo_contents_removal`, restore, compact — it is always one key. Since the fix is shape-specific, and some shapes still fail, it matters whether the reporter's file matches the shape you fixed. Asking them for the `.licht` (or its chunk directory) would settle it.

Two smaller notes: the same `set_preview`-per-`THMB` pattern is still in `repair_project` (`project_operations.cpp:2471-2483`), so Repair on an affected project should hit the identical error. And `tests/test_project_operations.cpp` lives in `LICHT_FORMAT_NON_WINDOWS_FILES`, so the new native test never runs on Windows CI.

francescofugazzi:
Addressed the review findings.

Historical preview selection is now preserved from matching published heads rather than inferred from the current preview key or THMB count. Restore keeps the selected preview key and locator for both copy and in-place destinations, while ambiguous historical selections fail explicitly instead of publishing the wrong preview.

Repair now recovers a damaged-head preview selection only when the authoritative head fields can be validated against the scanned commit and the locator matches a live THMB row. Otherwise it reports an explicit ambiguity error without creating a destination.

The retained-THMB path through set_project_preview is covered and no longer leaves unresolved chunks. Windows identity-swap coverage no longer uses broad skips, and project readers are released before callbacks and atomic recovery replacement.

Windows Python suite: 25 passed, 21 skipped, 1 xfailed. The skips are symlink/fixture availability cases; the xfail is the separate known CJK-path limitation.

MrNeRF (commented):
Re-reviewed at `8dcfc0fa8`. Almost everything from the last round is properly fixed, and the native work goes deeper than I asked for — you found the actual root cause rather than patching around it. One thing is still red.

## What's fixed

**The two failing Python tests.** `test_contents_kill_rolls_back_on_restart` now passes the parent's `sys.path` to the child as `PYTHONPATH`, so the spawned process can import `lichtfeld`. The whole file is green here: **45 passed, 2 skipped, 0 failed**, up from 42/2/2-failed.

**`set_project_preview` on a project with a retained thumbnail.** `project_document.cpp` now skips only the *rewritten* preview key instead of every `THMB` row, so the case the previous commit routed around by hand-rolling a writer actually works now. That was the bug I was most concerned about, because it meant the users who hit #2171 also could not set a thumbnail.

**Historical preview selection.** `ProjectReader::open_generation` recovering the published head instead of discarding its preview locator is the right fix. My review suggested preserving or recovering the selection and you did exactly that, which removes the guessing chain rather than adding another fallback to it.

**The Windows coverage.** `_skip_locked_identity_swap_on_windows` is gone and `_overwrite_file` changes identity in place, so the fifteen identity-refusal assertions run on Windows instead of being skipped. The remaining `strict` xfail is now one narrow test that genuinely is about CJK paths, which is what I was asking for — the ASCII rename in the `identity_project` fixture is the right call too.

## What's still red

`ProjectOperations.RestoreSaveWithRetainedThumbnailChunks` fails, and not only on my machine — `build-ubuntu (Debug)` on this PR has the identical assertion:

```
tests/test_project_operations.cpp:162: Failure
Expected equality of these values:
  require_result(repaired.read_preview())
    Which is: { 'p', 'n', 'g' }
  newer_preview
    Which is: { 'n', 'e', 'w' }
```

The restore paths pass. It is the **repair** path: after both head slots are damaged, the recovered project comes back with the generation-2 preview instead of the current one. Comment inline on the two places I would look.

Everything else in the suite is unaffected — 6 passed, 2 skipped, 1 failed, and the two skips are the usual fixture-gated ones.

## Small note

The PR body still describes the Windows run as "10 passed, 34 skipped, 2 expected failures", which was true of the earlier commit. Worth refreshing once this is green, since the skip count should be much lower now.

src/io/project/project_operations.cpp
```diff
@@ -961,19 +1034,23 @@ namespace lfs::io::project {
...
+                        return fail<RepairCandidate>(
                             lfs::ErrorCode::PermissionDenied, temporary,
                             "The repair staging head could not be written.", "write failed",
                             "repair.staging");
                     }
                 }
                 auto reader = ProjectReader::open(temporary);
                 if (reader && reader->verify_all()) {
-                    return std::pair{temporary, generation};
+                    return RepairCandidate{
```
MrNeRF, line 1045: This is where I would start on the repair failure.

`recovered_preview` is built by reading the preview locator out of the *damaged* head slots and cross-checking it against the candidate commit record. The test flips one byte at `HEAD_SLOT_OFFSETS[slot] + 200` in both slots, which is past the preview fields at 112-127, so the locator bytes themselves survive — but the head CRC at 4092 does not, and whether the cross-check still accepts the slot depends on what `sanitized_head` zeroes before recomputing it.

The observed result is that the repaired project ends up with the generation-2 preview (`png`) rather than the current one (`new`). So either the loop accepted a head belonging to a different generation than the candidate it settled on, or it accepted no head at all and something downstream supplied the older selection anyway.

Worth checking explicitly which generation `find_repair_candidate` returns for this fixture, and what `recovered_preview` holds at that point — the fixture is cheap to run under a debugger and it should be obvious immediately.

src/io/project/project_operations.cpp
```diff
@@ -2378,6 +2497,19 @@ namespace lfs::io::project {
                                              "The project identity changed before repair started.",
                                              "recovered source does not match the selected project", "repair.identity");
         const auto& source_commit = reader->commit();
+        if (!candidate->preview_selection_recovered) {
```
MrNeRF, line 2500: This guard is the other half of the puzzle, and the fact that it did **not** fire is itself informative.

If the damaged heads had given no trustworthy selection, `preview_selection_recovered` would be false, this branch would find the live `THMB` row and return "The recovered project preview is ambiguous." — which is the correct, safe outcome for that fixture and is what the neighbouring `ambiguous_repair_source` case checks for.

Instead the repair succeeded and published the wrong preview, so `preview_selection_recovered` came back **true** with a stale locator. That points at the recovery loop above having accepted a selection it should have rejected, rather than at this guard being too weak.

One thing worth deciding while you are in there: right now this only refuses when a live `THMB` row exists. If the selection is untrustworthy but there is no retained thumbnail, the repair proceeds with whatever was written into the reconstructed head. That is probably fine — nothing to be ambiguous between — but it would be good to know it is deliberate.

MrNeRF (approved):
Re-checked at `e43c893f9`. Everything from both rounds is resolved and CI is green including Windows. Approving.

The last commit is test-only, which I went and looked at carefully before accepting — a test-only fix for a failing assertion is usually a warning sign. It isn't here. The repair block was copying `source` **after** the in-place `restore_save(source, 2, source)` had already rewound that file's current preview to `selected_preview`. So `damaged` was a copy of an already-restored project and repair correctly recovered *its* current preview, which is exactly the `png` the assertion saw. Moving the block above the in-place restore samples the fixture in the state the assertion describes. The assertions themselves are unchanged, so nothing was weakened — production was right and the fixture was being read at the wrong point.

I re-ran the three container shapes from my first review against a build of this head. Then and now:

| shape | before | now |
|---|---|---|
| noncanonical selected preview | copy published `RETAINED NOT SELECTED` — the wrong thumbnail, silently; in-place failed with "resolved more than once" | both refuse with **"The selected save's preview is ambiguous."** |
| distinct retained key | copy refused with "no preview"; in-place failed with "resolved more than once" | both refuse with the same explicit message |
| single retained preview | copy worked; in-place failed with "unresolved project chapters" | **both succeed and return `OLD SELECTED`** |

That is the outcome I was hoping for. The silent wrong-thumbnail case is gone, the in-place failures are gone, the case that should work now works on both destinations and returns the correct historical preview, and what remains is an explicit typed refusal for containers that genuinely are ambiguous — with a test pinning that message.

Verified on this head:

- native: **117 passed, 3 skipped, 0 failed**, against 116/3 on master — one test added, nothing regressed, and the three skips are the usual fixture-gated ones
- `test_project_operations.py`: **45 passed, 2 skipped, 0 failed**, from 42/2/2-failed when I first reviewed
- still red on master's production files, with "The save has unresolved project chapters" — so it remains real red/green coverage rather than a test that passes either way

Thanks for sticking with this one. Going from a fallback chain to recovering the published head in `open_generation`, and fixing `set_project_preview` on multi-thumbnail projects along the way, made it a much better change than the one I first reviewed.

Issue #2180, opened 2026-09-17
Projects that are listed outside of the main folder dont have information


kmeirlaen:
In the project manager, recent projects are visible under the "recent" folder.
Projects that are in the main folder have a context menu + file information (size, etc..)
Projects that are outside the main folder do not have this information.
Also, when right clicking these, the context menu only shows "open" - it is not possible to view details of these files.

<img width="924" height="318" alt="Image" src="https://github.com/user-attachments/assets/bfe78f43-7539-4c2f-9dd3-b29c0f4211b5" />

kmeirlaen:
Verified in f87d90893 - closing

PR #2183, opened 2026-09-18
Show project details and actions for external recent files

francescofugazzi:
## Summary

- inspect recent projects located outside the managed library
- display native file size, saved date, preview, commit information, and Inspector details
- provide Open, Show in Folder, Contents, Export As, Update Thumbnail, and Rename actions
- omit Show in Folder when the recent file no longer exists
- keep Gallery actions disabled because temporary Recent entries have no catalog identity
- keep library-only actions unavailable
- do not register external recent projects implicitly in the managed library
- retain inspected thumbnail resources while the projects remain in the Recent scope
- add focused coverage for metadata, menus, missing files, inspection scheduling, and catalog isolation

## Related issue

Related to #2180


src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -2771,7 +2766,30 @@ def _show_shared_context_menu(
...
+                    "export_as": "projects.action.export_as",
+                    "update_thumbnail": "projects.action.update_thumbnail",
+                    "rename": "projects.action.rename",
+                }
+                for operation in operation_actions(asset):
+                    action = str(operation.get("action") or "")
+                    if action in labels:
+                        items.append({
+                            "label": tr(labels[action]),
+                            "action": "project:" + action,
```
MrNeRF, line 2847: These four actions cannot succeed. The row id reaches native identity validation as if it were a project UUID.

`_handle_asset_context_action` does `self._select_asset_id(asset_id)` and then `open_project_operation(...)`, which builds `asset = dict(self._asset_dict(asset_id))` and calls `ProjectOperations.run(operation_id, asset, ...)` with `asset["id"]` still set to `recent:<sha256(path)>`. From there:

- `backup=False` (Export As) hits `project_operations.py:64-66`: `if str(card.project_uuid) != asset["id"]: raise ValueError("The project identity changed. Refresh Projects and try again.")`. That comparison can never be true for a `recent:` id.
- `backup=True` (Rename, Update Thumbnail, Contents edits) hits `self.io.run_project_operation(path, asset["id"], ...)` at `:88`, and native rejects it with `Invalid reference UUID: recent:...`.

Both fail with `operation_called=False`, so nothing is written — but the user gets an error that reads like their project is broken, and the `backup=True` path still leaves a failed recovery row in `data/asset_library/contents.json` keyed by the fake id.

There is a related giveaway a few lines up in `open_project_operation`: `if card is not None and str(getattr(card, "project_uuid", "")) == asset_id: asset["commit_uuid"] = ...`. For a Recent row that is never true, so the commit guard is silently skipped as well.

The fix is to separate the UI row id from the validated native identity — you already have the real UUID in `self._inspection_by_asset[...]["card"].project_uuid` by the time these actions are offered, since the menu only shows them once `details` is present. Pass that (and its `commit_uuid`) through instead of the row id, and keep the guard meaningful.

Worth noting the error feedback here is invisible too: these operations have no `content_row`, so the failure is stored with `row_id=""` and `get_contents_rows` only renders feedback that matches a real row. The dialog closes and nothing is shown.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -1138,15 +1138,9 @@ def _inspect_contents(self, path):
     def _start_inspection_refresh(self) -> None:
         if not self._asset_index or not self._panel_mounted or not self._handle:
             return
-        entries = [
-            asset
-            for asset in self._window_assets(self._filtered_assets())
-            if not asset.get("recent_only")
-        ]
+        entries = self._window_assets(self._filtered_assets())
```
MrNeRF, line 1164: Dropping the `recent_only` filter here means raw Recent rows now go into the inspection pipeline, and those rows carry no stat identity.

`_recent_only_assets` builds entries without size, mtime or commit, so `project_inspector.inspection_cache_key` computes the constant `(0, 0, 0, 0, '')` for every one of them. The cache therefore never invalidates: inspect an external project, save over it from somewhere else, come back and select it again, and the old title, saved date, size, preview revision and Contents are still what you see. I reproduced it by writing version A, inspecting, replacing the file with a larger version B, rebuilding the Recent row and refreshing — only version A was ever read.

`_refresh_after_project_write` does not help either; it returns early for paths outside the mapped folders, which is precisely this case.

These rows need a real cache key. You are already stat-ing the file for the size and date columns, so feeding that into the key should be cheap.

src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -1138,15 +1138,9 @@ def _inspect_contents(self, path):
...
         selected = self.get_selected_asset_id()
-        self._ensure_inspection_pipeline().refresh(
-            entries, "" if selected.startswith("recent:") else selected
-        )
+        self._ensure_inspection_pipeline().refresh(entries, selected)
```
MrNeRF, line 1166: Two things follow from passing the selected Recent id through here.

First, a failed inspection now has nowhere to go. `_on_inspection_result` stores the exception in `self._inspection_errors[asset_id]`, and for these rows nothing ever reads that dictionary: the status stays empty, the selected health model reports `READING`, `selected_has_problem()` is False and no catalog notice appears. So a corrupt, locked or unreadable external `.licht` just sits there with no details, no actions and no explanation. I confirmed the model state by injecting a `ValueError('corrupt project')` through the real callback. Windows file locking and a dead network share both land on this same path.

Second, the preview only makes it half way. `_format_asset_for_ui` overlays `card.has_preview` / `commit_uuid` onto a temporary dictionary, so the list card renders `image(preview://kind=licht&...)`. But `_sync_info_thumbnail` and `get_selected_asset_thumbnail_decorator` read the raw Recent entry, whose `has_preview` is always False — so the Inspector and Space/Quick Look show nothing for a project whose thumbnail is visible one column to the left.

MrNeRF (changes requested):
Tried this on Linux. It merges clean on top of master now that #2177 has landed, and the panel suite is green on that merge (326 passed), so nothing here is a rebase problem.

The idea is right and #2180 is a real gap. Keeping the Gallery verbs off these rows, keeping them out of the managed library, and dropping Show in Folder when the file is gone are all the correct calls, and I checked the isolation claim specifically: inspection reads the card and details natively and does not write a `records/<uuid>.json` or a catalog entry.

The problem is that none of the new write actions work. Export As, Rename, Update Thumbnail and every Contents edit fail before they do anything, because the synthetic `recent:<sha256(path)>` row id is handed to code that requires a real project UUID. That is the headline of the PR, so it is the one blocker. Details inline.

I reproduced it end to end against a real `.licht` (a copy of `tests/data/portable-sog.licht`), driving the production `ProjectOperations.run` with exactly the asset dictionary this panel builds:

```
Export As: ValueError: The project identity changed. Refresh Projects and try again.; operation_called=False
Rename/Update Thumbnail/Contents edit: ProjectOperationFailure: Invalid reference UUID: recent:23d05de4...; operation_called=False
Source bytes unchanged: True
Recovery asset_id: recent:23d05de4...
Actual native UUID: f88e136c-4173-46f7-b5f5-3e9aeb242fc8
```

Three smaller things beside that, all inline: the preview reaches the list card but not the Inspector or Quick Look, the inspection cache can never notice that an external file changed, and a failed inspection is stored and then never shown to anyone.

On the tests: the five new/changed cases are genuinely change-sensitive — they all fail against base production, I checked that in a scratch copy. But they assert menu shape and scheduling, not that the actions actually execute, which is why the blocker slipped through. A test that runs one of these operations against a real fixture would have caught it immediately.

No conflict with #2181: it adds the More button to list rows, which routes through this same shared menu, so these actions become reachable from there too.

francescofugazzi:
Recent rows now keep their synthetic `recent:` identifier strictly for UI state. Before an operation starts, the panel takes the real project UUID and commit UUID from the completed native inspection and passes those identities to ProjectOperations. This covers both unguarded Export As and guarded Rename, Update Thumbnail, and Contents operations, without registering the external project in the managed library or writing catalog records under the synthetic ID.

External Recent rows now include a real file-stat identity (size, mtime, device, and inode), so replacing or modifying the file invalidates the inspection cache. Successful writes force a fresh inspection, while Rename no longer attempts to update a catalog entry for an unmanaged file.

Inspected preview data is now shared consistently by the list, Inspector, and Quick Look. Inspection failures mark the selected external row as UNREADABLE and expose the actual reason; operation failures without a Contents row are also surfaced instead of disappearing.

Coverage now verifies stat-based invalidation, preview and error propagation, both ProjectOperations paths (`backup=False` and `backup=True`), native project/commit identity forwarding, and a real write through the native bindings against a temporary copy of `portable-sog.licht`.

Windows panel test suite: 160 passed.

PR #2197, opened 2026-09-19
Fix stale GUT state when switching to IGS+

francescofugazzi:
## Summary

Prevent a retained GUT value in the destination IGS+ preset from surviving a GUI-driven strategy switch or reset.

Strategy presets keep independent editable values. This allowed a hidden GUT value to remain stored in the IGS+ slot and later produce a valid but irrelevant GUT/IGS+ conflict message, even when the user had not explicitly selected that combination.

The strategy boundary now normalizes the destination IGS+ preset by disabling GUT whenever IGS+ becomes active.

## Behavior preserved

- Explicit GUT/IGS+ conflicts continue to use the existing confirmation flow.
- Other strategy presets retain their independent settings.
- Project import, trainer reset, and general parameter validation are unchanged.
- The change applies cleanly to the current `dev` branch.

## Tests

- Property-system suite: 35 passed.
- Regression coverage verifies that a stale GUT value cannot survive re-entering the IGS+ preset.
- Existing strategy-default switching behavior remains covered.

Related to #1534


MrNeRF:
The standalone Python test has no ParameterManager, so it misses the stale per-strategy settings and passes without this fix. Please cover a real parameter manager too. I checked that path: switching back to IGS+ leaves GUT on before this PR and clears it afterward.

PR #2198, opened 2026-09-19
Fix Unicode filenames across Windows UI boundaries

francescofugazzi:
## Summary

Preserve Unicode filenames when filesystem paths cross UI, native-dialog, localization, and MCP boundaries.

On Windows, `std::filesystem::path::string()` uses the native narrow encoding and may corrupt or reject characters that are not representable in the active code page. These boundaries expect UTF-8 strings instead.

## Changes

- preserve the current project filename in the Save As dialog
- preserve Python script names in the Python Console Save As dialog
- display Unicode log-export filenames correctly in the success message
- safely inspect Unicode directory names while scanning project locations
- return UTF-8 Python script paths from MCP editor tools and resources

All affected paths now use the existing `path_to_utf8()` conversion helper. No dialog, persistence, or filesystem behavior is otherwise changed.

## Validation

- focused Unicode path tests passed
- project Save As verified with a Unicode filename
- Python Console Save As verified with a Unicode filename
- log export verified with a Unicode filename


MrNeRF:
This changes several Windows file-picker and path-display paths, but has no regression test for them. Please add a Windows check using a Unicode filename through the actual handlers; Linux path tests cannot catch the original conversion bug.

PR #2199, opened 2026-09-19
Preserve Unicode filenames in serialized error details

francescofugazzi:
## Summary

Preserve Unicode filenames when allowlisted filesystem paths are included in serialized error details.

On Windows, constructing a filesystem path directly from a UTF-8 `std::string` and converting its filename with `.string()` can pass through the active narrow code page. This may corrupt or reject characters that cannot be represented by that code page before the existing UTF-8 sanitizer is reached.

## Changes

- decode incoming path fields with the existing UTF-8 path helper
- extract the filename as a filesystem path
- serialize it through the UTF-8-safe path conversion
- remove the now-unused direct filesystem include
- add regression coverage for a mixed Japanese/Chinese filename

The existing privacy behavior is preserved: only the filename is exposed and parent directories remain excluded from the wire envelope.

## Validation

- complete `ErrorEnvelopeTest` suite passed
- regression test confirms Unicode filename preservation
- regression test confirms parent-directory redaction
- strict JSON serialization coverage remains passing


MrNeRF:
The Unicode error test passes both before and after this fix on Linux, and Windows CI does not run its test target. Please run it on Windows and check that it fails without the fix.

PR #2200, opened 2026-09-19
Return the default log file path as UTF-8

francescofugazzi:
## Summary

Return the resolved default log file path using the established UTF-8 path boundary.

`Logger::default_log_file_path()` previously used `std::filesystem::path::string()`. On Windows, this converts through the active narrow code page and can corrupt or reject user-directory names containing characters that are not representable by that code page.

## Changes

- serialize the resolved default log path with `path_to_utf8()`
- add cross-platform coverage using mixed Japanese, Chinese, and accented directory names
- verify both the returned UTF-8 value and its filesystem round trip

This does not change log placement, logger initialization, sinks, or console output behavior.

It completes a path-return boundary that was not covered by the Windows console Unicode work in #2042.

## Validation

- existing explicit-override test passed
- new Unicode override and round-trip test passed

Related to #2042


MrNeRF:
The new log-path test passes both before and after this fix on Linux, and Windows CI does not run its test target. Please run it on Windows and check that it fails without the fix.

PR #2202, opened 2026-09-19
Preserve Unicode filenames during drag and drop

francescofugazzi:
## Summary

Preserve Unicode filenames and extensions while classifying files received through drag and drop.

The drop handler converted filesystem components with `path::string()`. On Windows, this passes through the active narrow code page and can corrupt or reject filenames containing characters outside that code page before the file is classified.

## Changes

- convert dropped-file extensions with `path_to_utf8()`
- reuse the normalized extension for project and video detection
- convert COLMAP `.txt` and `.bin` filenames through the UTF-8-safe path boundary
- lowercase extensions and filenames using unsigned-character-safe transforms
- preserve the existing classification and command-dispatch behavior

## Regression coverage

The drag-and-drop Unicode test now covers:

- Japanese, Korean and Chinese basenames
- mixed Unicode filenames
- an uppercase `.TXT` extension
- a fully Unicode extension
- exact filename round trips
- safe lowercase normalization

The fixture now constructs paths with `utf8_to_path()`, ensuring it creates real Unicode Windows paths rather than narrow-codepage mojibake.

## Validation

- `UnicodePathTest.DragDropPathHandling` passed
- the test executes real Unicode filename and extension round trips


MrNeRF:
The new test checks the path helpers but never calls the drag-and-drop handler, so reverting the production change would still pass. Please add a Windows test that sends a Unicode filename through the actual drop handler.

PR #2204, opened 2026-09-19
Preserve Unicode paths in environment map diagnostics

francescofugazzi:
## Summary

- convert environment map paths explicitly to UTF-8 before including them in loader errors and logs
- preserve Unicode paths in CUDA allocation and upload failure diagnostics
- add coverage for successful environment map loading from a non-ASCII path
- verify that missing non-ASCII paths are reported without filename corruption

On Windows, `std::filesystem::path::string()` depends on the active narrow code page and can corrupt Unicode filenames or mask the original rendering error. This change uses the existing path conversion helpers without altering environment map loading, caching, rendering, or CUDA behavior.

## Testing

- `EnvironmentImageLoadTest.*`
- `ExportEnvCompositeTest.*`


MrNeRF:
The Unicode environment-map checks pass both before and after this fix on Linux, and Windows CI does not run their test target. Please run them on Windows and check that they fail without the fix.

PR #2205, opened 2026-09-19
Preserve Unicode filenames in preprocessing progress

francescofugazzi:
## Summary

- convert preprocessing image filenames explicitly to UTF-8
- reuse the converted filename for both terminal progress and callback reporting
- avoid Windows code-page corruption for non-ASCII image names
- leave inference, output generation and scheduling unchanged

`std::filesystem::path::string()` uses the active narrow code page on Windows and may corrupt Unicode filenames. The existing UTF-8 path helper is now used once per processed image before reporting progress.

## Testing

- `UnicodePathTest.PathToUtf8Conversion`


MrNeRF:
There is no regression test for the changed progress messages. Please check a Unicode filename through the actual preprocessing callback on Windows, with the test failing without this fix.

PR #2213, opened 2026-09-19
Export menu bar test dependencies on Windows

francescofugazzi:
## Summary

- export `RmlMenuBar` from the Windows visualizer DLL
- export `RmlTooltipController` from the same DLL
- allow the native menu bar title tests introduced by #2203 to link on Windows
- preserve the existing static-library behavior on Linux

`lfs_visualizer` is a shared library on Windows and uses hidden visibility. The new native tests call these implementations from `lichtfeld_tests.exe`, but the classes were not exported, producing unresolved external symbols during linking.

Related to #2203.


MrNeRF:
Windows CI only builds the format-test target, so it never checks the menu-bar test link this PR fixes. Please add a Windows build and run of `lichtfeld_tests`. The current Debug failure is C1128 in `scene_manager.cpp`, before that link step.

francescofugazzi:
Added Windows CI coverage for `lichtfeld_tests` in both Release and Debug and run `MenuBarTitleTest.*` explicitly. Also applied `/bigobj` only to `scene_manager.cpp` so the existing Debug C1128 failure no longer prevents CI from reaching the native test link and execution steps.

.github/workflows/windows.yml
```diff
@@ -148,7 +148,7 @@ jobs:
         shell: cmd
         run: |
           call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
-          cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
+          cmake -B %BUILD_DIR% -S . -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DLFS_FAST_COMPILE=ON -DBUILD_TESTS=ON -DBUILD_FORMAT_TESTS=ON -DBUILD_PYTHON_STUBS=OFF -DLFS_DEV_IMPORT_SOURCE_PYTHON=OFF -DLFS_DEV_IMPORT_SOURCE_RESOURCES=OFF -DLFS_ENABLE_CUDA_DEVICE_DEBUG=OFF
```
MrNeRF: My earlier request to build all of `lichtfeld_tests` was too broad. It requires LibTorch, which we should not add to this CI. Please build and run just the menu-bar tests in a small CPU-only target linked to `lfs_visualizer`, so it still checks the DLL exports.

MrNeRF (changes requested):
Please keep this CI test independent of LibTorch.

tests/CMakeLists.txt
```diff
@@ -31,6 +31,40 @@ endif()
...
+if(BUILD_VISUALIZER_TESTS)
+    set(LICHTFELD_VISUALIZER_TEST_SOURCES
+        test_menu_bar_title.cpp
+    )
+    add_executable(lichtfeld_visualizer_tests EXCLUDE_FROM_ALL
+        ${LICHTFELD_VISUALIZER_TEST_SOURCES}
+    )
+    target_compile_features(lichtfeld_visualizer_tests PRIVATE cxx_std_23)
+    target_include_directories(lichtfeld_visualizer_tests PRIVATE
+        ${CMAKE_SOURCE_DIR}/src/visualizer
```
MrNeRF: The new target fails to compile: `core/export.hpp` cannot be found. The visualizer keeps its core include paths private, so linking it does not pass those paths to this test. Giving the target the visualizer include directories fixes this locally: the target then builds and all eight tests pass without LibTorch.

MrNeRF (changes requested):
The LibTorch dependency is gone, but the new test target still needs its header paths.

tests/CMakeLists.txt
```diff
@@ -31,6 +31,40 @@ endif()
 # Find GTest
 find_package(GTest CONFIG REQUIRED)
 
+# =============================================================================
+# CPU-only visualizer tests (no LibTorch)
+# =============================================================================
+if(BUILD_VISUALIZER_TESTS)
```
MrNeRF, line 37: Turning on localization tests alongside this option silently skips the visualizer tests when full/format tests are off. I reproduced the missing target. Please also include `BUILD_VISUALIZER_TESTS` in the earlier localization return condition.

MrNeRF (commented):
One additional test-option issue.

francescofugazzi:
Updated the Windows CI test setup after reproducing the remaining issues locally.

The focused visualizer suite now:
- keeps the production lfs_visualizer DLL linkage requested by the review;
- runs without requiring an installed NVIDIA display driver by delay-loading nvcuda.dll only in the dedicated MSVC test profile;
- uses the correct Release/Debug runtime directories in CI;
- removes the ineffective DLL-copy workaround;
- fixes the tooltip-path assertion to compare the platform-native normalized UTF-8 path while preserving the markup-escaping coverage.

The executable now starts locally and reaches all 8 MenuBarTitleTest cases; the Windows-specific failure found in the final tooltip test has been corrected.

MrNeRF (approved):
Conflict resolved, duplicate Windows exports removed, and all three CPU test suites preserved. Eight menu-bar tests and 65 Unicode tests pass locally.

PR #2217, opened 2026-09-19
Preserve Unicode paths in weight file errors

francescofugazzi:
## Summary

- convert failed LichtFeld weight file paths explicitly to UTF-8
- preserve complete non-ASCII filenames in structured error details
- add a CUDA-independent regression test for a missing Unicode-named `.lfw` file
- verify that the existing `NotFound` error classification remains unchanged

On Windows, `std::filesystem::path::string()` depends on the active narrow code page and can corrupt the path included in the error or obscure the original mapping failure. The error path now uses the existing UTF-8 filesystem conversion helper.

## Testing

- `WeightFileTest.MissingUnicodePathIsReportedAsUtf8`


tests/test_nn_ops.cpp
```diff
@@ -233,6 +236,21 @@ namespace {
 
 } // namespace
 
+TEST(WeightFileTest, MissingUnicodePathIsReportedAsUtf8) {
```
MrNeRF: This test passes both before and after the fix on Linux, and Windows CI does not run its test target. Please include it in Windows CI and check that it fails without the fix and passes with it.

francescofugazzi:
Implemented the requested Windows CI coverage.

- Kept the original regression test exercising the real `WeightFile::open()` path.
- Moved the test into a shared source used by both the full test suite and the existing CPU-only `unicode_path_test` target.
- Split the CPU-only weight-file opening/parser implementation into its own translation unit so the regression can be linked without the CUDA runtime.
- Added a static module-interface mode for standalone tests, avoiding MSVC exporting and pulling in the complete CUDA-backed Tensor surface.
- Windows CI now builds `unicode_path_test` and runs `WeightFileTest.MissingUnicodePathIsReportedAsUtf8` explicitly.
- Preserved `lichtfeld_format_tests` when Unicode and format coverage are enabled together.

PR #2219, opened 2026-09-19
Preserve Unicode filenames in converter overwrite prompts

francescofugazzi:
## Summary

- convert output filenames explicitly to UTF-8 before displaying the interactive overwrite prompt
- prevent Windows narrow-code-page conversion from corrupting non-ASCII names
- add a regression test using an existing `出力_é.ply` output
- verify that declining the overwrite displays the complete filename and records the file as skipped

The converter previously used `std::filesystem::path::string()` for the prompt. On Windows this depends on the active narrow code page and can corrupt Unicode filenames or raise a conversion error before the user can answer.

The conversion pipeline and overwrite behavior are otherwise unchanged.

## Testing

- `ConverterLichtTest.OverwritePromptPreservesUnicodeFilename`


tests/test_converter_licht.cpp
```diff
@@ -294,4 +312,31 @@ namespace {
         EXPECT_EQ(loaded->value.size(), kCount);
     }
 
+    TEST_F(ConverterLichtTest, OverwritePromptPreservesUnicodeFilename) {
```
MrNeRF, line 315: This test passes both before and after the fix on Linux, and Windows CI does not run its test target. Please include it in Windows CI and check that it fails without the fix and passes with it.

francescofugazzi:

The real converter overwrite prompt has been extracted into a small CPU-only component, while preserving the existing end-to-end converter regression test. A focused regression test now exercises the same production prompt implementation with a non-ASCII output filename.

The focused test is registered with `unicode_path_test`, so it is executed by the complete CPU-only Unicode suite on Windows CI. No additional workflow change was needed because PR #2222 already made Windows CI run the entire Unicode test executable.

PR #2222, opened 2026-09-19
Preserve Unicode paths in crash diagnostics

francescofugazzi:
## Summary

- convert the installed crash log path explicitly to UTF-8 before printing it
- preserve native crash-log opening (`CreateFileW` on Windows)
- keep allocation and conversion outside signal and exception handlers
- add a child-process regression test using a Unicode `LFS_HOME`
- restore the test environment exactly after execution

`std::filesystem::path::string()` depends on the active narrow code page on Windows and can corrupt the diagnostic path or throw while crash handlers are being installed. The path is now converted through the existing UTF-8 helper before it is written to stderr.

## Testing

- `CrashHandlerTest.*` — 8 tests passed on Windows


tests/test_crash_handler.cpp
```diff
@@ -62,6 +92,24 @@ TEST(CrashHandlerTest, ExceptionFirewallPropagatesNormalReturnValue) {
     EXPECT_EQ(result, 7);
 }
 
+TEST(CrashHandlerTest, InstallationReportsUnicodeCrashLogPathAsUtf8) {
```
MrNeRF, line 95: This test passes both before and after the fix on Linux, and Windows CI does not run its test target. Please include it in Windows CI and check that it fails without the fix and passes with it.

francescofugazzi:

The CPU-only Unicode test target now includes the existing crash-handler regression test and its real supporting implementation, without linking the CUDA-dependent teardown code.

Windows CI now runs the entire `unicode_path_test` suite. This also corrects the oversight introduced in PR #2217, where execution was unintentionally restricted to `WeightFileTest.MissingUnicodePathIsReportedAsUtf8`.

As a result, both the coverage added by #2217 and `CrashHandlerTest.InstallationReportsUnicodeCrashLogPathAsUtf8` from this PR are now executed on Windows CI.

MrNeRF (approved):
Fixed the missing GCC stacktrace dependency. The build failed before the fix; all 65 tests pass afterward. Windows Release also passed all 65 tests before this GCC-only change.

PR #2223, opened 2026-09-19
[OPTIONAL] Fix TinyUSDZ compilation with GCC 15 on dev 

shadygm:
## Issue
GCC 15 cannot compile TinyUSDZ's fixed-width integer types without `<cstdint>`.

## Change
Add the missing include. Dev counterpart of #2221.

## Verification
Full Release build passed on the identical dev source tree. Wait for CI build checks before merging.


MrNeRF:
I simply merge your fix from master into dev. This way it would produce merge conflicts. I close. Thanks anyway!

PR #2224, opened 2026-09-19
Preserve Unicode paths throughout project save and recovery

francescofugazzi:
## Summary

- preserve native filesystem paths while creating Save As staging files
- make project writer, preview, and recovery diagnostics explicitly UTF-8
- preserve Unicode names while discovering and cleaning recovery artifacts
- construct recovery session, sidecar, temporary, lock, and quarantine paths without using the Windows narrow code page
- retain native path comparisons where no display conversion is required
- add an end-to-end regression test covering Unicode source and destination project names

On Windows, converting a `std::filesystem::path` through `.string()` can throw or corrupt names that are not representable in the active code page. This affected project opening, Save As, compaction, validation, preview selection, and recovery cleanup.

The project path remains native for filesystem operations and is converted explicitly only when a UTF-8 diagnostic or comparison key is required.

## Testing

- `ProjectDocumentTest.SaveAsPreservesUnicodeSourceAndDestinationPaths`
- Save As and recovery regression tests

All tests passed on Windows.


tests/test_project_document.cpp
```diff
@@ -4672,6 +4673,38 @@ namespace {
             << lfs::format_for_developer(appended.error());
     }
 
+    TEST(ProjectDocumentTest, SaveAsPreservesUnicodeSourceAndDestinationPaths) {
```
MrNeRF, line 4676: This test passes both before and after the fix on Linux, and Windows CI does not run its test target. Please include it in Windows CI and check that it fails without the fix and passes with it.

francescofugazzi:

The original end-to-end `ProjectDocument` regression test remains in place. Since `test_project_document.cpp` is intentionally excluded from the Windows CPU-only format target because it pulls headless recovery, training, and visualizer state dependencies, the actual Save As staging-path construction has been extracted into a small internal helper used directly by `ProjectDocument::save_as`.

A focused regression test now exercises that same production helper with a non-ASCII destination filename and is registered with `unicode_path_test`, which Windows CI already runs in full after PR #2222.

PR #2225, opened 2026-09-19
Fix native Wayland startup, scaling, and window handling

shadygm:
## Issue
Native Wayland startup lacked Vulkan surface support; fractional scaling left UI layout and pointer coordinates mismatched.

## Change
Enable Wayland Vulkan support, align UI/input in framebuffer pixels, and use native window-state and drag-hover handling.

## Verify
Release build and native Wayland resize/move/menu checks passed at 125% scaling before porting to dev. The dev port passes 13 C++ syntax checks and 11 input tests; await CI.

Omarchy suppresses native maximize requests by default, so maximize behavior needs verification on a desktop that permits them.


MrNeRF:
Native Wayland startup and rendering worked here with a real scene. Please also add a regression test for the mouse-coordinate conversion: my 1.5x and 2x checks fail before this change and pass afterward. There is no committed test covering that conversion.

shadygm:
Thanks for the review @MrNeRF, fix is in place.

MrNeRF (approved):
Coordinate regression checks fail before the fix and pass with it. The additional event-position check and CI pass.

PR #2227, opened 2026-09-19
Preserve Unicode filenames in converter outputs

francescofugazzi:
## Summary

- preserve native filesystem paths when deriving converter output names
- fix default SSOG destinations for Unicode input filenames
- fix multi-file SSOG output names containing Unicode characters
- preserve Unicode stems for converted outputs in other formats
- convert filenames explicitly to UTF-8 only when displaying overwrite prompts
- add end-to-end coverage using Japanese, Chinese, and Cyrillic filenames

On Windows, deriving output names through `std::filesystem::path::string()` depends on the active narrow code page. Filenames containing characters outside that code page could therefore be corrupted or cause conversion to throw.

Output names now remain native `std::filesystem::path` values throughout filesystem operations.

## Testing

- `SsogFormat.ConverterPreservesUnicodeOutputNames`
- `SsogFormat.ConvertBundleDirectoryDefaultAndBack`


tests/test_ssog_format.cpp
```diff
@@ -712,3 +714,70 @@ TEST(SsogFormat, ConvertBundleDirectoryDefaultAndBack) {
     ASSERT_TRUE(loaded) << loaded.error().format();
     EXPECT_EQ(std::get<std::shared_ptr<SplatData>>(loaded->data)->size(), 128);
 }
+
+TEST(SsogFormat, ConverterPreservesUnicodeOutputNames) {
```
MrNeRF, line 718: This test passes both before and after the fix on Linux, and Windows CI does not run its test target. Please include it in Windows CI and check that it fails without the fix and passes with it.

francescofugazzi:
The converter’s output-path derivation has been extracted into a small CPU-only component used directly by the production converter. A focused regression test now exercises Unicode filenames for:

- default SSOG outputs;
- generic `_converted` outputs;
- batch SSOG outputs.

The test is registered with `unicode_path_test`, which Windows CI already executes in full after PR #2222. The original end-to-end SSOG converter test remains in place.

Issue #2228, opened 2026-09-19
Image thumbnail update not immediate visible


kmeirlaen:
After updating a thumbnail, the new image is not immediately visible in the project manager

How to simulate:
* open LFS
* drag dataset into LFS
* create new project
* save project
* close LFS
* open LFS
* open project manager
* open previously saved project
* right click on the previously saved project, choose "update thumbnail"
* choose "image file"
* select an image 
* wait for thumbnail image to update
* --> image thumbnail does not automatically update

Expected:
* thumbnail is updated after right clicking the project or restarting LFS
* expected: thumbnail is updated after selecting it (no other user interaction required)

kmeirlaen:
verified in e82b6326a - OK - closing

PR #2230, opened 2026-09-19
Refresh project thumbnails immediately after updates

francescofugazzi:
## Summary

Refresh Project Manager thumbnails immediately after an explicit thumbnail update.

## Changes

- Reverify the affected catalog asset after updating its thumbnail from:
  - an image file
  - the current viewport
  - the first dataset image
  - the first embedded image
- Refresh the project record with its new commit and generation metadata before rebuilding the UI.
- Change the preview URL revision immediately so the stale cached texture is released and replaced.
- Keep the update targeted without rescanning project folders.
- Add regression coverage for verification ordering and thumbnail revision invalidation.

The updated thumbnail is now displayed without requiring another click, reopening the panel, or restarting LichtFeld Studio.

Related to #2228


src/python/lfs_plugins/asset_manager_panel.py
```diff
@@ -2737,6 +2738,10 @@ def complete(error=None, facts=None, inspection_error="") -> None:
                 if facts:
                     self._inspection_by_asset[asset_id] = facts
                 self._inspection_errors.pop(asset_id, None)
+                if reverify_asset and not asset.get("recent_only"):
+                    verify_asset = getattr(self._asset_index, "verify_asset", None)
+                    if callable(verify_asset):
+                        self._library_command("verify_asset", asset_id)
```
MrNeRF, line 2746: This still leaves the old thumbnail visible when the project already has cached inspection details. The refreshed catalog gets the new revision, but `_inspection_by_asset` overwrites it with the old one. I reproduced this on this head. Please clear or refresh that cache before rebuilding the row, and cover this case in the test.

MrNeRF (changes requested):
Please fix the cases below before merging.

francescofugazzi:
Cleared the cached inspection entry before rebuilding a reverifed thumbnail row, while preserving any fresh operation facts collected afterward. The regression now starts with a stale cached card and verifies that it is discarded and the row advances from the old thumbnail revision to the newly verified one.

MrNeRF (approved):
Rechecked the cached-thumbnail case: it fails on the previous version and passes with this fix. All 193 panel checks pass.

Issue #2209, opened 2026-09-19
Published project outside of "projects" folder cannot be updated


kmeirlaen:
* open LFS
* open .ply
* save project in a CUSTOM location (outside of project folder)
* close LFS

* open LFS
* open previously save project
* file, publish to gallery 
* close LFS 

Confirm project is in gallery

* open LFS
* open project manager
* widen the screen so that the folders appear on the left
* select "recent"
* confirm the project above is displayed in the list (OK)
* NOK: Gallery state displays "not in gallery"
* open the project
* use File, Publish to gallery
* NOK: no publish window appears and nothing in project manager can be used any more (no buttons, no menus, ...)

kmeirlaen:
@francescofugazzi even after https://github.com/MrNeRF/LichtFeld-Studio/pull/2233 this remains an issue.  The first NOK (gallery state displays "not in gallery") is solved, but the second "publish to gallery" remains an issue - nothing seems to happen and project manager is still locked.

MrNeRF:
The hidden publish window is addressed on master by #2244 and #2248. Tests pass for restoring the review after a layout change, reopening it without losing choices, and keeping the real review visible during the publish handoff. Recent files also use the saved project's identity once inspected.

A remaining external-file publishing dead end is fixed in #2256: a deleted online copy can now reach the publish-again review. Leaving this issue open until the original Windows steps confirm both the status and the freeze are gone.

PR #2233, opened 2026-09-19
Fix Gallery state and publishing for external recent projects

francescofugazzi:
## Summary

- resolve Gallery relationships for projects shown only in Recent using the native project UUID obtained through project inspection
- retain the temporary Recent row identity and existing catalog safety restrictions
- prevent published external projects from being treated as unlinked solely because they are outside configured project folders
- resume File → Publish to Gallery automatically after the required Gallery refresh
- avoid opening a modal refresh notice while Project Manager remains active
- revalidate the Gallery account and active project path before continuing the publishing workflow
- keep external Recent projects outside the catalog unless the user explicitly adds them
- add regression coverage for external Recent project identity and deferred File-menu publishing

Related to #2209


src/python/lfs_plugins/asset_gallery_ui.py
```diff
@@ -147,10 +147,18 @@ def _gallery_changed(self, snapshot):
 
     def _has_gallery_link(self):
         asset = self._get_selected_asset() or {}
-        return asset.get("id") in self._gallery_state.get("links", {}) or bool(self._gallery_scene(asset))
+        return self._gallery_project_id(asset) in self._gallery_state.get("links", {}) or bool(self._gallery_scene(asset))
```
MrNeRF, line 150: This now shows the Gallery section for external Recent projects, but the published format, size and time stay blank. `_gallery_published_summary()` still looks up the temporary Recent ID. Please use the inspected project ID there too. I reproduced it; that change makes the test pass.

MrNeRF (changes requested):
One detail is still missing for external Recent projects.

francescofugazzi:
Updated `_gallery_published_summary()` to resolve links through the inspected native project UUID for external Recent entries. The regression now also verifies that the published format, size, and relative time are displayed while the project remains outside the catalog.

MrNeRF (approved):
Rechecked the missing Gallery summary: it fails on the previous version and passes with this fix. All 358 panel, menu and Gallery checks pass.

Issue #2215, opened 2026-09-19
Cannot publish as SOG


kmeirlaen:
It is not posibble to publish a splat that has been edited as SOG into the portal.

* open LFS
* create new project
* drag a .ply into the project
* edit the .ply by removing some splats
* save the project
* file, publish to gallery
* choose "SOG"
--> ERROR: the download file is damaged

<img width="753" height="510" alt="Image" src="https://github.com/user-attachments/assets/e294410c-1554-4889-874d-a0b99b2a67bf" />

<img width="765" height="910" alt="Image" src="https://github.com/user-attachments/assets/72c3e851-4616-4c72-906f-e4b1e95342be" />

kmeirlaen:
tested in e82b6326a - ok

PR #2234, opened 2026-09-19
Fix SOG publication after soft-deleting splats

francescofugazzi:
## Summary

Fix SOG gallery publications being rejected as damaged after splats were removed from a project.

## Problem

Gallery publication records the number of visible splats in the portable project metadata. However, the SOG encoder previously serialized every stored row, including rows marked as soft-deleted.

This produced a mismatch between:

- the visible splat count stored in the publication metadata; and
- the actual point count encoded in the SOG payload.

The portable-project validation correctly rejected that inconsistent download.

## Changes

- Filter soft-deleted splats before SOG encoding.
- Compact a cloned `SplatData` instance so exporting does not mutate the source model.
- Preserve the existing path when no splats require filtering.
- Reject empty visible datasets and inconsistent compaction results explicitly.
- Apply the correction at the SOG encoder boundary so it also covers direct and directory-based SOG exports.
- Add a gallery publication regression test that verifies:
  - the publication metadata contains the visible count;
  - the generated sidecar is a SOG file;
  - reopening the SOG returns only the visible splats.

## Scope

The change is limited to SOG handling and its gallery publication regression coverage. 

Related to #2215


MrNeRF (approved):
Checked deleted-splat export on CPU and CUDA, including empty results and preserving the source scene. The regression fails before this change and passes afterward.

